PR #3, opened 2019-03-22
Added test lib download and build

db4ple:
This is a test. The Travis Build should run through without problems and also without building the STM32 binaries due to lack of certain Travis variables for download of STM32 StdPeriph lib set.

Ok, seems to work.

To build with STM32, one needs to specify STDLIBURL (a download url for an encrypted ZIP Archive of STM32F4xx_DSP_StdPeriph_Lib_V1.8.0) and STDLIBKEY (the key to decrypt the encrypted archive).  The variables must be defined in the Travis Settings for the build.


drowe67:
@db4ple thanks this is great :+1: So is the idea that each of us (if we want to build the stm32 code) sets up our own STDLIBURL and STDLIBKEY pointing to our copy of the stm32 code?

Also I'm trying to set up GitHub so that a PR needs to be reviewed before merging, or that only I can merge. Not sure what button to press on GitHub (tried a few last night).  I like the idea of all the collaborators have push access to branches (so we can develop together), just perhaps not master.

db4ple:
To question 1: Yes. 
To question 2: No idea if this is possible in GitHub. 
However, I have no problem to use by default the PR as the way of working unless there is a good reason to use the "collaborator's super powers".

drowe67:
Yeah re (2) I cant find the **Restrict who can push option** mentioned on on https://help.github.com/en/articles/enabling-branch-restrictions when I look at the codec2 repo settings.  Maybe its a paid-for option.

So as you suggest, lets just make it a convention that I get to press the merge button.

drowe67:
Ahh .. looks like I need to be an "organisation" to use this feature.

PR #6, opened 2019-03-26
Missing Chart in ofdm_demod_c.m

srsampson:
When viewing the code to see exactly what this program was displaying, I noticed that figure 4 was used twice, so renamed the last one figure 5, and the missing chart now displays.

### DIFF
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1305,21 +1305,20 @@ int freedv_rx(struct freedv *f, short speech_out[], short demod_in[]) {
     
     if ( (FDV_MODE_ACTIVE( FREEDV_MODE_1600, f->mode)) || (FDV_MODE_ACTIVE( FREEDV_MODE_700, f->mode)) || 
          (FDV_MODE_ACTIVE( FREEDV_MODE_700B, f->mode)) || (FDV_MODE_ACTIVE( FREEDV_MODE_700C, f->mode))) {
-
-        float gain = 1.0;
         
         /* FDM RX happens with complex samps, so do that */
         COMP rx_fdm[f->n_max_modem_samples];
+
         for(i=0; i<nin; i++) {
-            rx_fdm[i].real = gain*(float)demod_in[i];
-            rx_fdm[i].imag = 0.0;
+            rx_fdm[i].real = ((float) demod_in[i] / 32767.0f);
+            rx_fdm[i].imag = 0.0f;
         }
+
         return freedv_comprx(f, speech_out, rx_fdm);
     }
 
     if (FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode)) {
-        float gain = 2.0; /* keep levels the same as Octave simulations and C unit tests for real signals */
-        return freedv_shortrx(f, speech_out, demod_in, gain);
+        return freedv_shortrx(f, speech_out, demod_in);
     }
     
     return 0; /* should never get here */
@@ -1346,9 +1345,11 @@ int freedv_comprx_fsk(struct freedv *f, COMP demod_in[], int *valid) {
         /* 2400B needs real input samples */
         int n = fmfsk_nin(f->fmfsk);
         float demod_in_float[n];
+
         for(i=0; i<n; i++) {
             demod_in_float[i] = demod_in[i].real;
         }
+
         fmfsk_demod(f->fmfsk,(uint8_t*)f->tx_bits,demod_in_float);
         f->nin = fmfsk_nin(f->fmfsk);
     }
@@ -1394,9 +1395,10 @@ int freedv_floatrx(struct freedv *f, short speech_out[], float demod_in[]) {
     assert(nin <= f->n_max_modem_samples);
     
     COMP rx_fdm[f->n_max_modem_samples];
+
     for(i=0; i<nin; i++) {
         rx_fdm[i].real = demod_in[i];
-        rx_fdm[i].imag = 0;
+        rx_fdm[i].imag = 0.0f;
     }
 
     return freedv_comprx(f, speech_out, rx_fdm);
@@ -1765,7 +1767,7 @@ static int freedv_comprx_700(struct freedv *f, COMP demod_in_8kHz[], int *valid)
     [ ] deal with out of sync returning nin samples, listening to analog audio when out of sync
 */
 
-static int freedv_comprx_700d(struct freedv *f, short demod_in_8kHz[], float gain, int *valid) {
+static int freedv_comprx_700d(struct freedv *f, short demod_in_8kHz[], int *valid) {
     int   bits_per_codec_frame, bytes_per_codec_frame;
     int   i, j, bit, byte, nout, k;
     int   n_ascii;
@@ -2024,7 +2026,7 @@ int freedv_comprx(struct freedv *f, short speech_out[], COMP demod_in[]) {
 }
 
 /* 700D version */
-int freedv_shortrx(struct freedv *f, short speech_out[], short demod_in[], float gain) {
+int freedv_shortrx(struct freedv *f, short speech_out[], short demod_in[]) {
     assert(f != NULL);
     int                 bits_per_codec_frame, bytes_per_codec_frame;
     int                 i, nout = 0;
@@ -2036,7 +2038,7 @@ int freedv_shortrx(struct freedv *f, short speech_out[], short demod_in[], float
     bytes_per_codec_frame = (bits_per_codec_frame + 7) / 8;
 
     if (FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode)) {
-        nout = freedv_comprx_700d(f, demod_in, gain, &valid);
+        nout = freedv_comprx_700d(f, demod_in, &valid);
     }
     
     if (valid == 0) {
@@ -2110,7 +2112,7 @@ int freedv_codecrx(struct freedv *f, unsigned char *packed_codec_bits, short dem
     int bytes_per_codec_frame = (bits_per_codec_frame + 7) / 8;
 
     if (FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode)) {
-        freedv_comprx_700d(f, demod_in, 1.0, &valid);
+        freedv_comprx_700d(f, demod_in, &valid);
 
         int data_bits_per_frame = f->ldpc->data_bits_per_frame;
         int frames = data_bits_per_frame/bits_per_codec_frame;
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1799,8 +1799,6 @@ static int freedv_comprx_700d(struct freedv *f, short demod_in_8kHz[], float gai
     int parityCheckCount = 0;
     uint8_t rx_uw[ofdm_nuwbits];
 
-    float new_gain = gain / OFDM_AMP_SCALE;
-    
     /* echo samples back out as default (say if sync not found) */
     
     *valid = 1;
@@ -1813,13 +1811,13 @@ static int freedv_comprx_700d(struct freedv *f, short demod_in_8kHz[], float gai
     /* looking for modem sync */
     
     if (ofdm->sync_state == search) {
-        ofdm_sync_search_shorts(f->ofdm, demod_in_8kHz, new_gain);
+        ofdm_sync_search_shorts(f->ofdm, demod_in_8kHz);
     }
 
      /* OK modem is in sync */
     
     if ((ofdm->sync_state == synced) || (ofdm->sync_state == trial)) {
-        ofdm_demod_shorts(ofdm, rx_bits, demod_in_8kHz, new_gain);
+        ofdm_demod_shorts(ofdm, rx_bits, demod_in_8kHz);
         ofdm_disassemble_modem_frame(ofdm, rx_uw, payload_syms, payload_amps, txt_bits);
 
         f->sync = 1;
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1418,14 +1418,10 @@ static int freedv_comprx_fdmdv_1600(struct freedv *f, COMP demod_in[], int *vali
     bytes_per_codec_frame = (bits_per_codec_frame + 7) / 8;
     nout = f->n_speech_samples;
 
-    COMP ademod_in[f->nin];
-    for(i=0; i<f->nin; i++)
-        ademod_in[i] = fcmult(1.0/FDMDV_SCALE, demod_in[i]);
-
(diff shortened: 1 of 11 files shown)


srsampson:
The previous commit passed the Travis check.

I added a new commit to change the way the shorts are converted to complex in ofdm.c.

I noticed gigantic float values which come from the shorts being converted to complex, without first removing the offset +/- 32768. So I changed it to divide by 32768.0 before multiplying by the scale factor. Essentially converting the shorts between +/- 1.0 values.

This change produces the same output and charts in ofdm_demod, and seems to work the same, and not any better, but the "make test" fails.

```
$ make test
Running tests...
Test project /home/ssampson/okc/codec2/build_linux
      Start  1: test_FDMDV_modem_octave_port
 1/11 Test  #1: test_FDMDV_modem_octave_port ...............   Passed    9.50 sec
      Start  2: test_OFDM_modem_octave_port
 2/11 Test  #2: test_OFDM_modem_octave_port ................   Passed   12.12 sec
      Start  3: test_OFDM_modem_octave_port_Nc_31
 3/11 Test  #3: test_OFDM_modem_octave_port_Nc_31 ..........   Passed   15.07 sec
      Start  4: test_OFDM_modem_AWGN_BER
 4/11 Test  #4: test_OFDM_modem_AWGN_BER ...................   Passed    1.56 sec
      Start  5: test_horus_binary
 5/11 Test  #5: test_horus_binary ..........................   Passed    0.16 sec
      Start  6: test_ldpc_enc_dec
 6/11 Test  #6: test_ldpc_enc_dec ..........................   Passed    0.16 sec
      Start  7: test_ldpc_enc_dec_HRAb_396_504
 7/11 Test  #7: test_ldpc_enc_dec_HRAb_396_504 .............   Passed    0.42 sec
      Start  8: test_ldpc_check
 8/11 Test  #8: test_ldpc_check ............................   Passed    0.33 sec
      Start  9: test_FSK_modem_mod_demod
 9/11 Test  #9: test_FSK_modem_mod_demod ...................   Passed    0.06 sec
      Start 10: test_FreeDV_API_700D_AWGN_BER
10/11 Test #10: test_FreeDV_API_700D_AWGN_BER ..............***Failed    0.31 sec
      Start 11: test_FreeDV_API_700D_AWGN_BER_Interleave
11/11 Test #11: test_FreeDV_API_700D_AWGN_BER_Interleave ...***Failed    0.31 sec

82% tests passed, 2 tests failed out of 11

Total Test time (real) =  40.02 sec

The following tests FAILED:
	 10 - test_FreeDV_API_700D_AWGN_BER (Failed)
	 11 - test_FreeDV_API_700D_AWGN_BER_Interleave (Failed)
Errors while running CTest
Makefile:72: recipe for target 'test' failed
make: *** [test] Error 8
```

drowe67:
@srsampson Yes I've noticed some large values too, @donmr do you have any thoughts on this?  Also I think complex input at this stage would be useful for @db4ple work to avoid the need for an outboard SSB filter.

Steve - it's possible to run those failing tests stand alone in verbose mode to get a feel for what's happening:
```
ctest -V -R test_FreeDV_API_700D_AWGN_BER
```

srsampson:
I added some parens to make both changes the same. This is merely a hint to the compiler that we want the short converted first, and then gain multiplied. No change in test error.

srsampson:
Yes, that's my next test to see why the BER tests fail.

srsampson:
Just an FYI so far,

Without interleave, test fails high BER
```
BER......: 0.1016 Tbits: 12896 Terrs:  1310
Coded BER: 0.0294 Tbits:  5824 Terrs:   171
```
With interleave, test fails due to high BER
```
BER......: 0.1051 Tbits:  8106 Terrs:   852
Coded BER: 0.0290 Tbits:  3584 Terrs:   104
```

donmr:
My changes were to reduce the memory needed.  Storing an array of complex
values takes 4 times as much memory as an array of shorts.  So I delayed
the conversion as far in as practical.

For the conversion, we want to avoid large or small exponents to get the
most resolution so these changes sound good.

If we wanted to maintain an exact binary match between implementations
then we would have to fix the order of operations like this to be the
same.  But I don't think we are trying to do that here.

Don

drowe67:
@srsampson those results are pretty close to a pass, although IIRC the coded BER for the interleaved case should be batter.  Maybe see if there are a burst of errors at the start, that can skew the entire BER.  For example with small changes it might sync up with a different freq offset, leading to some initial errors until it tracks out the error.  A longer simulation might let it get better results too.

srsampson:
I changed the CMakeLists.txt to count=500 for AWGN and count=800 for AWGN with interleave.

"make test" passes and doppler tests give the same results.

Travis fails

srsampson:
Haven't figured out why Travis build is failing, but a local build passes.
```
      Start 10: test_FreeDV_API_700D_AWGN_BER
10/11 Test #10: test_FreeDV_API_700D_AWGN_BER ..............   Passed    1.23 sec
      Start 11: test_FreeDV_API_700D_AWGN_BER_Interleave
11/11 Test #11: test_FreeDV_API_700D_AWGN_BER_Interleave ...   Passed    1.96 sec

100% tests passed, 0 tests failed out of 11
```

drowe67:
Hi Steve, I just pulled yr branch and am failing the last two tests as well - just like Travis.  master seems OK, about 8% raw errors.  Do you have a feel for why this branch has a higher BER, e.g. initial freq offset?  Seems like such a simple change - just amplitude scaling.

srsampson:
It started failing on me today. No passes at all. That was weird.

I've been playing with it for an hour or so, and the new version is
definitely different. The biggest thing I see, is that the constellation is
not centered on the new version. It is shifted to the right.


dd bs=1280 count=120 if=/dev/zero | ./freedv_tx 700D - - --interleave 8
--testframes | ./cohpsk_ch - - -20 --Fs 8000 -f -10 | ./ofdm_demod
--high_doppler 0 --out /dev/null --testframes --nc 17 --verbose 1 --log
no_doppler.log


original

[image: Screenshot from 2019-03-27 14-42-38.png]

New

[image: Screenshot from 2019-03-27 14-43-15.png]

Will have to test more. I guess the modem likes the gigantic numbers :-)

srsampson:
P.S. What is considered passing??

Try as I might, the docs in the cmake are not enough to determine where the test is judged pass or fail.

drowe67:
> P.S. What is considered passing??
> 
> Try as I might, the docs in the cmake are not enough to determine where the test is judged pass or fail.

ofdm_demod returns 0 for pass, 1 for a fail, the threshold are set in there:
```
           if ((uncoded_ber >= 0.1f) && (coded_ber >= 0.01f))
                return 1;
  
```

srsampson:
OK, I may have a fix that needs review. The theory is we don't need the gain feature of the short functions. If I get rid of all that, my constellation is centered and the tests pass with only 120 frames.
```
$ dd bs=1280 count=120 if=/dev/zero | ./freedv_tx 700D - - --testframes | ./cohpsk_ch - - -20 --Fs 8000 -f -10 | ./freedv_rx 700D - /dev/null --testframes
cohpsk_ch ----------------------------------------------------------------------------------
Fs: 8000 NodB: -20.00 foff: -10.00 Hz fading: 0 inclip: 1.00 ssbfilt: 1
cohpsk_ch ----------------------------------------------------------------------------------
120+0 records in
120+0 records out
153600 bytes (154 kB, 150 KiB) copied, 0.177146 s, 867 kB/s
cohpsk_ch -----------------------------------------------------------------------------
SNR3k(dB): -2.00 C/No: 32.8 PAPR: 10.4 
cohpsk_ch -----------------------------------------------------------------------------
BER......: 0.0890 Tbits: 14384 Terrs:  1280
Coded BER: 0.0182 Tbits:  6496 Terrs:   118

$ dd bs=1280 count=120 if=/dev/zero | ./freedv_tx 700D - - --interleave 8 --testframes | ./cohpsk_ch - - -20 --Fs 8000 -f -10 | ./freedv_rx 700D - /dev/null --interleave 8 --testframes
cohpsk_ch ----------------------------------------------------------------------------------
Fs: 8000 NodB: -20.00 foff: -10.00 Hz fading: 0 inclip: 1.00 ssbfilt: 1
cohpsk_ch ----------------------------------------------------------------------------------
120+0 records in
120+0 records out
153600 bytes (154 kB, 150 KiB) copied, 0.168129 s, 914 kB/s
cohpsk_ch -----------------------------------------------------------------------------
SNR3k(dB): -2.35 C/No: 32.4 PAPR: 10.7 
cohpsk_ch -----------------------------------------------------------------------------
BER......: 0.0834 Tbits: 10030 Terrs:   837
Coded BER: 0.0047 Tbits:  4480 Terrs:    21
```

srsampson:
Looks like Travis ran out of virtual machines...

donmr:
On Wed, Mar 27, 2019 at 02:29:22PM -0700, srsampson wrote:
> OK, I may have a fix that needs review. The theory is we don't need the gain feature of the short functions. If I get rid of all that, my constellation is centered and the tests pass with only 120 frames.

Before I created ofdm_demod_shorts, freedv_rx() made a copy of the input
(demod_in[]) multiplied by gain.  That was passed to freedv_comprx() which
called freedv_comprx_700d() which made another copy of the input divided
by OFDM_AMP_SCALE.  Finally that array was passed into ofdm functions.

In my changes to reduce the memory I wanted to get the same results as
before so I passed gain and OFDM_AMP_SCALE down to be used when the
shorts were finally converted to floats.

I see that the gain from freedv_rx is a constant of 2.0 so it doesn't do
a lot but changing it could change the results for some cases.

Don

srsampson:
That's a good point, and the gain in freedv_rx() in freedv_api.c should also be removed I think.

Especially the nothing-burger that uses a scale of 1.0 :-)

I believe the scaling was originally just for the transmit, as it is such a small value for the soundcard, but the receive I would think is easily adjusted using the sound card volume.

I note in the freedv_rx() reference to Octave simulations.

srsampson:
OK, I cut the cancer out, now I'll have to probe everything for bleeders (fmdv, cohpsk, etc)

srsampson:
Ran tests in README_fsk.txt and they all passed except:

Octave tfsk

It is looking for a function called fsk_horus_mod() which is MIA

1. Looks like fsk_horus_2fsk is commented out, but uncommenting it causes other failures.
I assume this test is obsolete.

```
>> tfsk
error: 'fsk_horus_mod' undefined near line 280 column 12
error: called from
    fsk_mod_test at line 280 column 10
    test_mod_horuscfg_randbits at line 295 column 26
    test_fsk_battery at line 561 column 10
    tfsk at line 621 column 7
```

srsampson:
Octave fdmdv_ut no-worky
```
>> fdmdv_ut
error: 'Nc' undefined near line 24 column 17
error: called from
    fdmdv_ut at line 24 column 9
```

drowe67:
> Ran tests in README_fsk.txt and they all passed except:
> 
> Octave tfsk
> 
> It is looking for a function called fsk_horus_mod() which is MIA
> 
>     1. Looks like fsk_horus_2fsk is commented out, but uncommenting it causes other failures.
>        I assume this test is obsolete.

Thanks Steve - yes I did some partial refactoring there.  Maybe log it as an [Issue](https://github.com/drowe67/codec2/issues)?

src/freedv_api.c
```diff
@@ -1593,18 +1591,9 @@ static int freedv_comprx_700(struct freedv *f, COMP demod_in_8kHz[], int *valid)
...
-    for(i=0; i<freedv_nin(f); i++)
-        demod_in[i] = demod_in_8kHz[i];
-
-    i = quisk_cfInterpDecim((complex float *)demod_in, freedv_nin(f), f->ptFilter8000to7500, 15, 16);
-    //if (i != f->nin)
-    //    printf("freedv_comprx decimation: input %d output %d\n", freedv_nin(f), i);
-
-    for(i=0; i<f->nin; i++)
-        demod_in[i] = fcmult(1.0/FDMDV_SCALE, demod_in[i]);
+    i = quisk_cfInterpDecim((complex float *)demod_in_8kHz, freedv_nin(f), f->ptFilter8000to7500, 15, 16);
```
drowe67: @srsampson This code affects FreeDV 700C - perhaps try a manual test or add a Ctest to check 700C is getting a similar BER to before?

src/horus_demod.c
```diff
@@ -227,28 +238,35 @@ int main(int argc, char *argv[]) {
...
             #endif
 
 	    fprintf(stderr,"]}\n");
             stats_ctr = stats_loop;
         }
+
         stats_ctr--;
 
-        if (fin == stdin || fout == stdin){
+        if (fin == stdin || fout == stdin) {
```
drowe67: @srsampson A lot of these are personal preference style changes.  We don't really have a coding convention so I'm prepared to let individual authors use their own style.  I'm not too bothered either way, but some authors may be ... so pls be careful to check with original authors before you make these sorts of changes.

drowe67:
> OK, I may have a fix that needs review. The theory is we don't need the gain feature of the short functions. If I get rid of all that, my constellation is centered and the tests pass with only 120 frames.

It would be nice to get to root cause here, as a QPSK modem shouldn't be sensitive to gain.  It might point to a deeper issue in the modem.

Do these changes affect the signal levels at the output of freedv_tx?  The idea is that each waveform (1600, 700C, 700D etc) should drive the SSB radio with about the same power output as the mode is changed, so users of FreeDV don't have to adjust their sound card levels as they change modes.

srsampson:
I did not change any TX scaling, I know that +/- 1.0 will not drive a sound card, so the scaling is important there.

drowe67:
Steve - could you pls move the fsk_xx changes to another branch and PR?  When I review the code here I'm having trouble working out what the important changes are that I need to review. Lets focus on the key changes - merge them, then perhaps handle stylistic changes and fsk_xx in another PR?

src/horus_demod.c
srsampson: Yes, I have a bad habit of touching code as I try to read through the compressed styles. I'll terminate that activity.

srsampson:
Will do. Too many commits outstanding, so it is too hard to follow now.

src/horus_demod.c
drowe67: Haha :actually it's a common habit - I do it too.

At $dayjob we have some serious style-Nazi's so this issue pops up a lot there too. I'm actually moving to a more compressed style myself but am happy to let individuals do their own thing in their modules (you get to make the call on ofdm.c for instance).  

I make a weak effort to stick with the style of the source file I am in at the time :-)

One way to handle this is to use auto-formatting, e.g. clang has emacs modules that formats the C code when you save a file.  But we would need to have everyone install it.

PR #15, opened 2019-04-02
Modification to add pilots to Stats

srsampson:
Fixes a bug in ofdm_get_demod_stats() to use the correct carriers in rx_symbols.

originally it copied carriers 0 to 16 where I think it should be using carriers 1 to 17 as the first and last carriers are just used by the pilots. and this section is for the data carriers.

Also

Adds another "optional" Stat which is the four pilot symbols (previous, this, next, future). By collecting these stats I envision a display view in a GUI or analysis program.

### DIFF
--- a/src/ofdm.c
+++ b/src/ofdm.c
@@ -86,6 +86,10 @@ static complex float *tx_uw_syms;
 static int *uw_ind; 
 static int *uw_ind_sym;
 
+#if defined(PILOT_STATS)
+static float rx_pilots[4][MODEM_STATS_NC_MAX];
+#endif
+
 static float ofdm_tx_centre; /* TX Center frequency */
 static float ofdm_rx_centre; /* RX Center frequency */
 static float ofdm_fs; /* Sample rate */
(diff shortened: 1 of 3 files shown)


src/ofdm.c
```diff
@@ -1660,14 +1677,25 @@ void ofdm_get_demod_stats(struct OFDM *ofdm, struct MODEM_STATS *stats) {
     assert(ofdm_rowsperframe < MODEM_STATS_NR_MAX);
     stats->nr = ofdm_rowsperframe;
 
-    for (c = 0; c < ofdm_nc; c++) {
+    /* Just copy data frames */
+
+    for (c = 1; c < (ofdm_nc + 1); c++) {
         for (r = 0; r < ofdm_rowsperframe; r++) {
```
drowe67: @srsampson - around line 1390 of ofdm.c:
 ```
          ofdm->rx_np[(rr * ofdm_nc) + (i - 1)] = rx_corr;
```
so I think the original rx_np index might have already accounted for the pilots?  Be good to have some way to check, like linking with freedv-gui and seeing if the scatter diagram works.

srsampson: You are correct. My eyes no longer saw "np".

srsampson:
Not working as planned

PR #13, opened 2019-03-31
Removed implict double promotion and other warnings

db4ple:
Mostly by using the float returning function instead of the function returning doubles
Hide warnings in external StdPeriph lib code via compiler option

It complements some other pull requests.

PR #8, opened 2019-03-28
FreeDV 2020 support

ac0zj:
implementing FreeDV 2020 mode into codec2 repo, mainly freedv_api.c changes, and linking in LPCNet.

PR #16, opened 2019-04-05
700D freedx_rx working with shorts and complex input

drowe67:
Building on @donmr 's fine work for low memory short array inputs.  We wish to provide complex float sample input for 700D as well, without changing the low memory short array functionality.  Complex input also useful for @db4ple 's projects (saves a SSB filter).

700D with optimised short code path:
```
/codec2/build_linux/src$ ./freedv_tx 700D ../../raw/ve9qrp.raw - - | ./freedv_rx 700D - -  | aplay -f S16_LE
```
700D with complex valued code path:
```
/codec2/build_linux/src$ ./freedv_tx 700D ../../raw/ve9qrp.raw - - | ./freedv_rx 700D - -  --usecomplex | aplay -f S16_LE
```

I'm sure this is non-optimal (lots of repeated code), but just looking to get a working 700d freedv_comprx() function for this PR, as I need similar code for FreeDV 2020.

Do we have any way to Ctest the full speech path?  Guess we could use a BER test, as modem performance will be sensitive to bugs the complex code path, we just won't be testing the speech decoding parts of the code.

### DIFF
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1317,6 +1317,7 @@ int freedv_rx(struct freedv *f, short speech_out[], short demod_in[]) {
         return freedv_comprx(f, speech_out, rx_fdm);
     }
 
+    /* special low memory version for 700D, to help with stm32 port */
     if (FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode)) {
         float gain = 2.0; /* keep levels the same as Octave simulations and C unit tests for real signals */
         return freedv_shortrx(f, speech_out, demod_in, gain);
@@ -1795,6 +1796,8 @@ static int freedv_comp_short_rx_700d(struct freedv *f, void *demod_in_8kHz, int 
 
     float new_gain = gain / OFDM_AMP_SCALE;
     
+    assert((demod_in_is_short == 0) || (demod_in_is_short == 1));
+
     /* echo samples back out as default (say if sync not found) */
     
     *valid = 1;
@@ -1809,6 +1812,8 @@ static int freedv_comp_short_rx_700d(struct freedv *f, void *demod_in_8kHz, int 
     if (ofdm->sync_state == search) {
         if (demod_in_is_short)
             ofdm_sync_search_shorts(f->ofdm, (short*)demod_in_8kHz, new_gain);
+        else
+            ofdm_sync_search(f->ofdm, (COMP*)demod_in_8kHz);
     }
 
      /* OK modem is in sync */
@@ -1816,6 +1821,9 @@ static int freedv_comp_short_rx_700d(struct freedv *f, void *demod_in_8kHz, int 
     if ((ofdm->sync_state == synced) || (ofdm->sync_state == trial)) {
         if (demod_in_is_short)
             ofdm_demod_shorts(ofdm, rx_bits, (short*)demod_in_8kHz, new_gain);
+        else
+            ofdm_demod(ofdm, rx_bits, (COMP*)demod_in_8kHz);
+            
         ofdm_disassemble_modem_frame(ofdm, rx_uw, payload_syms, payload_amps, txt_bits);
 
         f->sync = 1;
@@ -1969,11 +1977,11 @@ static int freedv_comp_short_rx_700d(struct freedv *f, void *demod_in_8kHz, int 
     return nout;
 }
 
-/* Original version for all but 700D */
+/* complex input rx API function */
 int freedv_comprx(struct freedv *f, short speech_out[], COMP demod_in[]) {
     assert(f != NULL);
-    int                 bits_per_codec_frame, bytes_per_codec_frame;
-    int                 i, nout = 0;
+    int bits_per_codec_frame, bytes_per_codec_frame;
+    int i, nout = 0;
     int valid = 0;
     
     assert(f->nin <= f->n_max_modem_samples);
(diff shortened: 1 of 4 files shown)


src/freedv_api.c
```diff
@@ -1996,49 +2000,71 @@ int freedv_comprx(struct freedv *f, short speech_out[], COMP demod_in[]) {
...
+/* memory efficient real short version - just for 700D  */
 int freedv_shortrx(struct freedv *f, short speech_out[], short demod_in[], float gain) {
     assert(f != NULL);
     int                 bits_per_codec_frame, bytes_per_codec_frame;
     int                 i, nout = 0;
     int valid = 0;
+
+    // At this stage short interface only supported for 700D, to help
+    // memory requirements on stm32
+    assert(FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode));
```
drowe67: @donmr I was concerned about people calling this function for non-700D modes.

src/ofdm.c
```diff
@@ -944,10 +944,9 @@ int ofdm_sync_search(struct OFDM *ofdm, COMP *rxbuf_in) {
 
     memcpy(&ofdm->rxbuf[0], &ofdm->rxbuf[ofdm->nin],
         (ofdm_rxbuf - ofdm->nin) * sizeof (complex float));
-    
     memcpy(&ofdm->rxbuf[(ofdm_rxbuf - ofdm->nin)],
-        rx, (ofdm_rxbuf - ofdm->nin) * sizeof (complex float));
-
+        rx, ofdm->nin * sizeof (complex float));
+
```
drowe67: @srsampson I was getting segfaults until I modified this.

drowe67:
OK so I tested this with freedv-gui in 700D mode and it worked OK (I used alsa loopback and full duplex).  i could even click-tune a few hundred Hz eitehr side which is a good test of complex freq shifts.  Built codec2/master and 700D wouldn't demodulate, but Tx signal looked good.

donmr:
I have been trying to get freedv_rx tests based on this to run here.

The new code is using more statically allocated memory, and then
running out of heap space as it runs.

I spent some time decodeing map files and found the older code (SVN
4118) uses 4800 bytes of .data in RAM with these as the larges chucks:

    1920 ../../../src/filter.o
    1068 /usr/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/libg.a(lib_a-impure.o)
    1040 /usr/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/libg.a(lib_a-mallocr.o)
    364 /usr/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/libg.a(lib_a-locale.o)
    168 ../../../src/gp_interleaver.o
    120 ../../../src/fdmdv.o
    56 ../../../src/cohpsk.o
    32 ../../../src/mpdecode_core.o
    12 /usr/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/librdimon.a(rdimon-syscalls.o)
    6 ../../../src/freedv_data_channel.o

Newer code (Git branch donmr-tst_api) uses 6888 bytes of .data in RAM with
these as the larges chucks:

    2560 ../../libcodec2_prof.a(filter.c.obj)
    1068 /usr/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/libg.a(lib_a-impure.o)
    1040 /usr/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/libg.a(lib_a-mallocr.o)
    808 ../../libcodec2_prof.a(codec2.c.obj)
    520 ../../libcodec2_prof.a(freedv_data_channel.c.obj)
    364 /usr/bin/../lib/gcc/arm-none-eabi/7.3.1/../../../../arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/libg.a(lib_a-locale.o)
    168 ../../libcodec2_prof.a(gp_interleaver.c.obj)
    144 ../../libcodec2_prof.a(fdmdv.c.obj)
    104 ../../libcodec2_prof.a(cohpsk.c.obj)
    32 ../../libcodec2_prof.a(mpdecode_core.c.obj)

Several chunks grew.  (Some were not in the top 10 before.)

Is there something I can do to select fewer mode functions in these tests?
Or anything else to reduce the memory again?

Otherwise we won't be fitting this into the smaller pcocessors.


Don

srsampson:
cohpsk.c, fdmdv.c, freedv_data_channel.c, and filter.c should be optioned out in an stm32 700D build.

db4ple:
For memory restricted systems we now have the defines to control which modes are compiled in:

-DFREEDV_MODE_EN_DEFAULT=0 -DFREEDV_MODE_1600_EN=1 -DFREEDV_MODE_700D_EN=1
-DCODEC2_MODE_EN_DEFAULT=0 -DCODEC2_MODE_1300_EN=1 -DCODEC2_MODE_700C_EN=1

will (more or less) only leave the code for FreeDV 1600 & 700D active.
This should do the trick. If not, we should analyze the call graph and use the conditional macros FDV_MODE_ACTIVE & CODEC2_MODE_ACTIVE (see freedv_api.c and codec2.c) to disable these paths.

We should update stm32/CMakeLists.txt to use  -fdata-sections -ffunction-sections -Xlinker --gc-sections
which may help here (if Don is already using cmake built binaries, these compile options are used in the stm32/unittest builds). I have a pull request prepared for that but I am waiting for my previous PR #13  to get accepted before I push this one.

Additionally we could switch to use Link Time Optimization for stm32 builds which may (!) lead to significant savings  at the expense of compile time and tool chain complexity (we would need to use an archiver which supports libraries with LTO compiled archive members, which is not enabled by default in all arm tool chains).

drowe67:
Nice work Don.  Be good to have a utility to dump these, we could even put a test around static memory to trap this problem in future.  Some ideas:
1. It would be fine to use one of @db4ple's clever macros to remove #define out ```filter.c```in low memory systems.  The SSB filter will constrain the OFDM signal bandwidth adequately.  If it's really a problem for some users we can develop a less CPU intense filter.  As well as memory this filter also dominates modulator CPU load on other platforms.
1. You don't need ```cohpsk```, @db4ple's macros should remove that if 700C is disabled.  And also (I would guess) the memory used by ```codec2```
1. What are ```lib_a-impure.o``` and ```lib_a-mallocr.o```?  Are they toolchain supplied?

drowe67:
> cohpsk.c, fdmdv.c, freedv_data_channel.c, and filter.c should be optioned out in an stm32 700D build.

Oops missed this comment.  Agrred - although be nice to keep fdmdv.c so we can have a 1600/700D box.

donmr:
I'd like to support 1600 and 700D, but it may be simpler to start with a
new binary for just 700D.  If so how should we name them, and how should
we change the CMake files???

Don

db4ple:
I think, we should first see how my cmake changes affect the generated binaries and then we should integrate the required switches for just having 700D / 1600 into cmake (either conditionally or always active) and check again...

src/freedv_api.c
```diff
@@ -1754,18 +1755,12 @@ static int freedv_comprx_700(struct freedv *f, COMP demod_in_8kHz[], int *valid)
...
-    [ ] error pattern support?
-    [ ] deal with out of sync returning nin samples, listening to analog audio when out of sync
+  700D demod function that can support complex (float) or real (short)
+  samples.  The real short samples are useful for low memory overhead,
+  such at the SM1000.
 */
 
-static int freedv_comprx_700d(struct freedv *f, short demod_in_8kHz[], float gain, int *valid) {
+static int freedv_comp_short_rx_700d(struct freedv *f, void *demod_in_8kHz, int demod_in_is_short, float gain, int *valid) {
     int   bits_per_codec_frame, bytes_per_codec_frame;
```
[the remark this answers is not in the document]
drowe67: I partially agree ... but an hesitant to add yet another layer of similarly named functions. I suspect ```freedv_api.c``` is getting to the point where it needs a better design/major refactoring.

drowe67:
will close this as it's been merged into https://github.com/drowe67/codec2/pull/8

PR #18, opened 2019-04-08
Documentation changes and alignment of .travis.yml

db4ple:
Added information regarding the recommended location / names of builds
for linux and stm32 builds.
Aligned build script in .travis.yml to follow recommendation

### DIFF
--- a/stm32/unittest/README_unittest.txt
+++ b/stm32/unittest/README_unittest.txt
@@ -2,6 +2,9 @@ README_unittest.txt
 Don Reid 2018/2019
 
 This is the unittest system for the stm32 implementation of codec2/FreeDV
+It is currently only working on linux systems. It requires a STM32F4xx processor 
+development board connected to/having a ST-LINK , e.g. 
+a STM32F4 Discovery board. 
 
 Objectives
 ==========
@@ -49,6 +52,14 @@ Debug and semihosting
 Building and Running the stm32 Unit Tests
 =========================================
 
+0/ Build codec2 for linux and stm32 
+
+   Run the linux cmake build using /path/to/codec2/build_linux 
+   as build directory, see instructions in the codec2 top level directory.
+
+   Run the stm32 cmake build using /path/to/codec2/stm32/build_stm32 
+   as build directory, see instructions in the stm32 directory.
+   
 1/ Build stlink:
 
   $ cd ~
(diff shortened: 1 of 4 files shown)


stm32/unittest/README_unittest.txt
```diff
@@ -49,6 +52,14 @@ Debug and semihosting
...
 =========================================
 
+0/ Build codec2 for linux and stm32 
+
+   Run the linux cmake build using /path/to/codec2/build_linux 
+   as build directory, see instructions in the codec2 top level directory.
+
+   Run the stm32 cmake build using /path/to/codec2/stm32/build_stm32 
+   as build directory, see instructions in the stm32 directory.
+
```
drowe67: @donmr @db4ple can you pls confirm that the instructions in codec2/stm32/README build Don's stm32 unit tests now?  Previously these were compiled from codec2/stm32/unittest/src?

stm32/unittest/README_unittest.txt
```diff
@@ -71,32 +82,13 @@ Building and Running the stm32 Unit Tests
...
-   Preipheral Library, this can be set by adding a local.mak file in
-   the codec2-dev/stm32/unittest directory:
-
-   $ cd ~/codec2-dev/stm32/unittest/src
-   $ echo 'PERIPHLIBDIR = /periph/lib/path/STM32F4xx_DSP_StdPeriph_Lib_V1.8.0' > local.mak
-
-3/ Now we can build the unittests:
-
-   $ cd codec2-dev/stm32/unittest/src
-   $ make
```
drowe67: @donmr - so you are no longer building yr tests here?

drowe67:
Suggest we work together to reconcile the various stm32 READMEs:
1. Currently three READMEs are required to build and run Don's tests.
1. There aren't any instructions on installing the arm toolchain.
1. Can we retire codec2/stm32/README.txt?  I think so.

I'm happy to help later this week, but need to know the correct steps myself in order to document them.

db4ple:
Yes, the unittests are built by the CMakeLists.txt and also the scripts have been update already to use the cmake generated unittests in the described location. I don't know if Don still uses the old files as reference, but I have been running the stm32 unittests built by cmake.

drowe67:
> Yes, the unittests are built by the CMakeLists.txt and also the scripts have been update already to use the cmake generated unittests in the described location. I don't know if Don still uses the old files as reference, but I have been running the stm32 unittests built by cmake.

OK.  @donmr how do you feel about using CMake to build your tests?  I recall you had some reservations about CMake.  I figure as you are the main stm32 developer this is your call.  I am reluctant to have two build systems running in parallel, so would like to make sure everyone is happy with such a change.

db4ple:
To answer your list:
1. Now that all is build using a single make from the main stm32 directory we could merge most of the instructions. 
2. Well, yes. There is also the information that the Ubuntu 18.04 arm toolchain is not working, we should mention that, too.
3. Yes, together with the handwritten Makefiles. However, we should check if there is something in them which we should carry over.
4. I think so. However, it is only the execution scripts which require a linux host, if run with openocd, they should run fine on Windows or MacOS.  In any case, we surely can build the c2enc and c2dec tools somehow from the stm32 cmake so that everything becomes a single build. But on the other hand, this is not too difficult to do in two steps.

stm32/unittest/README_unittest.txt
donmr: Correct, I am using CMake which builds in build_stm32/unittest/src (usually located in codec2/stm32).

donmr:
The unittest scripts use bash and various *nix* commands like grep.  There are a couple of c programs too (compare_ints, compare_floats), and one uses Python3.   I have not tried to run them in any environment other than a modern desktop Linux.

donmr:
You can remove the Makefiles.  If I want to reference the old contents I can check out an older version (I am keeping the last SVN based workspace around for a while).

drowe67:
>     4. I think so. However, it is only the execution scripts which require a linux host, if run with openocd, they should run fine on Windows or MacOS.  In any case, we surely can build the c2enc and c2dec tools somehow from the stm32 cmake so that everything becomes a single build. But on the other hand, this is not too difficult to do in two steps.

Oh sorry I should have made this clearer.  I am assuming that we continue to develop on Linux machines, and don't think we need to maintain a stm32 build/test system for other OSes.

I was referring to the discrete step of building codec2, before we can build the stm32 side.  I agree with Danilo's summary on this point (4 above).

db4ple:
Please merge these changes, and then we should continue the cleanup in terms of documentation and removal of Makefiles.

db4ple:
BTW, the STM32 Readme.txt contains some (outdated) information on how to setup the build tool chain for ARM builds. We should integrate README.txt and README.cmake into a new consistent README and up-to-date document give exact instructions for building the STM32.

drowe67:
> BTW, the STM32 Readme.txt contains some (outdated) information on how to setup the build tool chain for ARM builds. We should integrate README.txt and README.cmake into a new consistent README and up-to-date document give exact instructions for building the STM32.

Yes I agree - happy to work on this with you.

Issue #20, opened 2019-04-11
STM32 Unittests Failing


donmr:
Many of the stm32 unittests are failing in the master branch, but not all.  I'll be digging into it tonight and tomorrow...

Don

drowe67:
Ach ... we need to stop breaking yr work Don!  Some tests perhaps.

db4ple:
Good Morning,
the culprit seem to be not so good settings in the link process, which I introduced to reduce the memory usage. Essentially we are missing support for %f in printf since I switched to the newlib "nano" libc which does not do float printf by default. 
I am not through all of the tests, but it looks good now once printf float gets enabled.

drowe67:
Fine Business @db4ple, BTW let me know if you want my Discovery/laptop turned on, I switched it off to save power ... happy to leave it it on if it's getting daily use.

db4ple:
Well, I will work on the "remoting" of stm32 unittests locally first in my internal network. I'll let you know once we are good to integrate this into travis and need to test it with your laptop.

@donmr : BTW, even now not all test pass (I am currently running the ofdm tests) but this seems to be due to slight changes in the output. E.g. the test_ofdm_demod_AWGN now reports sllightly different (lower) BER which lets the test report a FAIL.  I think/hope is an "acceptable" fail and can be resolved by changing the reference result. But this is not for me to decide/fix...

drowe67:
> @donmr : BTW, even now not all test pass (I am currently running the ofdm tests) but this seems to be due to slight changes in the output. E.g. the test_ofdm_demod_AWGN now reports sllightly different (lower) BER which lets the test report a FAIL. I think/hope is an "acceptable" fail and can be resolved by changing the reference result. But this is not for me to decide/fix...

The measured BER test is a random variable, whose mean is the actual BER.  Every time we run a test, we sample that random variable, and get a slightly different result. Someone who knows something about statistics (not me unfortunately!) could work out a 5% confidence interval or upper bound.  What we are trying to trap with these tests is an implementation error that has pushed the mean BER up.

donmr:
Is the fix going to be commited??
Yes, anyone changing the STM32 build files needs to run the tests we have.
I don't know when this change came in, I pulled the proposed changes I saw and it all worked for me.
Somehow this got missed, that happens...

donmr:
If you tell me what you changed to get it mostly working, I'll put that in and work on the rest of the issues.
[...]

PR #21, opened 2019-04-12
Fix for broken stm32 unittest (issue #20)

db4ple:
It was due to missing %f support in printf after switching to newlib nano.
Some additional unittest fail but this is not related to this issue, these
are impacted by code changes in codec2 source...


donmr:
I merged this with a new branch made from master and all of the ofdm tests passed.  Please pull it.

db4ple:
BTW, I made some progress in remote hardware use for unit testing. With an update-to-date openocd (and some modifications to the run_stm32_prog) I am able to run tests remotely. Some tests pass and some other tests produce "too much" output data and thus fail. But all in all it looks promising.

Issue #23, opened 2019-04-13
STM32 remote testing with OpenOCD


db4ple:
@donmr , @drowe67 
I converted the existing run_stm32_prog script, so that is uses openocd via ssh.
Some but not all tests passed as expected. When run with semihosting through gdb (i.e. remote fileio) some tests produce to short or too large output. When run exactly the same way using openocd BUT using local semihosting fileio (openocd command: semihosting_fileio disable vs. semishosting_fileio enable), everything passes as expected. Which means, there is an issue related to the fileio stuff either in openocd or in the gdb or in the actual binary. Further drilling down let me find that fread behaves sometimes strange (reads too much or not enough). There are two workarounds to fix that: 

1. We use `setbuf(fin,NULL);`  to disable buffering of the input file. Works but is ultra slow (6min runtime for a single tests vs. ~30s ).
1. We switch to basic fileio using `open/read/write` (as opposed to `fopen/fread/fwrite). This is almost as fast as the original code and does not seem to have the problems. 

Or we leave it as is and find another way to remotely execute the tests. But honestly, I think the conversion (option 2) is feasible. Takes a couple of minutes per test and is basically a no-brainer.

donmr:
I have tried to use #2 for newer tests.  There is no need for additional 
buffering by stdio.

We should also check that all tests close the files to flush them.

db4ple:
Thanks!
Using the "open/read/write/close" has the benefit of not requiring explicit flush before close.
I will run all existing tests using openocd and convert those failing because of fread in a first round. 
Second round will include all remaining tests for consistency.

db4ple:
@donmr : I noticed that the tst_codec2_enc_1600 test does not pass anymore. Reason is simple:
The new CMakeLists.txt disables mode 1600 (only 1300 and 700C are enabled). Reason for that is that FreeDV 1600 actually uses codec2 1300 not codec2 1600.
So we have a couple of options again:

1. Change the tests to use the codec2 1300 (which is in fact the most popular one and the one used in sm1000). Requires some minor edits to the tests.
2. Change the compile settings to include 1600 as well. This would increase the footprint of binaries and test a mode not widely used.

This time I vote for 1.

donmr:
My mistake.  Since the SM1000 use  FDV 1600 I made that test to verify my 
test code.  I can change it to  1300 instead.

Don

db4ple:
[...]

PR #24, opened 2019-04-13
Added support for STM32 testing with openocd (#23)

db4ple:
You'll need a recent openocd built from git!
By default st-util is run as before. Passing --openocd to the tests scripts
will activate openocd.

Some of the tests have been switched to use open etc. instead of fopen
since otherwise gdb_fileio with openocd will not work.

The local stm32f4_machdep.c has been removed, as it is more less identical
to the main one.

The main stm32f4_machdep.c now uses the length limited string functions
as a safeguard.

Tests with codec2 1600 have been replace with the codec2 1300 (which is used
in FreeDV 1600) so that it matches the codec used in sm1000.

The AWGN / fade tests for ofdm /ofdm_ldpc do not pass currently,
all other stm32 tests pass with st-util and openocd.

Removed stm32 unittest build from .travis.yml as it is now done using
the stm32 cmake


donmr:
Sounds good, I won't be able to try it until tomorrow though.

drowe67:
I followed the README_unitest.txt and it worked great for ```./run_all_ldpc_tests --openocd```.

Have ```./run_all_stm32_tests --openocd``` running now.  The Codec 2 700C test took a while to run and until I looked at the Discovery LEDs I thought it was stuck.  Is there any way to add some sort of progress indicator for the impatient like me?  Especially when we run "headless" tests and can't see the binky lights :-)

db4ple:
Well, I am with you on this. We could use stdout (or stderr) and show it on the console.

drowe67:
Wonder if there is a way we can run automated tests on a SM1000?  I could hook the SM1000 JTAG up to a STLINK, and pipe audio to the laptop mic/spkr ... perhaps even "push" the SM1000 buttons with some GPIOs and electronic switches.

Not sure how to evaluate modem tomes into speech out though.

donmr:
The trick is getting the levels right for mic and speaker in and out,
all 4 combinations are needed.  I have a test rig here with 2 USB sound
cards, one of which has a mic level input.  

The next time someone does a board like the SM1000, it would be good to
have jumpers to change the signals to all line level.  Then this would
be much simpler.

Don

donmr:
This works for me with st-util.  OpenOCD is not working with either the Ubuntu package or from git.

I'm getting an usage message from killall in ocd_stderr.log which is odd.
If I run gdb manually and give it 2 shell commands, one for killall and one for openocd then it all seems to work fine.  My guess is that gdb is interpreting part of the shell command not just passing it all to bash.

ase-2017:
Argh, yes. I broke openocd when used without ssh. Fixed that.

db4ple:
Argh, yes. I broke openocd when used without ssh. Fixed that.
Now tested st-util , openocd w/o ssh, openocd w/ ssh. Need unittests for that ;-)
BTW, Ubuntu openocd will not work (at least for 18.10 or older)

donmr:
Better, now it starts running but then hangs.  If I just a single test:

run_stm32_tst tst_codec2_enc 1300 --openocd

It hangs up.  The file "stm_stderr.txt" contains the message "Error opening config file" although the file exists and contains the correct data.  The end of ocd_stderr.log has:

semihosting is enabled
semihosting fileio is enabled
Info : Unable to match requested speed 2000 kHz, using 1800 kHz
Info : Unable to match requested speed 2000 kHz, using 1800 kHz
adapter speed: 1800 kHz
target halted due to debug-request, current mode: Thread 
xPSR: 0x01000000 pc: 0x08010f28 msp: 0x20020000, semihosting fileio

I am using openocd from the master branch of their repo, is that what you have?

db4ple:
@donmr 
I have no clear idea why this does not work for you. It worked for me and I think for David as well.

1. I assume testing usig st-util worked for you?
2. Did you try to run the tests from cmake as well (`ctest` or `make test`)? 
3. Is there any test which works for you with openocd or are all failing when accessing file input?

Is the git openocd (btw, I used the master branch like you) the only openocd installed? If not, check if not accidentially the wrong openocd gets opened.

donmr:
First, I am doing all of this in a copy of https://github.com/db4ple/codec2/tree/remote_testing

I started with "run_all_stm32_tests", when the first test hung I went to
the test run directory and tried just "run_stm32_prog ..." to see what
was happening.

On Mon, Apr 15, 2019 at 12:10:31AM -0700, db4ple wrote:
> @donmr 
> I have no clear idea why this does not work for you. It worked for me and I think for David as well.
> 
> 1. I assume testing usig st-util worked for you?

Yes.

> 2. Did you try to run the tests from cmake as well (`ctest` or `make test`)? 

Cmake uses st-util not openocd, it mostly works.

Running "make test" from the stm32/build_stm32 directory gets the first
13 tests passing, including tst_codec2_enc_1300, then tst_codec2_enc_700C
fails and then tst_codec2_dec_1300 hangs up.

tst_codec2_enc_700C and tst_codec2_dec_1300 pass when I run them by
themselves with run_stm32_tst and st-util.

Note that the codec2_dec tests play audio because I could not find a way to
compare the outputs usefully.  We may not want that in a remote testing suite!

> 3. Is there any test which works for you with openocd or are all failing when accessing file input?

I tried a few others which seemed to have the same problem.

> Is the git openocd (btw, I used the master branch like you) the only openocd installed? If not, check if not accidentially the wrong openocd gets opened.

Yes, just one openocd.

Don

db4ple:
If you pass -DUT_PARAMS=--openocd to the initial  stm32 cmake call, your "make test" tests will be run using openocd.
BTW, may it would be better to use --noload (which prevents to flash the executable to the board) instead of load. For manual testing this is the better default and for automated testing we can add --noload easily were it is applicable.

donmr:
OK, but that won't be much use to me until the tests stop hanging.

drowe67:
> Better, now it starts running but then hangs. If I just a single test:
> 
> run_stm32_tst tst_codec2_enc 1300 --openocd

@donmr _ just tried this on ```bear``` (my laptop) and it ran in about 10s,  You could perhaps try it there, openocd is installed, and you can find the compiler & stm32 peripheral lib in ~/home/david/Downloads, or feel free to download and install them yourself on that machine.

donmr:
Where is your arm compiler??

db4ple:
@donmr , @drowe67  : Just ran my branch db4ple/remote_testing on `bear`. It works w/o problems with st-util. 4 tests failed but it seems to  be an issue related to the linux code, since the refrerence results are different from my older Ubuntu, the stm32 output, however, is identical so these fails have nothing to do with the stm32 side.

Same result with openocd, but in about half the time....

I have no idea why it does not work for you, Don.

db4ple:
/home/david/gcc-arm-none-eabi-8-2018-q4-major/bin or /home/danilo/gcc-arm-none-eabi-8-2018-q4-major/bin

donmr:
Still having trouble getting cmake to work on bear.

I set PATH=$PATH:~/gcc-arm-none-eabi-8-2018-q4-major/bin and used the
standard stm32/cmake/STM32_Toolchain.make

I get this at the top:

/usr/bin/objcopy: Unable to recognise the format of the input file `sm2000_stw.elf'

and a similar error in unittest/src if I try there.

I don't know CMake well enough to debug this...

Don

db4ple:
My command line on `bear` for a clean cmake build (just tested it):
```bash
PATH=/home/danilo/gcc-arm-none-eabi-8-2018-q4-major/bin:$PATH
cd /home/danilo/git/codec2/stm32
mkdir -p build_stm32
cd build_stm32
rm -rf *
 cmake .. -DPERIPHLIBDIR=/home/danilo/STM32F4xx_DSP_StdPeriph_Lib_V1.8.0/ -DCMAKE_TOOLCHAIN_FILE=/home/danilo/git/codec2/stm32/cmake/STM32_Toolchain.cmake   -DUT_PARAMS=--openocd
make
```

donmr:
With Danilo's script it works, tst_codec2_enc_1300 Passed using openocd.  That is good enough to merge this, I will debug my system here.

drowe67:
> With Danilo's script it works, tst_codec2_enc_1300 Passed using openocd. That is good enough to merge this, I will debug my system here.

Good work Don - I'm glad you made some progress.  This is a large PR, and I agree we should merge without too much more work.  A few little tidy up issues:

1. @db4ple, can you pls edit README_unitest.txt (and maybe codec2/stm32/README.md) to bring them up to date with this PR? For example the command to run all tests, and expected results. Suggest merging the READMEs recently merged into master from https://github.com/drowe67/codec2/tree/issue-20. The goal is that anyone reading the documentation will be able to set up, run the tests, and a get a firm yes/no answer - we will be using this as a gateway for future merges.

2. Can either of you pls confirm all tests pass? I recall some issues with BER thresholds.  I'm looking to use this as a baseline for future work, for example when I merge https://github.com/drowe67/codec2/tree/brad-2020 - I want make sure I don't break any stm32 work.  So I need a passing set of tests as a baseline. Pls let me know if there is anything I can help with here (like BER thresholds).

Thanks

db4ple:
1. I was waiting for PR#22 changes to materialize before I touch the readmes.
2. Yes, if built and run on an Ubuntu 16.04 machine, all tests pass. If built and run on an Ubuntu 18.04 the tests tst_ofdm_fade, ofdm_ldpc_fade, ofdm_AWGN and ofdm_ldpc_AWGN fail altough the STM32 output is identical. The reason is that the generated reference results for these tests different between 16.0.4 and 18.04. So it is a Linux problem, not an STM32 problem.

donmr:
I think you mean the tests tst_ofdm_demod_... (not mod).

The reference data for those is generated with the PC versions of the
codec2 program ofdm_demod.

For example tst_ofdm_demod_fade creates input with:

ofdm_mod --in /dev/zero --out mod_bits.raw --testframes 60 > setup.log 2>&1
cohpsk_ch mod_bits.raw stm_in.raw -24 -Fs 8000 -f -10 -fast \
	    --raw_dir ${CODEC2_BASE}/raw >> setup.log 2>&1

The the reference data with:

ofdm_demod --in stm_in.raw --out ref_demod_out.raw --log ofdm_demod_ref_log.txt \
        	--testframes --verbose 1 > ref_gen_log.txt 2>&1

It might be useful to check the log files.

I use Mint-19 which is based on Ubuntu 18.04 and these tests work for
me (with st-util).

drowe67:
>     2. Yes, if built and run on an Ubuntu 16.04 machine, all tests pass. If built and run on an Ubuntu 18.04 the tests tst_ofdm_fade, ofdm_ldpc_fade, ofdm_AWGN and ofdm_ldpc_AWGN fail altough the STM32 output is identical. The reason is that the generated reference results for these tests different between 16.0.4 and 18.04. So it is a Linux problem, not an STM32 problem.

This could signal a sensitivity to initial states or small statistical variations.  @donmr do you have any thoughts?  I'm happy to work on these tests if it helps - just let me know.  The aim is to have the test system reasonably independent of the OS variant.

db4ple:
Yes, I meant tst_ofdm_demod_... . May it is important to say that the 16.04 vs. 18.04 is a theory based on the fact that both machines where these tests fail are 18.04 machines. That your Mint19 machines does not have the issue could be a hint that there may (or may not) be a different reason. 
In any case, they generate a different reference, that is clear. The ref_gen_log.txt files have small difference as well.
If I copy the compiled binary from a 16.04 machine to a 18.04 and generate the output, I get exactly the same result as with the binary compiled on 18.04. So the difference must be in one of the libraries (most likely libm) or somewhere else (Processor?).

db4ple:
I followed up on the 18.04 vs. 16.04 issue and I am now certain it is a libm/libc issue. In order to be sure, I statically linked ofdm_demod on my 16.04 machine and transfered it to the 18.04. And voilà, even on the 18.04 we get the indended result, with the dynamically linked executable from the same 16.04 machine, we would produce the "wrong" result. Libc 2.23 on 16.04, Libc 2.27 on 18.04

db4ple:
BTW, may be it could be the reworked math functions, see one of the first entries in the glibc 2.27 release
notes: 
https://lists.gnu.org/archive/html/info-gnu/2018-02/msg00000.html

https://abi-laboratory.pro/?view=changelog&l=glibc&v=2.27 (use this to go quickly to other release notes)

drowe67:
@db4ple nice work :+1: I've got some time off over Easter so (if it's not solved) I will dig into this one with you, I have a Ubuntu 17 and 18 machine handy.

donmr:
BTW it is pretty easy to use Virtual Box and set up a machine with any 
version you want.

drowe67:
@donmr I'm looking into the ```tst_ofdm_demod_AWGN``` log files, after obtaining fails on ```bear```.

1.  Re:
```
BER checks
Reference
BER......: 0.0826 Tbits: 11662 Terrs:   963
BER2.....: 0.0797 Tbits:  6902 Terrs:   550
Target
BER......: 0.0809 Tbits: 11662 Terrs:   943
BER2.....: 0.0793 Tbits:  6902 Terrs:   547
FAIL: BER
```
It's quite OK to have small variations in the BER when the same modem is run on two different platforms.  The demod has feedback loops which will make it hard to obtain bit-exact equivalent results on machines with different precision.  Suggest a 10% tolerance on BER for a PASS.

2.  Re: 
```
Output checks
Output frame 2 byte 1 mismatch: ref=0 tgt=1
Output frame 42 byte 71 mismatch: ref=1 tgt=0
Output frame 45 byte 26 mismatch: ref=0 tgt=1
Output frame 45 byte 167 mismatch: ref=0 tgt=1
FAIL: Output Differences = 462
<snip>
Test FAILED!
```
Are you expecting bit exact location of errors?  As per above, this is unlikely.  A match in  BER is fine. They should also have a similar distribution, for example long error bursts in an AWGN channel in the DUT would be a fail, we would expect the errors to be uniformly distributed across the frame. However that's probably tricky to check.

donmr:
I had been looking at the absolute difference.  Previously all tests passed with an allowance of 0.001.
For this test I had a reference BER of 0.0810 and the same Target of 0.0809.  I'll change it to this:
(((abs(ref_ber - tgt_ber) / ref_ber) <= tolerance_ber).  See PR25

I don't expect an absolute match of the outputs, but I had always seen a small number of differences.  This test was set to accept 2 bytes different.  You are seeing 432 bytes different, that is a big change.

drowe67:
> I don't expect an absolute match of the outputs, but I had always seen a small number of differences. This test was set to accept 2 bytes different. You are seeing 432 bytes different, that is a big change.

True it is a big change from the previous version of Ubuntu. However it may still be a reasonable result when comparing two implementations of the same modem.  The errors will be randomly distributed, and as long as the BER is within tolerance, we wouldn't expect the bit errors to occur in exactly the same position on two different machines.  

It's like rolling two dice at the same time - they'll both come up with 6 about 1/6 of the time, but not often at exactly the same time.

Is it just the position of the bit errors that has changed?  In that case it's OK, and we don't need to log it as a fail.

One possible reason is the random seed or generator algorithm has changed, or some other subtle change in the maths lib (or Octave if we use that to generate noise).

donmr:
I disabled failing due to output differences.  Still I am uncomfortable.  This test generates the input on the PC so the reference and target see the exact same input data.  If using different 'random' data (due to library differences) causes a large change in the test results then we don't have a very good test.  Perhaps we need to use longer data streams?

drowe67:
> I disabled failing due to output differences. Still I am uncomfortable. This test generates the input on the PC so the reference and target see the exact same input data. If using different 'random' data (due to library differences) causes a large change in the test results then we don't have a very good test. Perhaps we need to use longer data streams?

With a longer stream I'd expect bit-exact tests like the error position to diverge but the BER statistic (a mean) to converge.  It is puzzling that the results have changed with changing OS, but I am comfortable with the tests and results (although Ihaven't explored the other tests like fading). 

Feel free to dig into what's changed, e.g. plotting vectors.  With fixed point work it's easier to get bit exact results.

donmr:
I think this PR added the "monitor adapter_khz" gdb commands.  Those don't work with st-util which is fine with me.  I suggest we make openocd the standard without requiring an option.

PR #25, opened 2019-04-17
Change BER checks to a ratio of error/reference.

donmr:


drowe67:
Thanks for those changes Don.  I'm testing them on ```bear```:
```
david@bear:~/codec2/stm32/unittest/scripts$ git pull
git checkout donmr-PR23-24
./run_all_ofdm_tests
<snip>
7 ODFM Tests FAILED!
```
Can you pls confirm the tests run OK for you (i) on yr machine and (ii) on bear?

donmr:
Arrgh, the new code was dividing by ref_ber which is zero in some ideal cases.

drowe67:
> Arrgh, the new code was dividing by ref_ber which is zero in some ideal cases.

Thanks for finding that bug Don - can you pls confirm (i) and (ii) in my question above?

donmr:
I don't see a question with a (i) or (ii) in it.  Was that in another thread??

drowe67:
> I don't see a question with a (i) or (ii) in it. Was that in another thread??
> All of the OFDM tests pass for me with this branch.

This comment above https://github.com/drowe67/codec2/pull/25#issuecomment-484694696

"Can you pls confirm the tests run OK for you (i) on yr machine and (ii) on bear?"

donmr:
It all works fine on my system.  It won't compile on bear, with the problem we have seen before.
First is this:

/usr/bin/objcopy: Unable to recognise the format of the input file `sm2000_adcdump.elf'
CMakeFiles/sm2000_adcdump.dir/build.make:463: recipe for target 'sm2000_adcdump.elf' failed

The same error occurs for the unittests too.

I used this cmake command

cmake .. \
    -DCMAKE_TOOLCHAIN_FILE=${PWD}/../cmake/STM32_Toolchain.cmake \
    -DPERIPHLIBDIR=/home/don/STM32F4xx_DSP_StdPeriph_Lib_V1.8.0

Everything is in /home/don/codec2

db4ple:
@donmr: Please try to place the arm gcc path BEFORE the old path in .profile:
PATH=$HOME/gcc-arm-none-eabi-8-2018-q4-major/bin:${PATH}
It may help cmake to find the right objcopy...

srsampson:
Might this be a good time to move the sm2000 stuff to the attic?

db4ple:
Hi,
I will open a new issue where can collect "outdated" stuff for deletion or replacement , I have a couple of things, too. But this PR is not the best place to discuss this...

donmr:
There is only one arm-none-eabi-objcopy found,

/home/don/gcc-arm-none-eabi-8-2018-q4-major/bin/arm-none-eabi-objcopy

It looks like cmake using "objcopy" instead?  That will not work with any PATH setting.

db4ple:
Don, this is basically the only difference between your and my settings on the same machine. Maybe cmake looks for both objcopy and arm-non-eabi-objcopy. Just give it a try, If this doesn't work, we have to dig deeper...

db4ple:
@donmr: Just tested both variants and you are right, it makes no difference. However, in both cases I get the correct objcopy set in CMakeCache.txt.
Please delete the whole contents of build_stm32 and run cmake again.

donmr:
Well that works so something odd is happening.  But we know how to avoid it.
The next obstacle is finding st-util on bear.  My branch is based on master which does not have openocd support  yet.

db4ple:
Well, if you accept the stm32 debugging PR, we will have openocd in the master. And then you run a rebase for your branch on master, then you don't need to find st-util ;-)

donmr:
We are stuck in a loop, David won't merge Danilo's PR until the tests pass.  They won't pass until my changes to the check script are included.  He won't accept those changes until the tests pass on Bear.  That won't happen until Danilo's PR is merged.........

db4ple:
The tests will/are passing. I found my mistake in Travis and in a couple of minutes (more like an hour), we should have a passing run of Travis for my changes for PR #24 which includes all but one STM32 test. And the only test not included isn't there because of extreme runtime without output, which makes travis think, the build is dead, and then the travis run is killed.

db4ple:
BTW, in the meantime we just have to build st-util from source or use my build ~danilo/git/st-util/...

donmr:
I spoke too soon about tests passing on Bear.  Even with another code fix, some tests are too differenent.
In tst_ofdm_demod_ldpc_AWGN I see this:
BER checks
Reference
BER......: 0.0778 Tbits: 39200 Terrs:  3050
Coded BER: 0.0013 Tbits: 19600 Terrs:    26
Target
BER......: 0.0820 Tbits: 39872 Terrs:  3268
Coded BER: 0.0062 Tbits: 19936 Terrs:   124
FAIL: BER

On my system the numbers are almost identical!

Also the output is a different length which is currently a failure, but maybe should not be??

drowe67:
> I spoke too soon about tests passing on Bear. Even with another code fix, some tests are too differenent.
> In tst_ofdm_demod_ldpc_AWGN I see this:
> BER checks
> Reference
> BER......: 0.0778 Tbits: 39200 Terrs: 3050
> Coded BER: 0.0013 Tbits: 19600 Terrs: 26
> Target
> BER......: 0.0820 Tbits: 39872 Terrs: 3268
> Coded BER: 0.0062 Tbits: 19936 Terrs: 124
> FAIL: BER
> 
> On my system the numbers are almost identical!
> 
> Also the output is a different length which is currently a failure, but maybe should not be??

Thanks for all your hard work guys - I appreciate this is demanding work.

Yes I'd call that results a pass.  The different output length could be caused by one of the modems-under-test getting sync a little before the other.  The LDPC errors are often concentrated in the period just after sync, where the residual freq offset is quite high.  Just one or two frames where the LDPC decoder breaks down can cause a big difference in the coded BER.  This test is right on the "knee" in the LDPC decoders BER curve.

For the ofdm modem Ctests I placed this criteria in ```ofdm_demod.c```:
```
            if ((uncoded_ber >= 0.1f) && (coded_ber >= 0.01f))
                return 1;
```
Hmm, actually that ``&&``` should be a ```||```.

drowe67:
> Hi,
> I will open a new issue where can collect "outdated" stuff for deletion or replacement , I have a couple of things, too. But this PR is not the best place to discuss this...

Good idea. But first lets:

1. Get the stm32 tests passing on Ubuntu 18 (bear) as well as other machines
2. Get this PR merged with https://github.com/drowe67/codec2/pull/24
3. Merge https://github.com/drowe67/codec2/pull/24 with master

Too many balls in the air .... (I'm also juggling FreeDV 2020 integration work, FreeDV 700C bugs, and LPCNet R&D)

db4ple:
I think PR #24 and #25 are independent and can be merged independent from each other. PR# 24 is about get the the unittest environment ready for OpenOCD and remote testing. #24 does not change the test results compared to the STM32 testing without that PR. 

PR #25 is about getting the STM32 tests more "relaxed" related to what is considered acceptable differences, which at the moment appear on Ubuntu 18.04 only. This involves changing the test result checking but has nothing to do with the unittest infrastructure as such. 

Technically there is no overlap in files we changed in the two PRs. So, of course it would be easy to wait and merge both but in the same way Don could just base his work on the master branch which incorporates the improved testing infrastructure. From this your 2020 work could benefit, too. And to be honest, I woul have to worry that another PR interfering with my little PR.

drowe67:
> Technically there is no overlap in files we changed in the two PRs. So, of course it would be easy to wait and merge both but in the same way Don could just base his work on the master branch which incorporates the improved testing infrastructure. From this your 2020 work could benefit, too. And to be honest, I woul have to worry that another PR interfering with my little PR.

Fair points Danilo.  I'll go ahead and merge https://github.com/drowe67/codec2/pull/24 with master.

db4ple:
@drowe67 : Thanks!

drowe67:
@donmr on the stm32 speech codec tests, just 1 sec of hts1a would be fine to show if the speech codec is behaving, so 8000 x 2 byte samples dd-ed at the start.  That should speed up the execution time of the tests.

Be interesting to see how these tests go when run from different operating systems.  I had trouble getting bit exact results between x86 and stm32 when I initially ported codec2 to the stm32.  One issue was different random number generators.

drowe67:
@donmr @db4ple would you like me to work on getting these tests to pass on Ubuntu 18/bear?  I'm keen to close out this PR, and use the tests as a gateway for other merges.

@db4ple will the stm32 ```make test``` run using openocd be default?

donmr:
If you can figure out what is going on that would be great.

I run Mint 19 here which is based on Ubuntu 18 already so I don't understand
why there is such a difference.  I have been busy with other things but
my next plan is to set up a virtual box here with as close as I can get
to Bear's software then experiment to see what causes the differences.

Don

db4ple:
I just wanted to summarize my experiments:
I made two ofdm_demod binaries both compiled on Ubuntu 16.04 one dynamically linked, one statically linked (ofdm_demod_16_static, ofdm_demod_16_dynamic):

If run on Ubuntu 16.04:  Both binaries generate identical output for the tests failing on 18.04

If run on Ubuntu 18.04: ofdm_demod_16_static still produces same output as on Ubuntu 16.04. However, ofdm_demod_16_dynamic (which still uses the libcodec2.so compiled on 16.04) produces different output, which makes the tests fail. 

Analysis: Since the libcodec2.so of ofdm_demod_16_dynamic is the one from Ubuntu 16.04 the problem must be in one of the other dynamic system libraries. Of which libm is the hottest candidate. BTW, the output of ofdm_demod binaries compiled on Ubuntu 18.04 (static and dynamic) is identical to the output of the ofdm_demod_16_dynamic running on Ubuntu 18.04.

Finding out why the difference exists is a tedious process. One approach would be to identify the first different output data item and trace its "production" through the code and see where are difference in data used on the way. There must be a point where internal data is diverging if run on 16.04 and 18.04.
For that two "machines" one 16.04, one 18.04 are necessary. The good thing is, it is not related to the STM32, just code on linux boxes.

drowe67:
> If you can figure out what is going on that would be great. 

Actually my intention was to modify the test criteria along the statistically significant lines i have suggested above.  However if you guys want to run the Ubuntu 16/18 differences to ground, pls feel free.

db4ple:
I think, modifying the test is generally the better approach as we will probably be able to see where the difference it if digging deep enough. But not able to do much about it.

donmr:
The text output of ofdm_demod is different:
reid1 (and on a new ubuntu 18.04.2 VM):
   ct_est:  420 foff_est: -17.1 timing_valid: 1 timing_mx: 0.3694
   ct_est:    0 foff_est: -14.4 timing_valid: 1 timing_mx: 0.4529
   ct_est:    0 foff_est: -14.1 timing_valid: 1 timing_mx: 0.4357
   ...
Bear:
   ct_est:  420 foff_est: -17.1 timing_valid: 1 timing_mx: 0.3694
   ct_est:    0 foff_est: -14.0 timing_valid: 1 timing_mx: 0.4492
   ct_est:    0 foff_est: -14.0 timing_valid: 1 timing_mx: 0.4478

Adding the -log option to ofdm_demod produces files with significant differences too. First at line 33
 1.032903 1.375566 1.081644 1.799899 1.196983 1.485706 -0.670861 -0.356490 -0.703662 -1.039466 -1.924524 -2.603291 -2.615824 -2.924814 2.553222 2.132817 1.930354
vs
 -0.272247 -0.422774 -1.007485 -1.792496 -2.508064 2.658291 -3.051869 2.828238 -3.038302 1.724004 1.664316 1.514999 1.606130 1.188409 0.196036 -0.566545 -1.529124

db4ple:
Just a thought: Are reid1 and the VM running 32bit Ubuntu?

drowe67:
One other way these differences can crop up is borderline round off errors, like floor(1.9999/2) compared to floor(2.0000/2).  @srsampson and I hit this a few weeks ago when calculating the number of carriers for 2020 waveforms in ```ofdm.c```, I think @donmr came up with the fix.

donmr:
No, x86_64.

Linux reid1 4.15.0-47-generic #50-Ubuntu SMP Wed Mar 13 10:44:52 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

Linux don-ubu18-xfce 4.18.0-17-generic #18~18.04.1-Ubuntu SMP Fri Mar 15 15:27:12 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

Linux bear 4.15.0-20-generic #21-Ubuntu SMP Tue Apr 24 06:16:15 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

All 3 systems have the same version of the libc6 package which supplies the libm library.

db4ple:
That is very, very interesting. What CPU do you have? Bear has a rather vintage Core 2 Duo, my Ubuntu 18.04 is also running a Core 2 Duo (even a little older). I guess your machine's processor is a little newer. 
Maybe the optimizations in the glibc 2.27 produce a little different results on very old processors... I know this is a wild guess.

donmr:
On Tue, Apr 23, 2019 at 03:35:30PM -0700, db4ple wrote:
> That is very, very interesting. What CPU do you have? Bear has a rather vintage Core 2 Duo, my Ubuntu 18.04 is also running a Core 2 Duo (even a little older). I guess your machine's processor is a little newer. 

Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz

> Maybe the optimizations in the glibc 2.27 produce a little different results on very old processors... I know this is a wild guess. 

It's not old enough to have the floating point bugs is it??

db4ple:
We can test this idea in a quite simple way:
We build a statically linked ofdm_demod on Bear. If this very binary produces different output on your (newer) machine, it most likely is a processor thing (since both run the same software).

donmr:
Sounds good, do you have the recipe for a static version figured out?

db4ple:
Sure, there is somewhere in the CMakeFile.txt a line BUILD_SHARED_LIBS ON, you set this to OFF
and you have to change the linking to static by a trick.
- Clean out build directory (rm -rf *)
- Change the shared lib setting in CMakeFiles.txt to OFF
- cmake ..
- edit src/CMakeFiles/ofdm_demod.dir/link.txt -> -rdynamic => -static
- make ofdm_demod
Or you simply use ~danilo/ofdm_demod18.static on bear.

donmr:
Both the 16 and 18 static versions of ofdm_demod from bear produce the same output on reid1 as my locally compiled version.

db4ple:
And on bear, they generate different output. I just tested this. One does not even need to compare
the output to see there is a difference (the static variants have a number for their origin): 

-rw-rw-r-- 1 danilo danilo   96628 Apr 24 15:01 ref_demod_out16.raw
-rw-rw-r-- 1 danilo danilo   96866 Apr 24 15:01 ref_demod_out18.raw
-rw-rw-r-- 1 danilo danilo   96866 Apr 22 02:15 ref_demod_out.raw
-

drowe67:
Some thoughts:

1. I wonder if it's an un-initialised variable somewhere?
1.  Interesting that the tofdm test passes on bear.  This compares the C version to Octave by comparing vectors (very closely, in some cases to 1E-3 tolerance).  Although it just runs for a few 10's of frames, as things tend to diverge with feeback loops and different precision machines.
1. Try raising the SNR and see if you get a better match. If its a difference in say a trig library function implementation it might be outputting pi-0.00001, instead of -pi-0.0000001, especially with noise.
1. Try plotting the modem states from the two machines - like tofdm does.  In fact, tofdm dumps Octave files, you could compare them with a hacked version of tofdm that compares the two C versions, rather than C and Octave.  That's how I sniff out these sorts of problem between C and Octave.

donmr:
I realized that the changes I had made to tst_ofdm_demod_check were not right.  David's comments on allowed BER were absolute BER rates not difference between the referance and target.

New strategy, add tests for max BER from target and make comparisons to reference optional.  Disable comparisons on tests with large channel degredation.

I would still like to understand what is happening on Bear, but that is not actually an STM32 issue the differences are in the PC code used for the reference.  With these changes the tests should pass on Bear (I have checked several but not re-run the STM32 passes).

drowe67:
@db4ple I'm building the stm32 code and noted cohpsk.c is being built.

That's just used for FreeDV 700C so can be removed from the stm32 build if we want.  I don't think it has any dependencies we need.  Also fsk.c., fmfsk.c., freedv_vhf_framing.c, freedv_data_channel.c.  

I guess these all get dropped by the linker?  So maybe no big deal.

drowe67:
OK just tested latest sm1000.bin.  Mores prompts are nice, but when tested on a 1600 waveform Im getting:
1. Clip light blinking all the time
1. Poor audio quality, like there are bit errors.
1. colourful ring of death after 10-20s
Will poke about a bit ....

donmr:
A few weeks ago I was able to run it in 1600 mode but only tried for the 
length of a sample clip.

db4ple:
@drowe67 : If there is no dependency, the linker will totally take care of this. But there is one, freedv_comptx_700D uses cohpsk_clip.
But again, this is not a big deal. We are using function and data object level linking, which means, it is just pulling in cohpsk_clip (and all it references) but not all of cohpsk.c. I checked sm1000.map and indeed, only cohpsk_clip is in the binary. 
Could be that the clipping is not a "cohpsk" function but rather a general one, not for me to decide. Your call... Once that is done, we could indeed let go of cohpsk.c in particular, but it does not really impact compilation speed too much...

drowe67:
@donmr @db4ple what do we need to do to close out this PR?  

Quite a lot of files here, do they need to be merged?

I presented some options for handling the errors were discussed here https://github.com/drowe67/codec2/issues/26#issuecomment-486880821

I'd like to move on with 700D integration with the SM1000 ....

donmr:
I think all of mu changes got put back the way they were except tst_ofdm_demod_check.  I can make a separate branch/PR with just that if it is easier (or safer).

drowe67:
> I think all of mu changes got put back the way they were except tst_ofdm_demod_check. I can make a separate branch/PR with just that if it is easier (or safer).

Thanks Don.  So just to confirm - you are suggesting we close this PR without any merging?  We can leave the branch so the code won't be lost f needed in the future.

@db4ple pls let me know if that works for you too.

donmr:
That is just for my changes.  Most of the changes we Danilo's.

Don

drowe67:
@donmr can this PR be closed now?

donmr:
I still want the changes to stm32/unittest/scripts/tst_ofdm_demod_check which I think is all you will get after Danilo rebased this.

drowe67:
Thanks Don.  So @db4ple to merge the code Don wants should we close this PR and merge https://github.com/drowe67/codec2/pull/38?

db4ple:
Since no more commits have been made since my rebase #38 is the easiest path the get this done.

PR #27, opened 2019-04-23
work in progress branch for freedv 2020 integration

drowe67:


drowe67:
Not required any more as codec2/master is working with LPCNet

PR #28, opened 2019-04-23
STM32 Unittesting performance fixes

db4ple:
- After identifying the OpenOCD issue which broke buffered fread for us,
buffers help us to speed up testing significantly. We use CCM for
our I/O buffers so that the normal heap size for the application is not
reduced.
- We only use semihosting_fileio for OpenOCD if we are running remote
testing via ssh since it requires not yet mainstream openocd fixes.

Runtime for the STM32 test is about half of the time it took before these changes.
The results are identical.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ install:
 - export STDLIBDIR=$HOME/STM32F4xx_DSP_StdPeriph_Lib_V1.8.0
 - if [ ! "$STDLIBURL" == "" ]; then pwd; wget -q $STDLIBURL; 7z  x -p$STDLIBKEY -o$HOME
   stdperiph_lib.zip; fi
+- unset STDLIBURL STDLIBKEY # DO NOT TOUCH  
 script:
 - mkdir -p $BUILDSTD
 - cd $BUILDSTD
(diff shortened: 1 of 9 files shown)


donmr:
Yes I like it too, thanks for figuring out how to use CCM effectively.  We may need it on the SM1000.

db4ple:
We must keep in mind that without additional changes to the linker script and the startup code only uninitialized memory is provided in the CCM. Even if the compiler without warnings accepts global variables (which normally would be zeroed at startup)!

donmr:
I think access this CCM will be a bit slower too.  It is intended for code and must be accessed by via DCODE bus which will interfere with fetching code and constant data from the flash memory.  But that should not be an issue for IO buffers.

db4ple:
@drowe67 : The access to CCM will most likely not be slower since we are fetching our code from Flash and as far as I understood the cortex architecture, is a different path which can be running in parallel. Anyways, the differences will be neglectable, I'd say. But for our specific use case, it does not matter at all.

drowe67:
> I think access this CCM will be a bit slower too. It is intended for code and must be accessed by via DCODE bus which will interfere with fetching code and constant data from the flash memory. But that should not be an issue for IO buffers.

Perhaps also useful for the real time I/O buffers, DMA etc

drowe67:
so OK to merge @donmr ?

donmr:
Yes.

db4ple:
BTW, DMA is not possible in CCM. The DMA controller(s) cannot access the CCM.
Have a look at the schematic shown here: https://vjordan.info/log/fpga/first-steps-with-the-stm32f4.html, taken from the reference manual.
You'll notice it is very, very close to the actual core and is not reachable via the bus matrix.

Issue #29, opened 2019-04-23
700D failing to get interleaver sync


drowe67:
I have an off air FreeDV 1.3.1 700D sample that won't decode using codec2/master & freedv-gui/master built from Git Head.  It gets modem sync but not interleaver sync.

Suspect this is related to the work @donmr and myself did around low memory 700D optimisations for the stm32 in freedv_api.c, I have done some work on this in the codec2/brad-2020 and codec2/dr-2020 branches but it's still not decoding successfully.

We'll need to fix this before another FreeDV release.  A good start would be back tracking through Git revisions and perhaps testing with freedv_rx.

db4ple:
David, I guess you know `git bisect`. For those who don't: It helps you to find the git commit which broke something quickly. You need to give it one known to  work commit (ideally not too old) and will checkout commits and ask if they are good or bad. Often needs only a few steps. Sometimes fails.

drowe67:
Thanks Danilo.  Another idea is to cross check and earlier freedv_tx with a current freedv_rx to confirm the bug exists at the API level.  I'll think about it some more.

drowe67:
OK the culprit is me in 20d24ed2b, we changed the tuning of the 700D modem by one carrier I would guess

drowe67:
All fixed now

PR #30, opened 2019-04-24
Fix 700D to be compatabile with 1.3.1

drowe67:
Frequency offset of one carrier as per https://github.com/drowe67/codec2/issues/29, also added a Ctest for backwards compatibility of 700D.


drowe67:
Thanks @srsampson, however I still can't get freedv-gui to get sync on that file Mike sent us, or my other, earlier 700D test files.  However freedv_rx works OK.

Could be I'm using the wrong shared lib ... will keep trying.

drowe67:
Ahh OK, I forgot that complex input to freedv API 700D was broken in master.  When I merged this branch with codec2/dr-2020 freedv-gui now works OK.  Time to get a few of these branches merged me-thinks

PR #34, opened 2019-05-02
Brad 2020

drowe67:
Re-opened a PR here as this branch is being used for further CMake tuning - thanks @hobbes1069


drowe67:
@donmr @db4ple I'd like to merge this into master, can you pls tell me if the stm32 tests pass OK?
@hobbes1069 I figure this is ready for merge, but just et me know if its still a WIP.

db4ple:
@drowe67 : `docker logs docker_codec2_ub1604_s_run_6 | less` in a couple of minutes will tell you the stm32 ctest result.

hobbes1069:
Yup! I'm pretty much done for now unless we run into issues. I need to find some cycles for a few other projects :)

donmr:
Looks good from here, my tests pass.

src/ofdm.c
```diff
@@ -942,8 +942,9 @@ int ofdm_sync_search(struct OFDM *ofdm, COMP *rxbuf_in) {
      * so it is primed for when we have to call ofdm_demod()
      */
 
-    memcpy(&ofdm->rxbuf[0], &ofdm->rxbuf[ofdm->nin],
-        (ofdm_rxbuf - ofdm->nin) * sizeof (complex float));
+    /* note can't use memcpy when src and dest overlap */
+    memmove(&ofdm->rxbuf[0], &ofdm->rxbuf[ofdm->nin],
+           (ofdm_rxbuf - ofdm->nin) * sizeof (complex float));
```
drowe67: @srsampson Had some sync issues on 2020, and noticed use of memcpy on overlapping buffers/. Just happened to hit this one at $dayjob recently.  Subtle

srsampson: I think memcpy should be deprecated throughout the repository in favor of memmove. My thinking is that these two functions are optimized in the compiler, and will be faster than a for loop, but I've never actually looked at the output code.

drowe67: It's a very subtle one .... I've used it without considering overlap for years.  Took me weeks to find it at $dayjob, as it would only come up on occasional Travis runs.  I figure just use for loops.  There aren't many memory copies that are significant in terms of CPU on modern machines.

memmove adds another buffering step so it's not great either.

drowe67:
FreeDV 2020 is maturing, so just a heads up that I'll merge this branch some time this week.

PR #37, opened 2019-05-07
TTY for SM1000

drowe67:
Trying to build a simple. low memory way to printf debug messages (such as memory used) from the SM1000.  A USB comm port would be idea, and we have source for that, but I am concerned about run time memory usage.

### DIFF
--- a/stm32/CMakeLists.txt
+++ b/stm32/CMakeLists.txt
@@ -296,6 +296,19 @@ add_mapped_executable(pwm_ut ${PWM_UT_SRCS})
 target_link_libraries(pwm_ut stm32f4 sm1000base)
 elf2bin(pwm_ut)
 
+# ------------------------------------------------
+
+set(USART_UT_SRCS
+src/usart_ut.c
+)
+
+add_mapped_executable(usart_ut ${USART_UT_SRCS})
+target_link_libraries(usart_ut stm32f4 sm1000base)
+elf2bin(usart_ut)
+
+# ------------------------------------------------
+
+
 set(POWER_UT_SRCS
 src/power_ut.c
 )
--- a/stm32/CMakeLists.txt
+++ b/stm32/CMakeLists.txt
@@ -299,6 +299,7 @@ elf2bin(pwm_ut)
 # ------------------------------------------------
 
 set(USART_UT_SRCS
+src/stm32f4_usart.c
 src/usart_ut.c
 )
 
--- a/stm32/README.md
+++ b/stm32/README.md
@@ -46,8 +46,7 @@
    $ cmake /path/to/codec2-dev/stm32 -DUT_PARAMS=--openocd -DCMAKE_TOOLCHAIN_FILE=/path/to/codec2-dev/stm32/cmake/STM32_Toolchain.cmake -DPERIPHLIBDIR=/path/to/unzipped/STM32F4xx_DSP_StdPeriph_Lib_Vx.x.x  ..
    ```
    
-   
-6. Build binairies
+6. Build binaries
 
    Finally:
    ```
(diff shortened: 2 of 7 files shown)


drowe67:
@donmr I have a basic unit test compiling but not tested yet.  Feel free to work on this if it is of any interest (you may have other plans for SM1000 debug tools).  I'll move it forward over the next few weeks.

donmr:
A simple puts that is mostly called with strings from the ROM and a form of 
itoa can do a lot with just a few bytes of static memory.

Don

drowe67:
> A simple puts that is mostly called with strings from the ROM and a form of itoa can do a lot with just a few bytes of static memory. Don

Exactly - so we could have it 'counting down" the available stack and heap as the various modules come up.  Add a sentinel to the end of stack/heap and monitor that.

drowe67:
@db4ple I have the basic unit testing going but my baud rate is 3030 rather than 9600. I think #define HSE is set to 25MHz rather than 8MHz when the STM library is built.  Might be able to set this at build time.

db4ple:
@drowe67 

Found an interesting piece of code:

https://github.com/drowe67/codec2/blob/5866cb46ec844dfa3178156bdbddd53b5abda757/stm32/inc/stm32f4xx_conf.h#L26-L30

This set the HSE_VALUE to 80000000 if and only if someone set the HSE_VALUE to something.
If HSE_VALUE is not set, it will define nothing and we get the default 25 Mhz....
In a nutshell: if we define HSE_VALUE (regardless of the value), we will assume 8 Mhz clock. If not, we assume 25Mhz clock.
That does not make sense.
Either, 
1. we use 8 Mhz if nothing is defined and the defined value if HSE_VALUE is set, OR
2. we use 8 Mhz all the time.

But the current code is stupid. I vote for 1.

```c
#if !defined  (HSE_VALUE)
  /* Define the HSE value; it's equal to 8 MHz on the STM32F4-DISCOVERY Kit */
 #define HSE_VALUE    ((uint32_t)8000000)
#endif /* !HSE_VALUE */

drowe67:
Good catch. 

However I'm not sure if https://github.com/drowe67/codec2/blob/5866cb46ec844dfa3178156bdbddd53b5abda757/stm32/inc/stm32f4xx_conf.h#L26-L30 is included when we build the stm peripheral lib.  The USART functions are deep inside the peripheral lib, and call RCC_ functions that need the compiler time value of HSE_VALUE.

I've just pushed a fix to this PR, see what you think.

db4ple:
Well, this should work in any case due to the nature of the problem: if inc/stm32f4xx_conf.h is used, it will define 8Mhz clock since you set the HSE_VALUE. If not, the define HSE_VALUE will do the intended job.

db4ple:
HSE_VALUE is only used in the https://github.com/drowe67/codec2/blob/5866cb46ec844dfa3178156bdbddd53b5abda757/stm32/src/system_stm32f4xx.c
(and in the usb_vcp_ut.c) but not in the StdPeriphlib directly. The mentioned file has the code to initialize the STM32 clocks.

drowe67:
@donmr @db4ple I have a small usart library working now, which can do things like usart_puts("Hello World\n").  it's running on bear atm, just login and ```sudo tio -m INLCRNL /dev/ttyUSB0```. Feel free to modify the program with openocd/gdb if you like.

@db4ple - I found a stm32 peripheral library example ```USART_printf``` where they could redirect the printf output to a custom puts() function, but couldn't get it to work.  I imagine the default is weakly linked.  I'd be happy to try this approach again if you have any thoughts on how to get it working.

drowe67:
@donmr and @db4ple if you think this is OK lets merge it.

donmr:
I like it, it's working on an SM1000 board.   There is something that is changing the first byte after a boot to 0xFF but later output is fine.

PR #40, opened 2019-05-16
Add IQ input option for horus_demod.

ED6E0F17:
slightly revised from:
https://github.com/projecthorus/horusbinary/pull/4

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,6 +338,13 @@ if(UNITTEST)
              set_tests_properties(test_horus_binary PROPERTIES
              PASS_REGULAR_EXPRESSION "1C9A9545")
 
+     # same, but with audio as stereo
+     add_test(NAME test_horus_binary_mode_q
+             COMMAND sh -c "sox ${CMAKE_CURRENT_SOURCE_DIR}/wav/horus_binary_ebno_4.5db.wav -c 2 -r 48000 -t raw - | $<TARGET_FILE:horus_demod> -q -m binary - -"
+             )
+             set_tests_properties(test_horus_binary_mode_q PROPERTIES
+             PASS_REGULAR_EXPRESSION "1C9A9545")
+
     # tests ldpc_enc/ldpc_noise/ldpc_dec
     add_test(NAME test_ldpc_enc_dec
              COMMAND sh -c "$<TARGET_FILE:ldpc_enc> /dev/zero - --sd --code HRA_112_112 --testframes 200 | $<TARGET_FILE:ldpc_noise> - - 0.5 | $<TARGET_FILE:ldpc_dec> - /dev/null --code HRA_112_112 --sd --testframes"
--- a/src/horus_api.c
+++ b/src/horus_api.c
@@ -348,11 +348,45 @@ int extract_horus_binary(struct horus *hstates, char hex_out[], int uw_loc) {
     return hstates->crc_ok;
 }
 
-
 int horus_rx(struct horus *hstates, char ascii_out[], short demod_in[]) {
+    int i, result;
+
+    assert(hstates != NULL);
+
+    /* Note: allocating this array as an automatic variable caused OSX to
+        "Bus Error 10" (segfault), so lets malloc() it.
+        ( Maximum nin based on horus_get_max_demod_in() is N + 2*Ts ) */
+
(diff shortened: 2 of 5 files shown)


src/horus_api.c
```diff
@@ -348,8 +348,11 @@ int extract_horus_binary(struct horus *hstates, char hex_out[], int uw_loc) {
     return hstates->crc_ok;
 }
 
-
 int horus_rx(struct horus *hstates, char ascii_out[], short demod_in[]) {
+    return horus_rx_iq(hstates, ascii_out, demod_in, 0);
+}
+
+int horus_rx_iq(struct horus *hstates, char ascii_out[], short demod_in[], int quadrature) {
```
drowe67: Pls use the convention ```_comp``` rather than ```_iq```, to be consistent with the rest of codec 2.

src/horus_api.h
```diff
@@ -51,6 +51,7 @@ uint32_t      horus_nin   (struct horus *hstates);
 /* returns 1 if ascii_out[] is valid */
       
 int           horus_rx    (struct horus *hstates, char ascii_out[], short demod_in[]);
+int           horus_rx_iq (struct horus *hstates, char ascii_out[], short demod_in[], int quadrature);
```
drowe67: Could this be just ```horus_rx_comp (struct horus *hstates, char ascii_out[], short demod_in[])'``` ?  We probably don't need the flag at the API level.

drowe67:
Thanks for your PR, some minor edits suggested above.

Could you pls add a Ctest that exercises this new option?  We need some way of knowing that it works.  Something like https://github.com/drowe67/codec2/blob/master/CMakeLists.txt#L335 that demodulates a short sample file would be fine.

ED6E0F17:
Adding a Ctest would be a very good idea, and if I can clean up the code I will.

ED6E0F17:
Partial success - I retained the "_iq" suffix  for an audio stream variable name before the conversion to a floating point complex array.
.
The CTest is less functional than it might have been with a genuine IQ stereo sample, but it is sufficient to prove that the two formats are giving the same result.

src/horus_api.c
```diff
@@ -368,24 +399,12 @@ int horus_rx(struct horus *hstates, char ascii_out[], short demod_in[]) {
...
-       "Bus Error 10" (segfault), so lets malloc() it.  TODO: A real
-       short sample option for fsk_demod() would be useful */
-    
-    COMP *demod_in_comp = (COMP*)malloc(sizeof(COMP)*hstates->fsk->nin);
-    
-    for (i=0; i<hstates->fsk->nin; i++) {
-        demod_in_comp[i].real = demod_in[i];
-        demod_in_comp[i].imag = 0;
-    }
     fsk_demod(hstates->fsk, &hstates->rx_bits[rx_bits_len-Nbits], demod_in_comp);
```
drowe67: This code was reqd to prevent problems with the OSX build, as per the comments.  It would be wise to keep it in horus_rx() above.

@darksidelemm - you debugged this problem originally, what do you think?

CMakeLists.txt
```diff
@@ -338,6 +338,13 @@ if(UNITTEST)
              set_tests_properties(test_horus_binary PROPERTIES
              PASS_REGULAR_EXPRESSION "1C9A9545")
 
+     # same, but with audio as stereo
+     add_test(NAME test_horus_binary_mode_q
+             COMMAND sh -c "sox ${CMAKE_CURRENT_SOURCE_DIR}/wav/horus_binary_ebno_4.5db.wav -c 2 -r 48000 -t raw - | $<TARGET_FILE:horus_demod> -q -m binary - -"
+             )
+             set_tests_properties(test_horus_binary_mode_q PROPERTIES
+             PASS_REGULAR_EXPRESSION "1C9A9545")
+
```
drowe67: This tests that the code runs but simply copying the I to the Q doesn't test the ability of this option to demodulate complex valued signals, it's a real valued signal input.

src/horus_api.c
darksidelemm: Yes, that should probably be kept. Caused a lot of grief.

drowe67:
@ED6E0F17 could you pls fix the Ctest for the IQ option?

PR #41, opened 2019-06-14
OFDM Sync

drowe67:
Taking another look at OFDM modem acquisition and sync, in particular on HF channels.  I have some anecdotal reports (from the UK and VK5) and a sample suggesting 700D sync is very slow (around 10s) on fading HF channels.  The problem occurs even with high SNRs.

I suspect the averaging of the freq offset metric in ofdm.c:
```
ofdm->foff_metric = 0.9f * ofdm->foff_metric + 0.1f * (conjf(p1) * p2 + conjf(p3) * p4);
```
 This averaging was added late in 700D development, and may be doing strange things on real world channels.  For example an impulse (static crash) could add a large bias to the metric memory that takes a while to fade away.  I'm not sure the ofdm_dev.m acquisition tests exercise the averaging effect correctly, as they use the same freq offset for each test.

1. I have run the sample in freedv-gui and simulation and indeed sync seems to get stuck for about 10s. Pressing ReSync on the signals several time helps see the effect.
1. When I disable the averaging of the freq offset statistic the situation improves, but it can still take a few seconds, however it doesn't appear to get stuck.
1. I have re run the ofdm_dev.m acqusition tests (without averaging) and can see the probability of correct freq estimate to be about 0.5 on the 1Hz Doppler HF channel tested, even at high SNRs.  This suggests that sync will occasionally take a few seconds on fast fading HF channels. 

OK found a problem with slow fading channels.  Freq selective notches bias the estimator away from the +/-1.5Hz range the OFDM modem can handle.  In slow fading channels, the notches can last for several seconds.  In fast fading, the notches come and go faster, and the freq estimates are more evenly distributed, so we sync faster.

Developed a new estimator using pilot correlation.  Less noisy estimates, better P(freq) curves, but still suffers from bias with freq sel fading, and gets stuck with slow fading.

Added high/low bandwidth phase estimator, that switches gears using the state machine.  The high bandwidth can handle freq offset estimator errors better, so we have a solution for the slow fading problem.  Low bandwidth phase est maintains our low SNR perf once synced up.

Added parallel timing estimators at different frequency offsets to the modem can handle +/- 60Hz freq offsets.

Optimised for real time operation on the stm32.


src/ofdm.c
```diff
@@ -738,7 +738,7 @@ static float est_freq_offset(struct OFDM *ofdm, complex float *rx, int timing_es
      * improve estimate.  Small real 1E-12 term to prevent instability
      * with 0 inputs.
      */
-    ofdm->foff_metric = 0.9f * ofdm->foff_metric + 0.1f * (conjf(p1) * p2 + conjf(p3) * p4);
```
drowe67: freq statistic averaging

srsampson:
I added a small optimization. Using the macro is for stm. Hopefully my math outside of loop didn't blow it up.

drowe67:
> I added a small optimization. Using the macro is for stm. Hopefully my math outside of loop didn't blow it up.

Thanks Steve, good idea!  I was hoping you would take a look at this, was going to ask you to review.

srsampson:
> Thanks Steve, good idea! I was hoping you would take a look at this, was going to ask you to review.

There is one more optimization (reducing a function call) but I didn't add it as it might obfuscate what is being done):

```
            complex float w = cmplxconj(tmp * i);
            C_st = C_st + corr_st[i] * w;
            C_en = C_en + corr_en[i] * w;
```

src/ofdm.c
```diff
@@ -749,6 +749,48 @@ static float est_freq_offset(struct OFDM *ofdm, complex float *rx, int timing_es
...
+static float est_freq_offset_pilot_corr(struct OFDM *ofdm, complex float *rx, int timing_est) {
+    complex float corr_st[ofdm_m + ofdm_ncp], corr_en[ofdm_m + ofdm_ncp];
+    int SFrame = ofdm_samplesperframe;
+
+    // "mix" down (correlate) the pilot sequences from frame with 0 Hz offset pilot samples
+    for (int i = 0; i < (ofdm_m + ofdm_ncp); i++) {
+        complex float csam = conjf(ofdm->pilot_samples[i]);
+        corr_st[i] = rx[timing_est + i         ] * csam;
+        corr_en[i] = rx[timing_est + i + SFrame] * csam;
+    }
```
drowe67: @srsampson we could also combine this step with the correlation below, removing the need for storage of corr_st[] and corr_en[]

srsampson: Yes, you're going to get two multiplies inside the loop either way. I've made the change, so see if it works.

drowe67: Nice work Steve, ctests pass so looks good.

I've implemented a high/low bandwidth phase estimation scheme, an extension on your high_doppler support.  Lets us acquire when the freq offset est is a bit off.  We "switch gears" using the state machine to retain the low SNR performance of the low bandwidth phase est.

Only problem is I broke your manual phase est bandwidth setting (formally --high-doppler option in ofdm_demod).  It would be nice to set the bandwidth high/low/auto from the command line.  Could you work on that pls?  I've highlighted the code that needs some work.

drowe67:
@donmr @db4ple Don and Danilo - I've been messing with the ofdm modem freq offset estimator.  Could you pls run your stm32 tests on this PR/branch and make sure I haven't broken anything?  ```bear``` is up if that helps.

donmr:
Will try that tommorow my time.

(yes I am still here)

Don

src/ofdm.c
srsampson: > It would be nice to set the bandwidth high/low/auto from the command line. Could you work on that pls? I've highlighted the code that needs some work.

I'll take a look and play with it.

octave/ofdm_lib.m
```diff
@@ -388,11 +433,34 @@ function print_config(states)
...
+    %printf("afcoarse: %f atiming_mx: %f\n", afcoarse, atiming_mx);
+    
+    if atiming_mx > timing_mx
+      ct_est = act_est;
+      timing_valid = atiming_valid;
+      timing_mx = atiming_mx;
+      fcoarse = afcoarse;
+    end
+  end
+
```
drowe67: @srsampson @donmr I hope this runs OK in real time on the stm32.  Many opportunities for optimisation, e.g. recursions or precomputation at compile time of ```wvec``` above.

donmr:
I just ran them on master and found 2 issues.  One was a minor PATH
issue that may not affect others.

The second is more serious, a file was corrupted.  I don't know how to 
quickly track down where this happened, can one of you look.

    codec2/stm32/unittest/scripts/tst_ofdm_demod_check has been corrupted
    in master.  Starting at line 475 there is junk added. This junk looks
    like parts of the file from previous lines, some repeated several
    times.

This bad file is also in the dr-ofdm-sync branch.

Don

drowe67:
@donmr @db4ple I've been testing the current sm1000_main.c and have some stability issues when I try to decode a real time 1600 signal played into the sm1000 from my laptop:
- poor quality decoded audio
- error/clip led blinking continuously
- after a few 10's of seconds, colourful ring of death

@db4ple is the current FreeDV 1600 code working OK on your target platforms?  Think I'll try to get a test running for FreeDV 1600 using @donmr's ```tst_api_demod.c```

I've written some code to dump the state of memory and can see the stack isn't meeting the heap:
```
1600 Rx                                                                                                                   
chunks of RAM segment > 256 bytes containing start up pattern:                                                            
  start: 0x20000c20  end: 0x20001624  bytes: 2564                                                                         
  start: 0x20003538  end: 0x20004170  bytes: 3128                                                                         
  start: 0x20004184  end: 0x200065a8  bytes: 9252                                                                         
  start: 0x2000fe80  end: 0x20010100  bytes: 640                                                                          
  start: 0x200105d0  end: 0x200106d0  bytes: 256                                                                          
  start: 0x20011e7c  end: 0x200120a8  bytes: 556                                                                          
  start: 0x20013ec0  end: 0x20019040  bytes: 20864
```
I fill the memory with a known pattern at start up, then look for where the pattern has been over written.

drowe67:
> I just ran them on master and found 2 issues. One was a minor PATH issue that may not affect others. The second is more serious, a file was corrupted. 

Hmm, not sure how to quickly look at a file, git checkout will let you inspect each one, git log tell you who and when it was checked it. Could it be an accidental copy/paste issue?

drowe67:
@donmr I merged https://github.com/drowe67/codec2/pull/43 and ran the stm32 tests:
```
88% tests passed, 2 tests failed out of 16

Total Test time (real) = 597.50 sec

The following tests FAILED:
	  8 - tst_ofdm_demod_AWGN (Failed)
	  9 - tst_ofdm_demod_fade (Failed)
```

The "fails" were AWGN:
```
Reference
BER......: 0.0800 Tbits: 14518 Terrs:  1162
BER2.....: 0.0755 Tbits:  9758 Terrs:   737
Target
BER......: 0.0783 Tbits: 15232 Terrs:  1193
BER2.....: 0.0745 Tbits: 10472 Terrs:   780
FAIL: max BER
FAIL: BER compare
```

and fade:
```
Reference
BER......: 0.0139 Tbits: 97104 Terrs:  1351
BER2.....: 0.0122 Tbits: 92344 Terrs:  1122
Target
BER......: 0.0124 Tbits: 97818 Terrs:  1215
BER2.....: 0.0122 Tbits: 93058 Terrs:  1131
FAIL: max BER
FAIL: BER compare
```

I'm happy with fails like that :-)

srsampson:
Setting HIGH_PHASE_EST passes, LOW_PHASE_EST fails (the old normal).

srsampson:
tofdm.c needs to have high and low runtime options added.

src/codec2_ofdm.h
```diff
@@ -49,7 +49,11 @@ extern "C" {
 #define UN_SYNC      0  /* Used with the ofdm_set_sync() */
 #define AUTO_SYNC    1
 #define MANUAL_SYNC  2
-    
+
+#define AUTO_PHASE_EST 0
+#define LOW_PHASE_EST  1
+#define HIGH_PHASE_EST 2
+
```
drowe67: So why do we need #defines and the enum to label the same variable?

src/ofdm.c
```diff
@@ -187,6 +188,7 @@ struct OFDM *ofdm_create(const struct OFDM_CONFIG *config) {
         ofdm_ntxtbits = 4;
         ofdm_ftwindowwidth = 11;
         ofdm_timing_mx_thresh = 0.30f;
+        ofdm_phase_est_bandwidth = LOW_PHASE_EST;
```
drowe67: Should be high bandwidth by default, as this is the initial state for the state machine.  The state machine then switches to low bandwidth once we get sync.

unittest/tofdm.c
```diff
@@ -208,6 +208,10 @@ int main(int argc, char *argv[])
     ofdm = ofdm_create(ofdm_config_default);
     assert(ofdm != NULL);
     ofdm_config = ofdm_get_config_param();
+
+    /////////////// needs option added above
+    ofdm_set_phase_est_bandwidth_mode(ofdm, HIGH_PHASE_EST);
+    ///////////////
```
drowe67: This shouldn't be necessary - ofdm_create() should start in HIGH_PHASE_EST, as per the Octave code.  The tofdm system exercises both high and low modes as the state machine "switches gears" as the modem runs and achieves sync.

drowe67:
> tofdm.c needs to have high and low runtime options added.

It exercises both modes as part of normal operation of the state machine .... pls see other comments.

src/codec2_ofdm.h
srsampson: > Do we need #defines and the enum to label the same variable? We could perhaps expose the enum or uses the #defines internally.

Currently the enum is in the ofdm_internal and not exposed public. The system calls also expect an integer. Thus, the simplest form is to use a define for public call, and translate to an enum internally.

I'm not hard over on that requirement.

src/ofdm.c
srsampson: My bad.

donmr:
Re: PR #43 and tst_ofdm_demod test fails.

My appologies, I thought they were all passing.  Obviously not.
I have updated the check script to be more tolerant of the exact error counts for these tests.
See PR #44 (same branch)

unittest/tofdm.c
drowe67: Thanks Steve - looks good.

I wonder of it's possible to select high/low/auto phase est bandwidth selection through ofdm API calls and via command line options in ofdm_demod?  If the option is not auto, we would need to disable the state machine switching logic.  However just a thought - not something we really need right now.

The parallel/wide range freq offset estimation code at line 436 of ofdm_lib.m needs porting to C if you're interested, otherwise I'll get around to it this weekend.  I'm pretty happy with this - finally a wider range for freq offset acquisition on this modem :-)

src/ofdm.c
```diff
@@ -973,25 +972,73 @@ int ofdm_sync_search_shorts(struct OFDM *ofdm, short *rxbuf_in, float gain) {
...
+    complex float *rx = (complex float *) &ofdm->rxbuf[0]; // complex has same memory layout
+    complex float wvec[2][2 * ofdm_samplesperframe];
+    int i, ref, act_est, afcoarse;
+
+    /* insert latest input samples into rxbuf so it is primed for when we have to call ofdm_demod() */
+    /* note can't use memcpy when src and dest overlap */
+
+    memmove(rx, &rx[ofdm->nin], (ofdm_rxbuf - ofdm->nin) * sizeof (complex float));
+    memcpy(&rx[(ofdm_rxbuf - ofdm->nin)], rx, ofdm->nin * sizeof (complex float));
+
```
drowe67: @srsampson memcpy with overlapping buffers is problematic....

srsampson:
The main problem is it doesn't seem to SYNC in High mode.

src/ofdm.c
```diff
@@ -983,7 +983,7 @@ static int ofdm_sync_search_core(struct OFDM *ofdm) {
 
     int fcoarse = 0;
     int n = 0;
-    int timing_mx = 0;
+    float timing_mx = 0;
     int ct_est = 0;       /* this could be bad default [srsampson] */
```
drowe67: Lol :+1:

drowe67:
@srsampson I'm thinking the work you just did was made harder than it needed to be as the ```est_timing()``` function doesn't explicitly return ```timing_mx``` and ```timing_valid```, but hides them away in the ofdm structure.  We could modify ```est_timing``` to return those variables, then load the ofdm structure at the end of ``` ofdm_sync_search_core()```.

drowe67:
I guess another question is the impact of searching over a wider range on the stm32 execution time for 700D.

src/ofdm.c
srsampson: I would have never found that!

srsampson:
> I guess another question is the impact of searching over a wider range on the stm32 execution time for 700D.

The wvec vector uses quite a chunk of heap space. The (w * i) also repeats itself each 2pi so many of the multiplies times "i" become redundant. But you would still need the space for the signal.

drowe67:
> > I guess another question is the impact of searching over a wider range on the stm32 execution time for 700D.
> 
> The wvec vector uses quite a chunk of heap space. The (w * i) also repeats itself each 2pi so many of the multiplies times "i" become redundant. But you would still need the space for the signal.

Good point @srsampson .... That would be 1280*2 samples * 8 bytes/complex sample, or 20k - too much!  Couple of ideas:
1. The freq shift could be done in place on ```ofdm->rxbuf``` (then undone for the next candidate).
1. The freq shift could be rolled into the correlation in ```est_timing()```, with the freq shift calculated on the fly.
2. We could generate 3 versions of ```ofdm->pilot_samples``` with -40, 0, +40- freq shift at init-time, and supply them as an argument to ```est_timing()```.  So do the shift on that side of the correlation, rather than ```ofdm->rxbuf```.
3. We could generate the three versions of ```ofdm->pilot_samples``` at ***build_time*** (with a C utility - like we build vector quantiser codebooks) and put them in a look up table in Flash.

Would you like to try optimising this?  I'm happy to review and help out if you get stuck

srsampson:
Here's a version I did to reduce the (w*i) redundancy. Passes all tests.

Doesn't reduce space though, just reduces multiples.

srsampson:
This is an attempt at putting the (w * i) multiplication in Read-Only-Memory, thinking of the stm firmware.

drowe67:
> This is an attempt at putting the (w * i) multiplication in Read-Only-Memory, thinking of the stm firmware.

Nice one Steve :+1: Now to how to test it on the stm32.  I think we need a new PR, that merges this one and https://github.com/drowe67/codec2/pull/46, which I have just added a bunch of tests too.

srsampson:
The table of complex values was generated from the cmplxconj() which has already swapped the signs, so I am using "* I" in the table. It seemed to drink in the values without the "I" being present. But maybe I should swap the imaginary signs and use * -I to be less confusing for first time code readers, I don't think it's confusing enough though...

drowe67:
@srsampson - will the modulo here ```(ofdm_wval[i % ofdm_nval]``` mean an integer divide in the inner loop?  Might be worth looking at the output assembler (gcc -s).  We have plenty of flash so no big deal to store the entire table.

drowe67:
@srsampson see what you think of https://github.com/drowe67/codec2/pull/41/commits/ff79e43959da47b8196149418a064d14df7dc2e9

I'm a bit concerned about a case statement (and modulo) inside the inner loop, but I guess it only happens once every 80ms.  The stm32 tests will hopefully show if there is a CPU issue.

srsampson:
> @srsampson - will the modulo here `(ofdm_wval[i % ofdm_nval]` mean an integer divide in the inner loop? Might be worth looking at the output assembler (gcc -s). We have plenty of flash so no big deal to store the entire table. Or reset the table pointer every `ofdm_val` samples.

There is no free lunch, is there? Well my thoughts were that an integer divide was cheaper than a floating point. But maybe just putting the whole table in flash would be better if there's room.

donmr:
A second index that is reset is fast and cheap.

srsampson:
The remainder op inside est_freq_offset_pilot_corr() and est_timing() aren't really needed as I believe they only reach 160 (144 + 16) and ofdm_nval is 200 (with a +/- 40 window.

srsampson:
Argh... Didn't notice interldpc.c used ofdm mod/demod functions.

src/ofdm.c
```diff
@@ -661,10 +681,22 @@ static int est_timing(struct OFDM *ofdm, complex float *rx, int length) {
...
+                csam = conjf(ofdm_wval[j % ofdm_nval]*ofdm->pilot_samples[j]);
+                break;
+            case 0:
+                csam = conjf(ofdm->pilot_samples[j]);
+                break;
+            case 40:
+                csam = ofdm_wval[j % ofdm_nval]*conjf(ofdm->pilot_samples[j]);
+                break;
+            default:
+                assert(0);
```
drowe67: @srsampson Hi - did you notice this code, that doesn't need wvec any more?  Would be interested in yr thoughts....

srsampson: I like how it goes frame by frame instead of across the whole RX buffer. This also doesn't need the ofdm_nval contraption, as the 160 samples is less than the 200 samples of wvec[].

srsampson: I was going to flip the wvec[] const array file to be non-conjugate and adjust the functions, but my math kind of blew up, so I left it alone. I was trying to reduce the number of conjf() calls, but I don't think there would be much change.

drowe67: > I like how it goes frame by frame instead of across the whole RX buffer. This also doesn't need the ofdm_nval contraption, as the 160 samples is less than the 200 samples of wvec[].

Oh cool I hadn't noticed that - was more interested in saving stack usage.  However I guess this is more inner-llop operations overall, we are down shifting the same samples again and again.  It's also a slightly different algorithm and the phase of the down shift changes for each ```corr[i]``` samples.  However as we take the abs() value it's probably OK.  And I guess tofdm passes.

I have tested the Octave code quite carefully with several simulations.  I might write some more challenging tests for the C version as well, to make sure it's getting good acquisition performance.

drowe67:
@srsampson I just merged master into this branch which causes all sorts of headaches, I had to manually apply many changes again.  Probably due to my Git skills.  Anyhoo, the tests are passing again. Pls take a look sometime  and see if I have messed anything up that you have added ....

drowe67:
Oops it's a bit slow on the stm32, when we go from search to synced:

```
~/codec2/stm32/build_stm32$ ctest -V -R tst_api_demod_700D_AWGN_codec
<snip>
22:   0 st: search euw:  0 0 f: -12.0 ist: search  0 eraw:   0 ecdd:   0 iter:   0 pcc:   0 vld: -1, nout:    0
22:   freedv_rx 415.17 msecs
22:   1 st: trial  euw:  5 1 f: -11.9 ist: synced  1 eraw:   0 ecdd:  62 iter:  10 pcc:  50 vld: 1, nout: 1280
22:   freedv_rx 52.44 msecs
```

I suspect that 415ms is the wide range timing and freq est.  Our budget is < 80ms. We'll have to figure out how to optimise it further.  Maybe put three versions of the pilot*w arrays in flash.  Might be worth another test to characterise the CPU load of those functions too.

However good news is all other stm32 tests pass.

srsampson:
Could be the GCC sin/cos are too slow for firmware.

srsampson:
> Pls take a look sometime and see if I have messed anything up that you have added ....

Looks good, my changes are still there.

drowe67:
> Could be the GCC sin/cos are too slow for firmware.

Hmm, well we could certainly remove the trig functions from the inner loop.  I'll do some profiling.

drowe67:
OK managed to reduce the freq est to trivial levels, but running out of tricks for timing, the CPU load is dominated by all those correlations:
```
frame: 3, 2560 bytes read
   ct_est:    0 foff_est: -4.0 timing_valid: 1 timing_mx: 0.4957
  3 st: search euw:  5 0 f:  -4.0 ist: search  1 eraw:   0 ecdd:   0 iter:   0 pcc:   0 vld: -1, nout:    0
  wvecpilot  0.04 msecs
  corr 37.46 msecs
  wvecpilot  0.01 msecs
  corr 37.46 msecs
  wvecpilot  0.04 msecs
  corr 37.46 msecs
  timing 113.72 msecs
  freq  6.21 msecs
  freedv_rx 121.46 msecs
```
Will have a think.

drowe67:
@db4ple  we're working on optimising ```est_timing()``` in https://github.com/drowe67/codec2/blob/dr-ofdm-sync/src/ofdm.c, it has a couple of big complex correlations.  Any thoughts on the most efficient way to do this on a stm32f4?  I've tried the f32 and q15 CMSIS functions.  q15 seems to be about the same speed as q32.  For the SM1000 we might be able to treat the input vector as real only.

db4ple:
Well, I looked at it for a couple of minutes but nothing "inspiring" came to my mind. 
Do you exactly know which line burns the majority of the cycles? I would assume arm_cmplx_dot_prod_f32 as it involve surely a number of floating point multiply ops. I would assume, that unless we can take advantage of some special conditions (like the imaginary part always 0 in one vector) there is not much reduction possible.

drowe67:
Thanks @db4ple.  Majority of cycles in the dot product, it's repeated at 1280 different timing offsets, which is then repeated 3 times in an outer loop to test different freq offsets.

Using real only I'm getting close, target is about 70ms:
``` 
wvecpilot  0.14 msecs
  corr 26.33 msecs
  wvecpilot  0.11 msecs
  corr 26.33 msecs
  wvecpilot  0.14 msecs
  corr 26.33 msecs
  timing 80.57 msecs
  freq  4.09 msecs
  freedv_rx 86.20 msecs
```

So I'm guessing integer maths is no faster than float on this chip?

drowe67:
Undersampling coarse timing did the trick.  The idea is that the peak in the coarse timing location is a slowly varying function, so I'm stepping two samples at a time.  We then do a fine est at 1 sample resolution after we get sync.  OFDM is robust to small timing offsets due to the cyclic prefix. Nice to know the modem algorithms when you do the optimisation :+1: 

```
  wvecpilot  0.14 msecs
  corr 13.18 msecs
  wvecpilot  0.11 msecs
  corr 13.18 msecs
  wvecpilot  0.14 msecs
  corr 13.18 msecs
  timing 41.11 msecs
  freq  3.97 msecs
  freedv_rx 46.62 msecs
```
I will need to test this further, e.g. on fading channels.

db4ple:
32bit mult should be a single cycle operation but only in best case (all operands available in registers).
Otherwise the pipeline gets stalled until both operands are loaded.

To get an idea, may use this as a info source: http://www.micromouseonline.com/2011/10/26/stm32f4-the-first-taste-of-speed/

drowe67:
Yes when I disassembled the library functions I saw a bunch of register/memory shuffling and just the occasional maths instruction.

drowe67:
@srsampson thanks for your input.  Do you have a stm32 Discovery card and/or a SM1000?

srsampson:
> @srsampson thanks for your input. Do you have a stm32 Discovery card and/or a SM1000?

I tried setting up the Discovery card, but I ran into problems getting the Make to build the sm1000.elf file. Right now I can't remember what it was.  I think it involved the sm2000 stuff.

srsampson:
I have rebuilt my computer after I did some terrible things to it. Now the stm32 compiles to completion, although has many warnings. Now I need to find my Discovery board :-)

drowe67:
Yes a lot of warnings out of cohpsk.c, @db4ple did some nice work to remove them, but when I later tested the cohpsk C/Octave port something subtle was broken.  It's probably just a small fix once it's tracked down.  There are also cases where you fix something for stm32 and x86 issues a warning.  So it's a balancing act.

drowe67:
@srsampson @donmr I think this is ready for merge.  Passes all tests (including many new ones), and running nicely on the stm32.

drowe67:
OK I merged master into this branch as a pre-merge sanity check.  18/18 x86 and 23/23 stm32 tests passing.  Yayyy for tests :+1:

drowe67:
Ach so now it passes with no changes.  I'll leave the verbose running of the script in case this pops up again.

PR #42, opened 2019-06-16
SM1000 Maintenance and 700D Integration

drowe67:
1. My sm1000.bin was crashing after a few seconds of decoding FreeDV 1600.  I'm adding USART support as a light weight way to find out why, e.g. dump stack/heap usage.  This will also set us up for integrating @donmr's 700D work.

1. Fixed the above using Don's stm32 test system to track down the problem https://github.com/drowe67/codec2/pull/46#issuecomment-504707023.  TTY support working well.

1. Modifying sm1000_main to integrate Don's fine 700D work.  Some challenges with RAM required for 700D ADC/DAC FIFOs, and support of both 1600 and 700D with the same instance of ADC/DAC drivers.

1. Modified the OFDM modem to use smaller maximum number of input samples to reduce buffer size requirements in https://github.com/drowe67/codec2/pull/50

1. Found and fixed memory leaks in 1600 and 700D in https://github.com/drowe67/codec2/pull/50

1. Moved the FIFO bulk storage and some buffers to CCM.

1. 700D decode working :1st_place_medal: Well don Don, Steve, Danilo, Richard.

1. Tx working - amazing 35dB SNR and tight scatter when connected to a laptop over the bench.  Decodes recordings of low SNR/fading channel off air samples just fine, showing FEC/modem is working well.  Small speaker quality not great, oh well.

1. Mutes output speech when FEC doesn't decode, so we don't hear R2D2 at the end of an over.  Seems to resync OK after PTT pressed (ie transition form tx to rx).

1. Measured CPU load around rx code by probing PE3 with an oscilloscope: 83ms in acquisition, 56ms peak when decoding vk2tpm_004.wav (off air low SNR fading).  700D frame period is 160ms ... actually I think I made a mistake previously and was targeting 80ms as real time budget!  Oops, it's too fast!

1. Added V2 version number to binary image and power on morse prompt.

1. Draft SM1000 Manual.


drowe67:
@donmr @db4ple I've been testing the current sm1000_main.c and have some stability issues when I try to decode a real time 1600 signal played into the sm1000 from my laptop:
- poor quality decoded audio
- error/clip led blinking continuously
- after a few 10's of seconds, colourful ring of death

@db4ple is the current FreeDV 1600 code working OK on your target platforms?  Think I'll try to get a test running for FreeDV 1600 using @donmr's ```tst_api_demod.c```

I've written some code to dump the state of memory and can see the stack isn't meeting the heap:
```
1600 Rx                                                                                                                   
chunks of RAM segment > 256 bytes containing start up pattern:                                                            
  start: 0x20000c20  end: 0x20001624  bytes: 2564                                                                         
  start: 0x20003538  end: 0x20004170  bytes: 3128                                                                         
  start: 0x20004184  end: 0x200065a8  bytes: 9252                                                                         
  start: 0x2000fe80  end: 0x20010100  bytes: 640                                                                          
  start: 0x200105d0  end: 0x200106d0  bytes: 256                                                                          
  start: 0x20011e7c  end: 0x200120a8  bytes: 556                                                                          
  start: 0x20013ec0  end: 0x20019040  bytes: 20864
```
I fill the memory with a known pattern at start up, then look for where the pattern has been over written.

donmr:
(I tried this reply by email a couple of days ago but it wouldn't go through...)

On Sun, Jun 16, 2019 at 01:11:27AM -0700, drowe67 wrote:
> @drowe67 pushed 1 commit.
>
> a13c062843f7ae853b8aae21bb47a131eb21b43f  first pass at a way to inspect used memory.
> Startup assembler fills memory with a known value, function then prints range of values
> filled with that value to serial port.  Some of these ranges will be heap that hasn't been
> filled yet

I tried this a few months ago.  I found a lot of small areas that were
not modified when the code ran which made it difficult to tell what the
actual usage was.

Instead I made a custom version of sbrk.c which prints the new heap
limit when it is expanded.

Don

donmr:
(I tried this reply by email a couple of days ago but it wouldn't go through...)

On Sun, Jun 16, 2019 at 01:11:27AM -0700, drowe67 wrote:
> @drowe67 pushed 1 commit.
>
> a13c062843f7ae853b8aae21bb47a131eb21b43f  first pass at a way to inspect used memory.
> Startup assembler fills memory with a known value, function then prints range of values
> filled with that value to serial port.  Some of these ranges will be heap that hasn't been
> filled yet

I tried this a few months ago.  I found a lot of small areas that were
not modified when the code ran which made it difficult to tell what the
actual usage was.

Instead I made a custom version of sbrk.c which prints the new heap
limit when it is expanded.

Don

drowe67:
> I tried this a few months ago. I found a lot of small areas that were
> not modified when the code ran which made it difficult to tell what the
> actual usage was.

Yep, as per the table above.  Although I think the large chunk at the end is close to what we want.  It decreases when freedv_api starts running.
 
> Instead I made a custom version of sbrk.c which prints the new heap
> limit when it is expanded.

Sounds useful.  Is it committed anywhere?  Might be useful to have the tests automatically fail if they detect running out of memory.

drowe67:
Ok FreeDV 1600 is working OK on the SM1000 now after the fdmdv modem fixes from https://github.com/drowe67/codec2/pull/46

TTY support is nice too.

drowe67:
@donmr @srsampson @db4ple I have 700D decoding on a SM1000.  Well done guys :+1: 

Tx does something when I press PTT, but I haven't tried to decode it.

Still some fine tuning and one of the x86 tests is failing above.  However it's quite stable, and you can cycle between analog-1600-700D with the select button and Stuart's menu system.  We'll need to run it through a test plan too (if only manually) like some on air tests, and simulated channel impairments.

Recommend hooking up a RS232 TTY if you want to try it, very useful.  See SM1000 schematic for TTY jumper pin out.

drowe67:
The fading channel test is failing.  Seems to be giving different results on each run. I wonder if it's using a different seed every time generate the simulated fading.  Problem with fading simulations is you need some big multiple of the fade period seconds to get the statistical average of the channel BER.  I worked it out once at a few hours, based on a ITUT standard for HF fading simulations.

srsampson:
Excelent work. Amazing that two vocoders and two modems fit in memory! Now that you have 700C vocoder, can you squeeze in 4FSK voice :-)

srsampson:
I wonder if smaller Morse Code strings might be better (send faster than numbers), say two/three letter codes?

"ANA", "FDM", "OFDM"

drowe67:
> I wonder if smaller Morse Code strings might be better (send faster than numbers), say two/three letter codes?
> 
> "ANA", "FDM", "OFDM"

Be nice to have an option for compressed voice prompts.  1 second of speech is just 700 bits (< 100 bytes).  Just have to choose which accent :)

drowe67:
@hobbes1069 is there a way we can add a git hash to sm1000_main.c such that:
1. The tty prints the hash (first 6 characters)
1. The hash gets played in morse (again just first 6 characters)
when the sm1000 boots?

drowe67:
@donmr fair bit of work in the PR so I will merge shortly, and kick off another branch iwhen we need some more changes. All stm32 tests passing for me.

donmr:
Great work.  Yes it seems like the best method for Git is to make a new branch for each "feature".

drowe67:
> Great work. Yes it seems like the best method for Git is to make a new branch for each "feature".

Yes I think the trick is to not make branches too bit before merging.  A skill I'm still working on :-)

PR #43, opened 2019-06-16
Fixes to STM32 tests

donmr:
Two things, first a simple change to put the local binaries and libraries at the front of their respective paths instead of at the end, needed because the library version changed.

Second, the file  codec2/stm32/unittest/scripts/tst_ofdm_demod_check has been corrupted in master.
Starting at line 475 there is junk added.  This junk looks like parts of the file from previous lines, some repeated several times.  I just fixed it, I didn't track down the source.


drowe67:
@donmr so is this ready for merge (i.e. all stm32 tests passing)?

drowe67:
So @donmr are ```tst_api_*``` run anywhere?  I'd like to add a FreeDV 1600 test.

donmr:
Yes, I wouldn't create a pull request if it wasn't ready.

Don

donmr:
On Mon, Jun 17, 2019 at 05:18:25AM -0700, drowe67 wrote:
> So @donmr are ```tst_api_*``` run anywhere?  I'd like to add a FreeDV 1600 test.

No.  As I recall they were not working because the outputs would not
match.

I'll take another look when we have the other tests all passing again.

Don

drowe67:
> On Mon, Jun 17, 2019 at 05:18:25AM -0700, drowe67 wrote: So @donmr are ```tst_api_*``` run anywhere? I'd like to add a FreeDV 1600 test.
> No. As I recall they were not working because the outputs would not match. I'll take another look when we have the other tests all passing again. Don

I assume you mean after this PR is merged, as you indicated all tests were passing above.  I'll merge now.

PR #50, opened 2019-07-06
Reduced ofdm_get_max_samples_per_frame()

drowe67:
While working on https://github.com/drowe67/codec2/pull/42 I re-discovered ofdm_get_max_samples_per_frame() is required to be two frames long.  This is because during acquisition, est_timing() will output coarse timing between 1 and 2 frames in the future, resulting on ```nin``` values between 1 and two frames in the future.  

For the stm32 his means we need buffers and DAC/ADC FIFOs at least 2 700D frames long, a lot of memory at Fs=16 kHz.

This PR is an attempt to reduce the memory required by reducing the coarse timing and ```nin``` range to 0 to 1 frame.

### DIFF
--- a/src/ofdm.c
+++ b/src/ofdm.c
@@ -687,8 +687,8 @@ static int est_timing(struct OFDM *ofdm, complex float *rx, int length,
         }
     }
 
-    *timing_valid = (*timing_mx > ofdm_timing_mx_thresh); /* bool but used as external int */
-
+    // only declare timing valid if there are enough samples in rxbuf to demodulate a frame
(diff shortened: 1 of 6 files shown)


src/ofdm.c
```diff
@@ -1038,7 +1038,7 @@ static int ofdm_sync_search_core(struct OFDM *ofdm) {
 
         /* calculate number of samples we need on next buffer to get into sync */
 
-        ofdm->nin = ofdm_samplesperframe + ct_est;
+        ofdm->nin = ct_est;
```
drowe67: @srsampson be interested in your thoughts on this one!

PR #52, opened 2019-07-14
Create a modem version with one carrier

srsampson:
The modem will be called with nc=1 thus ofdm_ncp will equal 0 so we need to correct for that where necessary.

### DIFF
--- a/src/ofdm.c
+++ b/src/ofdm.c
@@ -425,21 +425,28 @@ struct OFDM *ofdm_create(const struct OFDM_CONFIG *config) {
     idft(ofdm, temp, ofdm->pilots);
 
     /*
-     * pilot_samples is 160 samples, but timing and freq offset est
-     * were found by experiment to work better without a cyclic
-     * prefix, so we uses zeroes instead.
+     * Timing and freq offset est were found by experiment to
+     * work better with cyclic prefix using zeroes instead.
      */
 
-    /* zero out Cyclic Prefix (CP) time-domain values */
+    if (ofdm_ncp == 0) {
+        /* With no cyclic prefix just fill the symbol part */
 
-    for (i = 0; i < ofdm_ncp; i++) {
-        ofdm->pilot_samples[i] = 0.0f;
-    }
+        for (i = 0; i < ofdm_m; i++) {
+            ofdm->pilot_samples[i] = temp[i];
+        }
+    } else {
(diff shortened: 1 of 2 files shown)


srsampson:
I'm playing with the tx side here. The rx side will have problems. The unique word algorithm probably needs to be replaced. I'm using ofdm_mod to interface and looking at spectrum with audacity. So far it's pretty ugly.

srsampson:
> The modem will be called with nc=1 thus ofdm_ncp will equal 0 so we need to correct for that where necessary.

Basically it is just a QPSK signal with BPSK pilots.

srsampson:
```
+------------+------------ ... --------------+------------+--
| BPSK Pilot | QPSK Symbol     QPSK Symbol N | BPSK Pilot |
+------------+------------ ... --------------+------------+--
```

drowe67:
Nice work Steve, maybe start with a low symbol rate like 100 symb/s.  Average spectrum should look like a sync function.

I guess on Rx for phase estimation we will average the single pilot of the usual 4 positions in past and future.

We will need to put some thought into UW, and also initial timing estimation - not much info to correlate on in one carrier. Might need to extend to multiple "frames".

srsampson:
I guess in the end, we would need a symbol rate that would fit in a 300 - 3500 kHz or so bandwidth.

srsampson:
I was using:

```
./ofdm_gen_test_bits --out test.dat --frames 100
```
and
```
./ofdm_mod --in test.dat --nc 1 --tcp 0 --ts 0.0005 --out test.raw --verbose 2 --ns 113 --tx_freq 1600 --rx_freq 1600
```

Where ts = 1/2000

Results in:
```
ofdm_bitsperframe = 224
ofdm_rowsperframe = 112
ofdm_samplesperframe = 452
```

Time-domain Spectrum seems clipped.

drowe67:
Good point @srsampson. Unfiltered PSK at 2000 symbols/s has a main lobe 2000 Hz wide, so that should be OK.  But while developing might be nice to allow room for a few more lobes, so it's a nice scatter diagram on rx

srsampson:
I keep rediscovering that audacity has its limitations... 

2khz-1500-center

(2 kHz 1.5 kHz center)

Looks like it's shifted left 100 Hz or so. Guess I'll check the omega function...

srsampson:
I deleted my previous comment, as I was using the wrong signal. argh...

PR #53, opened 2019-07-16
Windows DFU cleanup

drowe67:
1. Correct documentation
1. Build system support for DFU file creation

Thanks @srsampson for helping out with this.


drowe67:
@srsampson I use a utility called ```grip``` to render .md files without having to push them to GitHub.

srsampson:
There is a TODO in the manual because I haven't run the software, but it looks like the DfuSeDemo software may also be able to create a DFU file??

drowe67:
> There is a TODO in the manual because I haven't run the software, but it looks like the DfuSeDemo software may also be able to create a DFU file??

@db4ple reported using DfuSeDemo was a complex procedure - so best to generate the DFU file as part of the build process, and make it the simpler process that Walter/Geroge have suggested.  I'm happy to help with cmake if you get stuck. I think there are targets for .elf -> .bin already, so it' similar syntax I imagine.

srsampson:
> > There is a TODO in the manual because I haven't run the software, but it looks like the DfuSeDemo software may also be able to create a DFU file??
> 
> @db4ple reported using DfuSeDemo was a complex procedure - so best to generate the DFU file as part of the build process, and make it the simpler process that Walter/Geroge have suggested. I'm happy to help with cmake if you get stuck. I think there are targets for .elf -> .bin already, so it' similar syntax I imagine.

I will defer to your expertise, as I can only spell 'cmake'. I uploaded the python the the stm32/support directory in case it will be helpful. We can delete it later, as the cmake will probably be a better path.

db4ple:
Now the default cmake build will also create sm1000v2.dfu 
It requires a working python2 install to succeed. Most linux machines have that out of the box.
@hobbes1069: I have no idea how to check that requirement in cmake easily

srsampson:
When I ran the hex2dfu.py manually on the published V2 .bin file I got a DFU file size of 300325, but when I do a build (using .hex) the size is 293645. This is not to say either is wrong, but maybe the .hex does a better job. I haven't uploaded the image yet. Will borrow a friends device again tomorrow, and install the DFU tool others are using to my laptop. This is just a comment, not a bug.

srsampson:
This build addition will simplify the Manual!

srsampson:
I installed the software and changed the manual. I haven't verified the install yet. The manual provides a link to David's system (file not there yet) so I expect you will want to provide that to the users. The one I created from a current build is in my Dropbox:

https://www.dropbox.com/sh/ga8pg0xn5e3jcip/AAB1COq7nPHQlsu13iyZc7KKa?dl=0

...or build yourself.

srsampson:
I'm not sure why I got that merge conflict, and I hope I haven't messed up.

drowe67:
Thanks @srsampson and @db4ple.  The .dfu file has been generated and I have placed it on my web site.  I haven't tried to flash using Windows yet. I'll merge shortly.

PR #57, opened 2019-07-25
Minor changes to remove double promotion warnings

srsampson:
These warning show up in stm32 build. tofdm.c has a slightly modified version of the algorithm in cohpsk.c.

These use double as the numbers are so small.


drowe67:
Looks good @srsampson - took me a moment to realise this is code to simulate the sample clock offset, and not part of the actual modem.

srsampson:
Yes, looks to be unittest only.

PR #58, opened 2019-07-26
SM1000 firmware maintenance

drowe67:
## Notes

1. Mal [reported](http://www.rowetel.com/?p=6835#comment-81036) reported the UI menu wasn't coming up.  This is fixed but now it's not saving the preferred mode to flash when it's 700D.

1. I'd also like to add a menu option for the Codec 2 700C Equaliser in this PR, or perhaps just hard wire it on, as it's working OK with the x86 version.

1.  There's been a bug fix https://github.com/drowe67/codec2/pull/81 to the timing acquisition on the OFDM modem for 2020, which also affects 700D and hence the SM1000.  This can be obtained just by merging with master & re-building, but is also a good reason for a new firmware release.

1. @srsampson has done some fine work in https://github.com/drowe67/codec2/pull/83 reducing stm32 RAM by a few more kbytes, which we can obtain by merging with master.

1. Monitor serial port with:
   ```
   tio -m INLCRNL /dev/ttyUSB0
   ```
## ToDo
1. ~~Fix saving operating mode bug~~
1. ~~Enable equaliser for 700D~~
1. ~~Set operating mode after saved in menu~~
1. ~~Announce operating mode on boot~~

## Test Plan

| Test | Description | Who |  Status |
| --- | --- | --- | --- |
| 1 | Saving op mode to flash | Mal  | **Pass** |
| 2 | stm32 unit tests |  David |  **Pass** |
| 3 | Tx/Rx FreeDV 1600/700D over the bench |  David | **Pass** |
| 4 | Tx/Rx FreeDV 1600/700D over the air |  Community | **Pass** |


drowe67:
@hobbes1069 is it possible please to add some automation around the versioning of the SM1000 firmware?  I'm currently manually changing the v2 to v3 string in several places (see diffs).

VK8MT:
On further investigation I can confirm that there were two bugs. The UI menu has been fixed. but the boot-up mode bug (which was also there in v.2) is yet to be resolved.

PR #59, opened 2019-07-28
Centralize LED and Switches to sm1000_leds_switches

srsampson:
Move functions from debugblinky to sm1000_leds_switches
Add functions led_debug0() through led_debug3()
rename "rt" to "sync"

flip negative PTT logic for the application programmer

Add schematic (rev F) to doc directory

Not fully tested


drowe67:
@srsampson - has yr SM1000 arrived Steve?

srsampson:
> @srsampson - has yr SM1000 arrived Steve?

Nope! Probably still in customs...

I'll fully test this when it comes in.

srsampson:
I should say I'm borrowing an sm1000 from a non-ham friend. He's got his kind of torn-up as he wants to put it in a bigger case for his workshop. He says he's going to take the test some day, ha.

So I haven't tested the audio or features. I only know the buttons all work, and there is no circle of death display.

srsampson:
I copied the files into the "inc" directory by mistake, so then moved them to "src" but I guess I didn't use a git move, so they remained. Deleted them now.

drowe67:
@srsampson - the reason for separate files for sm1000 leds/debug blinky is that some of the executables were not SM1000 based, but ran just on Discovery cards.  So in those cases we may not want to init the GPIOs used for the SM1000.

Also - many files changed here, so a lot of testing will be required to verify these changes.  Many of the programs changed in this PR are really just one off experiments and not in general use, so not sure if they warrant the effort of test and maintenance.

I'm OK with the change to sensible names for debug_blinky functions, but would prefer they remain in a separate debug_blinky module so they are orthogonal to the sm1000 specific GPIOs.  Would also suggest you only change those files in active use - such as those used in the SM1000.  Those changes will need to be carefully tested.

If you are interested in SM1000 work - some other tasks on going in #2

PR #61, opened 2019-08-07
Fix log likelyhood.

ED6E0F17:
Fixes: b8a8f669dc "modified ldpc_noise ... and refactored LLR mapping ..."
	which removed a normalisation step.

This is not apparent in your unit testing because that is using normalised samples, but it can have an adverse impact on anyone using your code without normalising the samples before applying ldpc error correction.

### DIFF
--- a/src/mpdecode_core.c
+++ b/src/mpdecode_core.c
@@ -591,7 +591,7 @@ void sd_to_llr(float llr[], double sd[], int n) {
 
     estEsN0 = 1.0/(2.0L * estvar + 1E-3);
     for(i=0; i<n; i++)
-        llr[i] = 4.0L * estEsN0 * sd[i];
+        llr[i] = 4.0L * estEsN0 * sd[i] / mean;
 }
 
 


drowe67:
Thanks, I'll need a few days to check this one out, as this code is important to  bunch of projects. Can you please suggest/add a ctest that demonstrates the problem and the fix?

ED6E0F17:
You will want to do a lot of testing. I confirmed it to my satisfaction by increasing the signal and noise by a factor of 10 in the existing unit testing, and that was enough to make it fail without the re-normalisation, but I have not done any testing on the optimal levels to minimise decoding passes and reach maximum BER correction.

drowe67:
@ED6E0F17 @darksidelemm I've taken a look at this one and agree it could be impacting the LDPC code performance for Horus applications.  The function in used in several places so will need to be tested properly:

1. Write automated unit tests for each place in Codec 2 it is used.
1. Specific code examples to demonstrate the problem and the solution.
1. Integrate into ctests.

I don't have the time of inclination to work on this right now, so will move it to an Issue and close this PR.  Happy to guide anyone from the Horus community who wants to progress this in future.

PR #62, opened 2019-08-07
Fix ASCII buffer size.

ED6E0F17:
Horus Binary output is ASCII hexidecimal encoded.

This may be related to the issue that you had with OSX builds of horus_demod.

### DIFF
--- a/src/horus_api.c
+++ b/src/horus_api.c
@@ -445,7 +445,8 @@ int horus_get_max_ascii_out_len(struct horus *hstates) {
         return hstates->max_packet_len/10;     /* 7 bit ASCII, plus 3 sync bits */
     }
     if (hstates->mode == HORUS_MODE_BINARY) {
-        return HORUS_BINARY_NUM_PAYLOAD_BYTES;
+        return (HORUS_BINARY_NUM_PAYLOAD_BYTES*2+1);     /* Hexadecimal encoded */
+
     }
     assert(0); /* should never get here */
     return 0;


drowe67:
@darksidelemm could you review this please?

darksidelemm:
I'm still having issues building the entirety of codec2-dev and freedv under OSX.

HOWEVER - I can build the horus_demod command-line utility with this change, and it works as expected - but it also worked fine without this change as well? 

I'll continue hacking around with the cmake files to get freedv built and see if that makes any difference.

darksidelemm:
Managed to build codec2-git by removing ofdm_stack from the unittest build CMakefiles (clang doesn't support -no-pie or the -Map options), but i'm still stuck building freedv due to all sorts of fun C++ library errors, such as:
`
In file included from /Users/darkside/Dev/codec2-git/src/codec2_ofdm.h:37:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/complex.h:23:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/ccomplex:21:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/complex:247:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/sstream:188:1: error: templates
      must have C++ linkage
template <class _CharT, class _Traits, class _Allocator>
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
`
Until I can figure out what the heck is going on here (I've solved this before, but of course I didn't take notes and can't remember what's going on), I won't be able to test how these changes affect the freedv GUI.

drowe67:
Thanks @darksidelemm.  

It's probably acceptable to verify any changes using Linux command line tests at the codec2 level (before freedv-gui integration). We have an automated test for horus binary that has passed ("All checks have passed" green tick above).  The command line for the test is [here](https://github.com/drowe67/codec2/blob/master/CMakeLists.txt#L353).  It exercises horus binary on a small test file of off-air samples.

You can run this test manually:
```
$ cd codec2/build_linux
$ ctest -V -R test_horus_binary
```

I think this PR fixes some sort of buffer overflow, the problem might be apparent from inspection of the C source.  if you get stuck let me know and I'll review (sometime).

Re building freedv-gui for windows we know have docker support to build from pretty much any Linux machine.  I can show you how if you like.

darksidelemm:
I think the concern here is more about interaction between the horus_api and freedv. The command line utilities have always worked. That said, horus_api+freedv has also worked fine for ages under Windows & Linux, so I'm not really sure how this PR changes things. Looking at the code, the horus_get_max_ascii_out_len function is only called from within the RTTY decoder function, which we don't really use. 

My build issues are around building freedv under OSX, not under windows (you clearly have that well in hand). Unfortunately building freedv under OSX is proving to be a huge challenge at the moment, and I think it's going to have to be done by someone with more experience in this area than me.

drowe67:
@ED6E0F17 can you pls respond to @darksidelemm comments above about why this PR is needed?

> I think the concern here is more about interaction between the horus_api and freedv. The command line utilities have always worked. That said, horus_api+freedv has also worked fine for ages under Windows & Linux, so I'm not really sure how this PR changes things. Looking at the code, the horus_get_max_ascii_out_len function is only called from within the RTTY decoder function, which we don't really use.

darksidelemm:
I've given up on trying to get freedv to compile under OSX for now. Have posted an email to freetel-codec2 asking for help. Suspect libc++ issues, but don't know how to resolve.

PR #66, opened 2019-08-11
Fix compile errors on OSX/LLVM due to stricter requirements for extern.

tmiw:
Because of where extern "C" is defined in various include files, FreeDV fails to build (complaining about "templates needing C++ linkage", among other issues). Moving extern "C" below #includes resolves those particular compiler errors on OSX when using LLVM.

In addition, ofdm_stack needs to be excluded from OSX builds on my 10.14 machine. I'm not sure if that's needed long-term or not.

### DIFF
--- a/unittest/CMakeLists.txt
+++ b/unittest/CMakeLists.txt
@@ -31,14 +31,13 @@ target_link_libraries(ofdm_mem m)
 
 add_library(function_trace STATIC ../unittest/function_trace.c)
 
-if (NOT APPLE) 
 add_executable(ofdm_stack ofdm_stack.c ../src/ofdm.c ../src/octave.c ../src/kiss_fft.c ../src/modem_probe.c ../src/mpdecode_core.c ../src/phi0.c ../src/filter.c)
 if (CMAKE_C_COMPILER MATCHES "gcc$")
+    target_link_libraries(ofdm_stack function_trace m -no-pie "-Wl,-Map=ofdm_stack.map")
     target_compile_options(ofdm_stack PUBLIC -fstack-usage -finstrument-functions -no-pie)
 else()
+    target_link_libraries(ofdm_stack function_trace m -no-pie)
     target_compile_options(ofdm_stack PUBLIC -finstrument-functions -no-pie)
-endif()
-target_link_libraries(ofdm_stack function_trace m -no-pie "-Wl,-Map=ofdm_stack.map")
 endif()
 add_definitions(-D__UNITTEST__)
 
--- a/unittest/CMakeLists.txt
+++ b/unittest/CMakeLists.txt
@@ -31,6 +31,7 @@ target_link_libraries(ofdm_mem m)
 
 add_library(function_trace STATIC ../unittest/function_trace.c)
 
+if (NOT APPLE) 
 add_executable(ofdm_stack ofdm_stack.c ../src/ofdm.c ../src/octave.c ../src/kiss_fft.c ../src/modem_probe.c ../src/mpdecode_core.c ../src/phi0.c ../src/filter.c)
 if (CMAKE_C_COMPILER MATCHES "gcc$")
     target_compile_options(ofdm_stack PUBLIC -fstack-usage -finstrument-functions -no-pie)
@@ -38,6 +39,7 @@ else()
     target_compile_options(ofdm_stack PUBLIC -finstrument-functions -no-pie)
 endif()
 target_link_libraries(ofdm_stack function_trace m -no-pie "-Wl,-Map=ofdm_stack.map")
+endif()
 add_definitions(-D__UNITTEST__)
 
 add_executable(tnewamp1 tnewamp1.c ../src/quantise.c ../src/newamp1.c ../src/mbest.c ../src/kiss_fft.c ../src/sine.c ../src/nlp.c ../src/dump.c ../src/octave.c ${CODEBOOKS})
--- a/src/codec2_ofdm.h
+++ b/src/codec2_ofdm.h
@@ -28,10 +28,6 @@
 #ifndef CODEC2_OFDM_H
 #define CODEC2_OFDM_H
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 /* Includes */
     
 #include <complex.h>
@@ -42,6 +38,10 @@ extern "C" {
 #include "modem_stats.h"
 
 /* Defines */
+
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 #define OFDM_AMP_SCALE (2E5*1.1491/1.06)   /* use to scale to 16 bit short */
 #define OFDM_CLIP (32767*0.35)             /* experimentally derived constant to reduce PAPR to about 8dB */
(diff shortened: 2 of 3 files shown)


drowe67:
thanks @tmiw - no ofdm_stack is just a helper program for the stm32 port. So just to confirm - this PR against the codec2 repo is all we need to allow freedv-gui to build OK for OSX?

tmiw:
There are additional changes for freedv-gui too. I'm not sure they're the correct ones, however, as my binary is segfaulting on startup. The current changes I've made there are in https://github.com/tmiw/freedv-gui/commit/b06cbc5129098b8c8576d1b388b7628ad11dd387.

unittest/CMakeLists.txt
```diff
@@ -31,13 +31,15 @@ target_link_libraries(ofdm_mem m)
 
 add_library(function_trace STATIC ../unittest/function_trace.c)
 
+if (NOT APPLE)
```
db4ple: Moving
``` 
target_link_libraries(ofdm_stack function_trace m -no-pie "-Wl,-Map=ofdm_stack.map")
```
into the "then" and
```
target_link_libraries(ofdm_stack function_trace m -no-pie ")
```
into "else" instead of wrapping it in the NOT APPLE if keeps it working for MacOS X.

tmiw: Updated. Thanks!

darksidelemm: This doesn't work for me. -no-pie doesn't exist for clang.

darksidelemm: I should note that my compiler is Apple Clang 7.3.0, not sure if the option is available in newer versions. Going back to the if (NOT APPLE) directive made this repo build again for me.

tmiw: On my 2017 MBP, FWIW:

MacBook-Pro:build_osx mooneer$ cc -v
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
MacBook-Pro:build_osx mooneer$ 

Also, my version prints a message indicating that -no-pie isn't used:

[ 69%] Linking C executable ofdm_stack
cd /Users/mooneer/devel/freedv-gui/codec2/build_osx/unittest && /opt/local/bin/cmake -E cmake_link_script CMakeFiles/ofdm_stack.dir/link.txt --verbose=1
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc  -Wall -Wno-strict-overflow -std=gnu11 -fPIC -g -O2 -DDUMP -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names  CMakeFiles/ofdm_stack.dir/ofdm_stack.c.o CMakeFiles/ofdm_stack.dir/__/src/ofdm.c.o CMakeFiles/ofdm_stack.dir/__/src/octave.c.o CMakeFiles/ofdm_stack.dir/__/src/kiss_fft.c.o CMakeFiles/ofdm_stack.dir/__/src/modem_probe.c.o CMakeFiles/ofdm_stack.dir/__/src/mpdecode_core.c.o CMakeFiles/ofdm_stack.dir/__/src/phi0.c.o CMakeFiles/ofdm_stack.dir/__/src/filter.c.o  -o ofdm_stack libfunction_trace.a -lm -no-pie 
clang: warning: argument unused during compilation: '-nopie' [-Wunused-command-line-argument]
[ 72%] Built target ofdm_stack

drowe67:
Looks like great progress guys. Can someone with an OSX machine pls tell me if this build of freedv-gui for OSX works?

A simple test plan would be to use the files in the [wav](https://github.com/drowe67/codec2/tree/master/wav) directory for some rx tests, in particular can you decode a:
1. 700D wave file
1. 2020 wave file

A better test would be loop back tx-rx, either via a loopback sound device or audio cables.

darksidelemm:
So a few notes:
- Horus Binary decoding works!  (tested with an actual payload)
- Decoding FreeDV 2020 via a loopback audio interface works.
- Decoding FreeDV 700D via a loopback audio interface works.
- I can transmit out of my IC-7610 via its USB audio interface, and control PTT (tested with both serial (RTS) PTT and Hamlib CAT PTT)
- The waterfall display is quite jerky. In fact, all the plots are quite jerky, on both receive and transmit. It doesn't seem to affect the decode or transmit performance however.
- I quite often get issues where the sample rate dropdown list in the audio config doesn't show any values. I've encountered this under Linux as well. Not sure if this is a portaudio issue.
- Something that is *definitely* a portaudio issue is the warnings I'm now noticing:
`2019-08-12 17:56:07.449 freedv[67808:6995508] 17:56:07.449 WARNING:  140: This application, or a library it uses, is using the deprecated Carbon Component Manager for hosting Audio Units. Support for this will be removed in a future release. Also, this makes the host incompatible with version 3 audio units. Please transition to the API's in AudioComponent.h.`
This is mentioned here: https://app.assembla.com/spaces/portaudio/tickets/218-pa-coreaudio-uses-some--quot-deprecated-quot--apis----this-is-by-design-but-need/details  Unfortunately it looks like nothing has happened on this issue for a very long time. I suspect it will only get fixed when these APIs disappear in newer versions of OSX.

I'd say we're onto a winner here!

unittest/CMakeLists.txt
darksidelemm: I guess they added support in for newer versions of clang, at least to note it's not used.

Still, maybe just disable it for APPLE entirely... unless there are actually developers who want to develop for the SM1000 under OSX? (and good luck to them...)

ra1nb0w:
The change fix the build of freedv-gui on macOS. Anyway, freedv-guid requires a cmake change because if you install `lpcnet` you will have `/opt/local/lib/cmake/lpcnet/lpcnet-config.cmake` and therefore you need to have `find_package(lpcnet REQUIRED)`. But this is another story also because freedv-gui `CMakeLists.txt` require some changes to avoid sneaky behaviour like compiler checks failure.


Now when `freedv-gui` starts arise the following problem

```
../src/common/sizer.cpp(2085): assert "!(flags & (wxALIGN_RIGHT | wxALIGN_CENTRE_HORIZONTAL))" failed in DoInsert(): Horizontal alignment flags are ignored with wxEXPAND
```


About `-no-pie`: if I remember well clang < 5.0 doesn't have `-no-pie` or `-nopie` so if it is really needed when need to check it with
```
CHECK_C_COMPILER_FLAG("-no-pie" NO_PIE_FLAG)
if(NO_PIE_FLAG)
  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -no-pie")
```

@drowe67 wav tests works correctly with latest commit and this patches on
macOS 10.14.6 18G87
Xcode 10.3 10G8



When everything works I will update codec2, lpcnetfreedv and freedv-gui ports on macports to easily use these great projects.

ra1nb0w:
Note: I pushed latest commit of lpcnetfreedv, codec2 and freedv-gui to macports. Seems to work fine. Just a note as already said by @darksidelemm: "The waterfall display is quite jerky"

PR #76, opened 2019-10-01
CMakeLists fixup for check_symbols_exist

hobbes1069:
Fix linking with math library for check_symbols_exists and correct getopt header.


drowe67:
Thanks @hobbes1069.  So this issue was particular to Fedora 30?  I haven't seen anything sim8lar on the various flavours of Ubuntu I use.

hobbes1069:
I guess so, I don't have another system to compare to. Perhaps the math library is implied on other systems? I may have to bootup Ubuntu in my VM and see. I'm not sure if I can dig into Travis CI enough to get the details.

drowe67:
Yep, builds fine on all Ubuntus I have tried.

hobbes1069:
Well it still built fine for me but configuration complained because the check_symbols_exists was not linking with the math library, however, that would affect all the "HAVE_<X>" definitions and I have no idea what differences in the resultant binaries there would be due to that.

drowe67:
All the tests run on Travis pass fine https://travis-ci.com/drowe67/codec2/builds/129861987, so I don't think there is a problem on Ubuntu.

drowe67:
@hobbes1069 just let me know when your OK with the Fedora 30 side and I'll merge.

hobbes1069:
This can get merged regardless but it looks like the problem with at least some of the tests is it assumes the build directory is in a subdirectory of the source... I create a build directory completely outside of the git directory. I'll what it'll take to use cmake variables so that won't matter.

Also, what is "pkg load signal"?

hogthrob:
https://octave.org/doc/v4.0.1/Using-Packages.html

drowe67:
"pkg load signal" is an Octave command to load the signal processing package (library), used to support some tests.

hobbes1069:
Guess I need to install octave then :)

hobbes1069:
Let's go ahead and merge this. I'll add a new PR to fix the tests. I'm probably going to have to move the shell files to a .in file and substitute in the correct path based on the build directory.

PR #80, opened 2019-10-28
bump version 0.9.1

ra1nb0w:
this bump permits to identify codec2 code that has fixed the "extern C"
problem.
For example, needed by gnuradio 3.8 to check if codec2 includes need
"extern C" (version 0.8.1) or not.

what do you think?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@ set(CODEC2_VERSION_MAJOR 0)
 set(CODEC2_VERSION_MINOR 9)
 # Set to patch level if needed, otherwise leave FALSE.
 # Must be positive (non-zero) if set, since 0 == FALSE in CMake.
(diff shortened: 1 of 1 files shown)


drowe67:
Actually we are coming up to a release, due to the freedv-gui 1.4 work. @hobbes1069 handles our releases,  @hobbes1069 are you OK with this version number bump?

hobbes1069:
Looks fine to me. No more changes to codec2 before the FreeDV 1.4 release?

drowe67:
> Looks fine to me. No more changes to codec2 before the FreeDV 1.4 release?

There will be some more changes, tracking down an acquisition bug on 2020...

hobbes1069:
If we don't want to wait, versions are "free". We can always release with 0.9.2 if needed.

ra1nb0w:
thank you!

PR #82, opened 2019-11-04
A version that uses flash constants rather than ram for W[] and w[] vectors

srsampson:
This is experimental right now


srsampson:
W[] used to be complex, but imaginary part was 0.0. Moving it to Flash saves 2048 bytes of RAM.
w[] is 1280 bytes of RAM. Total saving 3328 bytes.
We can also save RAM by moving the Parzan table to FLASH.

I highly edited the W[] table to make all 0.0 positive and values below .00008 were set to 0.0 also.

srsampson:
The problem with this mod is that it breaks 450PWB (pseudo-wideband) mode. While this mode isn't used currently in firmware, who knows, it might be desired later. In the 450PWB mode, the w[] vector is 640 length, and the hamming window stretched to fill it. This will create a completely different W[] vector, so two tables of constants would be needed.

drowe67:
Hi Steve, OK that's an interesting idea, I hadn't thought of the memory used there.  

As well as 450 bit/s c2sim also supports a 16 kHz mode for experimental x86 work, so useful to generate these tables at init time in x86 land.

As you said it's experimental, and we're OK for stm32 RAM at the moment, but if you ever want to take it a little further:
1. An init-time way to select the ROM tables or the current RAM based method, with the ROM method used for deeply embedded target, and RAM by default.  But we don't really want to modify the codec 2 API if we can help it, so maybe an extra function to swap ```c2->w``` etc to the ROM tables and free the previously malloc-ed memory.
1. A ctest to confirm identical results in decoded speech with and without the changes.  That's why an init time method is useful (rather than a #define), as you can run the code two ways and compare without recompiling.
1. I'd suggest not hand modifying any tables - too unpredictable and hard to measure the effect.

src/sine.c
```diff
@@ -497,9 +500,10 @@ float est_voicing_mbe(
 
         offset = FFT_ENC/2 - l*Wo*FFT_ENC/TWO_PI + 0.5;
 	for(m=al; m<bl; m++) {
-	    Am.real += Sw[m].real*W[offset+m].real;
-	    Am.imag += Sw[m].imag*W[offset+m].real;
-	    den += W[offset+m].real*W[offset+m].real;
+	    Am.real += Sw[m].real*W[offset+m];
+	    Am.imag += Sw[m].imag*W[offset+m];
+
+	    den += W[offset+m]*W[offset+m];
```
drowe67: Good catch - might be useful as a general mod (x86 and embedded) if W is always real.

src/c2sim.c
```diff
@@ -403,8 +405,6 @@ int main(int argc, char *argv[])
     codec2_fft_cfg  fft_fwd_cfg;
     codec2_fftr_cfg  fftr_fwd_cfg;
     codec2_fftr_cfg  fftr_inv_cfg;
-    float w[m_pitch];	        /* time domain hamming window            */
-    COMP  W[FFT_ENC];	/* DFT of w[]                            */
     MODEL model;
```
drowe67: this will break 16 kHz operation .....

src/codec2.c
```diff
@@ -182,8 +179,9 @@ struct CODEC2 * codec2_create(int mode)
 	c2->Sn_[i] = 0;
     c2->fft_fwd_cfg = codec2_fft_alloc(FFT_ENC, 0, NULL, NULL);
     c2->fftr_fwd_cfg = codec2_fftr_alloc(FFT_ENC, 0, NULL, NULL);
-    make_analysis_window(&c2->c2const, c2->fft_fwd_cfg, c2->w,c2->W);
+
```
drowe67: Rather than changing every single reference to w[] and W[] in ```codec2/src/*.c```, how about just (optionally) set up c2->w and c2->W to be pointers to your ROM table versions somewhere in the codec 2 init.

srsampson:
I appreciate all your comments. I did not know how difficult this would be, and I see now all the changes that would be needed. A simple idea gets buried in the reality.

Changing W[] to real saves memory, but I think it still has the same number of multiplies. I'll see if I can come up with just that in a modification, minus all this other stuff.

Yes, I got carried away with the table. I saw a lot of -0.0 and +0.0 so got rid of the negative zero's, and then noticed that the really small numbers didn't add much until you get closer to the impulse waveform. But I agree, the changes are a dangerous path.

srsampson:
P.S. I did do some audio tests using the 1300 vocoder and was looking for a type of noise during silence. To me the test of raw voice files sounded good though.

srsampson:
The pull request was to see if it compiled ok.

PR #83, opened 2019-11-05
Change W[] vector to float as imag always 0.0

srsampson:
I noticed the W[] imaginary values were always 0.0

### DIFF
--- a/misc/tnlp.c
+++ b/misc/tnlp.c
@@ -145,7 +145,7 @@ int main(int argc, char *argv[])
       dump_Sn(m, Sn); dump_Sw(Sw);
       #endif
 
-      f0 = nlp(nlp_states, Sn, n, &pitch_samples, Sw, &prev_f0);
+      f0 = nlp(nlp_states, Sn, n, &pitch_samples, Sw, W, &prev_f0);
 
       fprintf(stderr,"%d %f %f\n", frames++, f0, pitch_samples);
       fprintf(fout,"%f %f\n", f0, pitch_samples);
--- a/misc/tnlp.c
+++ b/misc/tnlp.c
@@ -91,7 +91,7 @@ int main(int argc, char *argv[])
     kiss_fft_cfg  fft_fwd_cfg;
     COMP  Sw[FFT_ENC];	        /* DFT of Sn[] */
     float w[m];	                /* time domain hamming window */
-    COMP  W[FFT_ENC];	        /* DFT of w[] */
+    float W[FFT_ENC];	        /* DFT of w[] */
     float pitch_samples;
     int   i;
     float f0, prev_f0;
@@ -145,7 +145,7 @@ int main(int argc, char *argv[])
       dump_Sn(m, Sn); dump_Sw(Sw);
       #endif
 
-      f0 = nlp(nlp_states, Sn, n, &pitch_samples, Sw, W, &prev_f0);
+      f0 = nlp(nlp_states, Sn, n, &pitch_samples, Sw, &prev_f0);
 
       fprintf(stderr,"%d %f %f\n", frames++, f0, pitch_samples);
       fprintf(fout,"%f %f\n", f0, pitch_samples);
--- a/src/c2sim.c
+++ b/src/c2sim.c
@@ -404,7 +404,7 @@ int main(int argc, char *argv[])
     codec2_fftr_cfg  fftr_fwd_cfg;
     codec2_fftr_cfg  fftr_inv_cfg;
     float w[m_pitch];	        /* time domain hamming window            */
-    COMP  W[FFT_ENC];	/* DFT of w[]                            */
+    float W[FFT_ENC];	/* DFT of w[]                            */
     MODEL model;
     float Pn[2*N_SAMP];	/* trapezoidal synthesis window          */
     float Sn_[2*N_SAMP];	/* synthesised speech */
@@ -584,7 +584,7 @@ int main(int argc, char *argv[])
 
 	\*------------------------------------------------------------*/
 
-        nlp(nlp_states, Sn, N_SAMP, &pitch, Sw, W, &prev_f0);
+        nlp(nlp_states, Sn, N_SAMP, &pitch, Sw, &prev_f0);
 	model.Wo = TWO_PI/pitch;
 
         dft_speech(&c2const, fft_fwd_cfg, Sw, Sn, w);
--- a/src/c2sim.c
+++ b/src/c2sim.c
@@ -584,7 +584,7 @@ int main(int argc, char *argv[])
 
 	\*------------------------------------------------------------*/
 
-        nlp(nlp_states, Sn, N_SAMP, &pitch, Sw, &prev_f0);
+        nlp(nlp_states, Sn, N_SAMP, &pitch, Sw, W, &prev_f0);
 	model.Wo = TWO_PI/pitch;
 
         dft_speech(&c2const, fft_fwd_cfg, Sw, Sn, w);
--- a/src/codec2.c
+++ b/src/codec2.c
@@ -2520,7 +2520,7 @@ void analyse_one_frame(struct CODEC2 *c2, MODEL *model, short speech[])
 
     /* Estimate pitch */
 
-    nlp(c2->nlp, c2->Sn, n_samp, &pitch, Sw, c2->W, &c2->prev_f0_enc);
+    nlp(c2->nlp, c2->Sn, n_samp, &pitch, Sw, &c2->prev_f0_enc);
     //PROFILE_SAMPLE_AND_LOG(model_start, nlp_start, "    nlp");
 
     model->Wo = TWO_PI/pitch;
--- a/src/codec2.c
+++ b/src/codec2.c
@@ -2520,7 +2520,7 @@ void analyse_one_frame(struct CODEC2 *c2, MODEL *model, short speech[])
 
     /* Estimate pitch */
 
-    nlp(c2->nlp, c2->Sn, n_samp, &pitch, Sw, &c2->prev_f0_enc);
+    nlp(c2->nlp, c2->Sn, n_samp, &pitch, Sw, c2->W, &c2->prev_f0_enc);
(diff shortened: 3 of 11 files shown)


srsampson:
OK, I screwed up with nlp.c where I didn't think W[] was used anymore, but I see now that the #ifdef POST_PROCESS_MBE option does use it. Will add that back and make sure all COMP are now float.

drowe67:
Hi Steve, code looks good.  Any thoughts on how we can test to ensure this hasn't affected codec 2 in some subtle way?  Not sure if any of the ctests cover this low level analysis?

An idea for a test would be to compare the "Am" amplitude estimates before and after this change for a test input.  c2sim can extract text files of codec 2 model parameters, which can be read by Octave and easily compared.  
  
Might be other ways to test too. Doesn't have to be an automated ctest, just something we run once before merging this PR is OK.

srsampson:
I'll see if I can come up with something. I'm thinking some inline code that can be optionally enabled during test.

srsampson:
Well that was a complete waste of time. Sorry for the wasted effort. I've been toying with this concept for a few months off and on, and I guess I never went back to verify the W[] values. Now that I did, it shows the imaginary values are NOT 0.0 after the FFT like I must have measured in error before.

I think I must start seeing things when I try to start saving bytes.

values

srsampson:
Looking back on why I started this, I think before the FFT you fill the array with 0.0 in the imaginary. So I was thinking then that the imaginary values of the FFT output were not necessary either.

But I think the complex math is beneficial over just the real values now.

drowe67:
I still think you might be onto something, this rings a bell from my distant past.  A real symmetric (about t=0) sequence should have a DFT with complex part == 0.  If some parts are very small, might just be an off by one error or round off.  I think the sequence has to be odd in length.

I'm off on a short holiday until Sunday, so won't be on line much. But happy to help you with this when I return.

srsampson:
That's interesting. I was using a Java test program with its FFT routine, and so threw that out, and just used the C algorithms of codec2.

I sometimes use Java for quick and dirty tests. That bit me here, because in-fact kiss FFT gives all zero's as I remembered. In actuality some are -0.0 and some are +0.0  which is one of those IEEE things I guess... So I'm happier now. Onward with some more tests.

srsampson:
Oh! P.S. Happy Holiday! I'm wearing a sweater because of arctic conditions :-)

srsampson:
OK, good news. The only place that would cause a problem is nlp.c in algorithm which is currently not used because POST_PROCESS_MBE is hardwired undefined. But just for testing I enabled it and compared the calculations in test_candidate_mbe() and the same results of Am, den, and error were received using float or COMP.

The rest of codec2 only uses the W[].real part in calculations.

For example, estimate_amplitudes() in sine.c does an Am calculation only using the real part. But in fact the calculation is never used anywhere. It's dead code.

est_voicing_mbe() in sine.c just uses the real part.

So this change would have no change in raw output. Whew...

srsampson:
Interestingly, I swapped out my Java FFT with another one, and it gave the same answer as the first (well, in this one I plotted 256-511 first, like I should have). So far I still believe KISS FFT over both of those however.
new

drowe67:
Nice work @srsampson :+1: 

> OK, good news. The only place that would cause a problem is nlp.c in algorithm which is currently not used because POST_PROCESS_MBE is hardwired undefined. But just for testing I enabled it and compared the calculations in test_candidate_mbe() and the same results of Am, den, and error were received using float or COMP.

Yes perhaps we can rm that POST_PROCESS_MBE  code.  It'll still be in the Git graveyard in the unlikely event it's needed again.

> For example, estimate_amplitudes() in sine.c does an Am calculation only using the real part. But in fact the calculation is never used anywhere. It's dead code.

Could you point out the dead code pls?  IIRC think the {Am} are used for Codec 2 700C, but it's been a while since I looked at that.

> [edit] In review, this change doesn't save much CPU, but saves 512 * sizeof(float) bytes of RAM.

That's 2k of run time memory, so a nice saving :+1: 

Re the Java code suspect it's a subtle set up error, e.g. care and feeding of how you load an odd symmetric sequence.  Here's an 8 point DFT example in Octave of the window 0.25 0.5 1 0.5 0.25, you need to modulo Nfft shift the samples to get the odd symmetry to work:
```
octave:5> sum(imag(fft([1 0.5 0.25 0 0 0 0.25 0.5])))
ans =     0.0000e+000
```

srsampson:
Dead code:

https://github.com/drowe67/codec2/pull/83/commits/80bcca386b4dbe5a4f77a6d19e9de8d8d5db20d4

srsampson:
> octave:5> sum(imag(fft([1 0.5 0.25 0 0 0 0.25 0.5])))
> ans =     0.0000e+000

ooh Octave! Good idea, I should have tested there first.

srsampson:
> Yes perhaps we can rm that POST_PROCESS_MBE code. It'll still be in the Git graveyard in the
> unlikely event it's needed again.

Let me know after your vacation, and I'll remove it in this branch.

srsampson:
I ran a small test:

```
xxd master-hts1a.raw master.txt
xxd ss-wreal-hts1a.raw ss-wreal.txt
diff -s master.txt ss-wreal.txt
Files master.txt and ss-wreal.txt are identical
```

drowe67:
Looks good :+1: 
```
git checkout master && make c2sim
./c2sim ../../raw/hts.raw --dump hts
git checkout ss-wreal && make c2sim
./c2sim ../../raw/hts.raw --dump hts_wreal
diff hts_model.txt hts_wreal_model.txt
(silence)
```
That c2sim option dumps some text files, the ``hts_model`` text files has pitch (Wo), number of harmonics L, and the set of L sinusoidal amplitudes {Am}.  So a diff of this file confirms the Am estimation hasn't been messed up.

Shall we merge now, or did you want to make some more changes?

drowe67:
@srsampson - Actually we should run are the stm32 unit tests before merging.  If you have a Discovery card you can run them, or I can run them when I get home.

srsampson:
I seem to be an idiot when it comes to the stm32 stuff, so I'll let you do it.

Did you want me to remove the POST_PROCESS_MBE code, or wait on that?

drowe67:
> Did you want me to remove the POST_PROCESS_MBE code, or wait on that?

Sure :+1:

srsampson:
Still have:

```
diff -s hts_model.txt hts-wreal_model.txt
Files hts_model.txt and hts-wreal_model.txt are identical
```
After last change.

srsampson:
I decided to leave the W[] parameter in call to nlp() as it is used in a lot of places.
Who knows we might want to use it again, and it is probably optimized out anyway.

drowe67:
Hi @srsampson, I'm back from my holiday (got a little sailing in), and trying out this PR on the stm32:
```
cd odec2/stm32/build_stm32
cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/STM32_Toolchain.cmake   -DPERIPHLIBDIR=~/Downloads/STM32F4xx_DSP_StdPeriph_Lib_V1.8.0 ..
make
<snip>
Scanning dependencies of target tst_codec2_fft_init
[ 88%] Building C object unittest/src/CMakeFiles/tst_codec2_fft_init.dir/tst_codec2_fft_init.c.obj
/home/david/codec2/stm32/unittest/src/tst_codec2_fft_init.c: In function 'main':
/home/david/codec2/stm32/unittest/src/tst_codec2_fft_init.c:88:39: error: request for member 'real' in something not a structure or union
                 (double)codec2->W[j+i].real);
                                       ^
/home/david/codec2/stm32/unittest/src/tst_codec2_fft_init.c:89:38: error: request for member 'real' in something not a structure or union
         if (!float_cmp(codec2->W[j+i].real, expect_W[i])) {
```

So looks like a few unittests in stm32 land that need tweaking.

@db4ple - you know I should probably set up the ```codec2/.travis.yml``` to at least build the stm32 code automatically and trap these sorts of problems, I know running the UTs is problematic.  Might add that this this PR

donmr:
I copied code from the standalone programs like c3enc and c2dec for these 
tests.  If this is in git I can take a look.

Don

srsampson:
OK, I fixed that file and it compiled to completion. Lots of warnings, but it produced an sm1000v2 files:
```
-rwxr-xr-x  1 ssampson ssampson 1407876 Nov 10 16:05 sm1000v2.elf
-rwxr-xr-x  1 ssampson ssampson  293712 Nov 10 16:05 sm1000v2.bin
-rw-r--r--  1 ssampson ssampson  781239 Nov 10 16:05 sm1000v2.hex
-rw-r--r--  1 ssampson ssampson  294021 Nov 10 16:05 sm1000v2.dfu
```

donmr:
I checked out ss-wreal.  There is one failure on the PC tests:

The following tests FAILED:
	 11 - test_OFDM_modem_time_sync_700D (Failed)

The stm32 code compiles and passes many tests.
I'm having some trouble with python, PYTHONPATH, and numpy but
that is my system.

Don

srsampson:
On test 11 I got a pass. Running manually, I got:

time_to_sync: 0.800000

FYI

drowe67:
> I checked out ss-wreal. There is one failure on the PC tests: The following tests FAILED: 11 - test_OFDM_modem_time_sync_700D (Failed) The stm32 code compiles and passes many tests. I'm having some trouble with python, PYTHONPATH, and numpy but that is my system. Don

Hi @donmr, yes that ctest requires python and numpy.  ```ctest -V -R test_xxx``` will let you see a bit more.  Travis must have liked it, as all x86 tests are run when we push.

drowe67:
Passes all stm32 tests for me too, nice work Steve :+1: I'll merge

Next step is to complete https://github.com/drowe67/codec2/pull/58, to include these changes and some sync/EQ improvements.

donmr:
Yes it was an issue of versions on my system.  Once fixed they all passed.

I'll look at PC test 11 this morning.

Don

donmr:
Yes fixing the python/nump versions fixed x86 test #11 too.

All working here now.

Don

PR #85, opened 2019-11-12
Standardize Include files

srsampson:
Change include files to have the IFDEF wrapper first, then the include files, and then the C++ wrapper.

In order to standardize, and clean up some files that had it backwards.

### DIFF
--- a/src/codec2.h
+++ b/src/codec2.h
@@ -26,14 +26,14 @@
   along with this program; if not, see <http://www.gnu.org/licenses/>.
 */
 
-#ifdef __cplusplus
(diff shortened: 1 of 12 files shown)


srsampson:
Possibly consider replacing the IFDEF at the top with:

#pragma once

This is compatible with 99% of modern compilers (GCC, Clang, ARM, etc).

drowe67:
Looks like a nice clean up, thanks @srsampson Let me know when you're good to merge

srsampson:
Tested and ready to merge, thanks.

PR #88, opened 2019-11-29
fix failing assertion

goblin:
fread() returns number of items read, not number of bytes. This assert
could never have succeeded.

### DIFF
--- a/src/c2dec.c
+++ b/src/c2dec.c
@@ -112,7 +112,7 @@ int main(int argc, char *argv[])
     char *ext = strrchr(argv[2], '.');
     if ((ext != NULL) && (strcmp(ext, ".c2") == 0)) {
         int nread = fread(&in_hdr,sizeof(in_hdr),1,fin);
-        assert (nread == sizeof(in_hdr));
+        assert (nread == 1);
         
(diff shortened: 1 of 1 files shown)


drowe67:
@goblin thanks - this is something that was contributed I've rarely used myself.  The syntax of the command isn't ideal:
```
./c2enc 700C ~/codec2/raw/hts1a.raw hts1a.c2
./c2dec 1600 ~/codec2/build_linux/src/hts1a.c2 /dev/null
```
.. as one still has to put *something* into the 1st (mode) argument, the.c2 processing then overrides that.  Oh well.

I'll merge and add a ctest for using ```.c2``` files.

goblin:
Thanks for merging :-)

`c2enc`'s usage actually encourages to use the .c2 extensions, and it made sense for me to name the file that way. I wasn't even aware it'd add extra headers as compared to naming it .bit (that's strange on its own), but obviously the failing assertion made it undecodable before the fix :-)

drowe67:
Thanks for fixing the bug, I think it had popped up before.  Everything I do with Codec 2 involves streaming the bits into other programs (like modems)

PR #17, opened 2019-04-07
cohpsk modem maintenance

drowe67:
Mike Riley has pointed out that this modem, as used for FreeDV 700C, appears to have applications for high Doppler rate channels.  So Steve and David are doing a little maintenance:

1. Making sure the scripts run.
1. Replacing the stored-file-based HF channel simulation with on the fly doppler spread generation like ofdm.
1. Creating Ctests for AWGN and Octave<->C port.
1. Write a README_cohpsk with some examples.

PR #94, opened 2020-01-20
Fix reversion caused by change in graycode variable initialziation

mcdermj:
While doing some work to rework the Flex Radio 6000 "Waveform" for FreeDV I encountered an issue where the audio was completely garbled.  After checking my code for the usual suspects, I started looking into the codec2 library itself.

After bisecting the code and testing, I found that a regression was introduced in c515387982bbe9f0ea9d6d0f50884716c0b1c6c4 that apparently causes the issue.  This change should fix the problem.  I have tested it in my code under 1600 and it looks good so far.  Have not checked other modes, but the change logically makes sense.

### DIFF
--- a/src/codec2.c
+++ b/src/codec2.c
@@ -216,7 +216,7 @@ struct CODEC2 * codec2_create(int mode)
         c2->bpf_buf[i] = 0.0;
 
     c2->softdec = NULL;
-    c2->gray = 0;
+    c2->gray = 1;
     
     /* newamp1 initialisation */
 


drowe67:
Nice catch @mcdermj - took me a while to reproduce it with c2enc/c2dec as they explicitly set gray to 1.  But I managed to get the garbled speech when trying a decode a FreeDV 1600 modem sample file with freedv_rx.  Thanks :+1: 

BTW we found an issue in FreeDV 700C a few months back in some code that has been optimised for the embedded CPU being used by Flex.  We had to remove the optimisation in https://github.com/drowe67/codec2/pull/17/commits/b24f40c40b72335fa2d9dda2c5e97cb27d3c9bb4 as it was breaking the modem.  Just mentioning this in case 700C is running slow on Flex machines.

mcdermj:
Thanks for the heads up @drowe67.  I was actually looking hard at the 700C code because it runs *really* slow on ARM right now.  I get similar results on the RPi.  I'm mainly in "brush up functionality" mode right now and getting ready for an initial release of the new code, but I definitely need to look at optimizations.  I'd also like to take a look at `down_convert_and_rx_filter()` in `fdmdv.c` because it's the big consumer of CPU in my profiling on the 1600 mode.  Have we thought about doing FIRs with an FFT approach rather than the MAC method?  My understanding is that with general purpose CPUs with SIMD instructions it can be quite a bit faster.  I may try it out if I have some time in the future.

drowe67:
Well the term "ARM" covers a lot of processors, internal architectures and clock speeds.  

We have optimised FreeDV 1600 and 700D such that they run happily on a 168 MHz stm32 ([700D port blog post](http://www.rowetel.com/?p=6835)), the code is in codec2/stm32.

Here are the results I get on a Pi for a 60 second audio file:
```
time ./freedv_tx 1600 ~/codec2/raw/ve9qrp.raw - - | ./freedv_rx 1600 - /dev/null
<snip>
real	0m5.606s
time ./freedv_tx 700C ~/codec2/raw/ve9qrp.raw - - | ./freedv_rx 700C - /dev/null
<snip>
real	0m7.726s
```

So both around 10x real time on my Pi.

I can help you out with some tests designs if you want to try some optimisations.  This needs to be done carefully, to avoid breaking the system in subtle ways.

PR #96, opened 2020-02-08
Add Horus_demod IQ input and unit test.

darksidelemm:
This PR adds the following features into horus_demod / horus_api:
- Ability to accept s16 complex samples as input.
- Ability to set upper and lower frequency estimator limits.

I have added a s16 sample of a horus binary signal, and created a unit test which uses it.

This does add an argument to the horus_rx function call to add the quadrature input switch, which affects freedv. A PR will be raised shortly to change this (and do other documentation updates relating to the Horus Binary mode).


darksidelemm:
Note that https://github.com/drowe67/freedv-gui/pull/46 makes the corresponding changes in freedv.

src/horus_demod.c
```diff
@@ -65,10 +68,12 @@ int main(int argc, char *argv[]) {
             {"help",      no_argument,        0, 'h'},
             {"mode",      required_argument,  0, 'm'},
             {"stats",     optional_argument,  0, 't'},
+            {"fsk_lower", optional_argument,  0, 'b'},
+            {"fsk_upper", optional_argument,  0, 'u'},
             {0, 0, 0, 0}
```
drowe67: Should this be required_argument?  The you won't need the ```if (optarg != NULL){``` below.

src/horus_demod.c
```diff
@@ -65,10 +68,12 @@ int main(int argc, char *argv[]) {
...
         };
         
-        o = getopt_long(argc,argv,"hvcm:t::",long_opts,&opt_idx);
+        o = getopt_long(argc,argv,"hvcqm:t::",long_opts,&opt_idx);
```
drowe67: Adding ```b:u:``` to this string means required argument.

CMakeLists.txt
```diff
@@ -380,6 +380,12 @@ endif()
              set_tests_properties(test_horus_binary PROPERTIES
              PASS_REGULAR_EXPRESSION "1C9A9545")
 
+    add_test(NAME test_horus_binary_iq
+             COMMAND sh -c "cat ${CMAKE_CURRENT_SOURCE_DIR}/raw/horusb_iq_s16.raw | $<TARGET_FILE:horus_demod> -q -m binary --fsk_lower=1000 --fsk_upper=20000 - -"
+             )
+             set_tests_properties(test_horus_binary_iq PROPERTIES
+             PASS_REGULAR_EXPRESSION "000900071E2A000000000000000000000000259A6B14")
+
     # tests ldpc_enc/ldpc_noise/ldpc_dec
```
drowe67: Nice :+1:

drowe67:
Looks good @darksidelemm, please take a look at those small issues with the command line parsing and we're good to merge.

src/horus_demod.c
darksidelemm: fsk_lower and fsk_upper are both optional arguments, and if present (detected by fsk_lower >0) override whatever the fsk demod sets up on initialisation.
Have I done the argument parsing correctly in this case?

drowe67: Yes you are right - I think what we need is the ```b:u:``` in the comment below, to force the user to enter the numerical arguments for the ```fsk_lower``` and ```fsk_upper ``` options.

darksidelemm:
I still can't quite make it work correctly unfortunately.

I've set the fsk_lower and fsk_upper to have required arguments, but due to the way horus_demod takes the required arguments at the end of the call, it doesn't quite work correctly.

For example, running:
`./horus_demod -q -m binary --fsk_lower --fsk_upper=1000  - -`
results in the the '--fsk_upper=1000' string being parsed as the argument for --fsk_lower, which of course fails, resulting in the default estimator limits being used.

Running `./horus_demod -q -m binary --fsk_lower=100 --fsk_upper  - -`
results in the first '-' being parsed as the argument to --fsk_upper, resulting in a 'Too few arguments' message due to both input and output not being defined.

TBH I think this behaviour is probably OK for now. It's probably possible to fix this with more messing about with the option values, but I don't think it's worth the effort at the moment.

srsampson:
I believe you need something like (numbers pulled out of a hat):
```
./horus_demod -q -m binary --fsk_lower 1200 --fsk_upper 2200 - -
```
That is, the arguments require a value to be given. Not sure of the "=" sign off the top of my head.

darksidelemm:
Sure - with the intended arguments it works fine. The issue was more around how to make it fail in an appropriate manner if no arguments are provided.
Since these options are really only used in various scripts that I've written , I don't consider it to be a big issue...

PR #97, opened 2020-02-17
Bug and Compiler fixes

srsampson:
Fix bug in phase.c and compiler changes to STM32

### DIFF
--- a/src/phase.c
+++ b/src/phase.c
@@ -241,6 +241,13 @@ void mag_to_phase(float phase[],             /* Nfft/2+1 output phase samples in
     int  Ns = Nfft/2+1;
     int  i;
 
+    /* initialize array */
+
+    for(i=0; i<Nfft; i++) {
+        Sdb[i].real = 0.0;
+        Sdb[i].imag = 0.0;
+    }
+
     /* install negative frequency components, 1/Nfft takes into
        account kiss fft lack of scaling on ifft */
 
--- a/stm32/src/si53xx.c
+++ b/stm32/src/si53xx.c
@@ -42,6 +42,7 @@
 #include "math.h"
 #include "si53xx.h"
 
+T_Si5351_data   Si5351_Config;
 
 void si5351_write(uint8_t REGaddr, uint8_t data) {
     // ignoring errors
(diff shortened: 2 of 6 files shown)


src/phase.c
```diff
@@ -241,6 +241,13 @@ void mag_to_phase(float phase[],             /* Nfft/2+1 output phase samples in
     int  Ns = Nfft/2+1;
     int  i;
 
+    /* initialize array */
+
+    for(i=0; i<Nfft; i++) {
+        Sdb[i].real = 0.0;
+        Sdb[i].imag = 0.0;
+    }
+
```
drowe67: Hi @srsampson - so why is this needed given the next few lines that fill ```Sdb[]```?  Do they miss some values or did the compiler complain?

stm32/src/si53xx.c
```diff
@@ -42,6 +42,7 @@
 #include "math.h"
 #include "si53xx.h"
 
+T_Si5351_data   Si5351_Config;
```
drowe67: As you suggested on the mailing list, I'm OK simply rm-ing this and other sm2000 specific code.  It will still be there in the Git basement if we ever want it.

drowe67:
@srsampson I think the Travis fail is an intermittent LDPC test failure I've also been experiencing. It's probably a precision thing.  I will need to sort that out some time.

Also - for any stm32 PRs:
1.  We need to run and pass all of the stm32 unit tests
1.  And test the sm1000 binary on a SM1000, e.g. try decoding some canned files (70)D and 1600), and throw it out to sm100 users for testing.

src/phase.c
srsampson: I believe the next few lines only initialize 0 - Ns while the array is 0 - Nfft thus half the values I assume will be random stack values.

This seemed suspicious to me.

srsampson: > I believe the next few lines only initialize 0 - Ns while the array is 0 - Nfft thus half the values I assume will be random stack values.
> 
> This seemed suspicious to me.

Oh darn, I have made a BIG mistake here. I completely missed the code inside the loop!

I feel maybe this whole PR is a wash. Probably needs deleted. argh!

PR #99, opened 2020-03-18
Use symmetrical hann in FSK

srsampson:
The current Hann table is not symmetric and goes negative in the last two indexes, and also goes over 1.0 in the center. The replacement formula was taken from elsewhere in the library and produces symmetrical values.

### DIFF
--- a/src/fsk.c
+++ b/src/fsk.c
@@ -95,18 +95,8 @@ static void fsk_generate_hann_table(struct FSK* fsk){
     int Ndft = fsk->Ndft;
     size_t i;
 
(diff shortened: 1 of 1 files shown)


drowe67:
@srsampson sorry I just noticed this PR, will review within 24 hrs and get back to you :+1:

srsampson:
> @srsampson sorry I just noticed this PR, will review within 24 hrs and get back to you

At your leisure !

drowe67:
@srsampson thanks.  Not sure why this was written using a recursion, as I guess it's only called at init time so CPU is not an issue.

PR #101, opened 2020-03-28
HorusBinary Updates & Doppler Testing

darksidelemm:
Working branch for updates to the Horus Binary modem, and the FSK demod in general.

Added so far:
- Test script (octave/fsk_demod_BER_test.py) to do BER testing on C implementation of fsk_mod and fsk_demod, including simulation of LEO satellite doppler on the modem signal.
- sat_doppler.py script, to generate doppler data for the above.

Changed:
- Default frequency estimator settings in fsk.c for the low-bit-rate mode have been widened.
- horus_api.c has changes:
  - horus_open now takes a baud rate argument.

These changes should not preclude the use of the modem at higher baud rates, for example as an upgrade to the existing Horus Binary mode.

# Project Lockdown - Low MDS Open Source Telemetry Waveform 

Our goal is to develop a FSK/LDPC telemetry system that can decode telemetry packets at a very low MDS on a sateillte AWGN + Doppler channel.  Use cases:
1. Recovery of HAB payloads on the ground where their LOS signal is blocked from terrestrial stations, but may be detectable from LEO satellites passing overhead.
1. Low MDS Telemetry for HAB missions.
1. Low SNR Terrestrial HF/VHF/UHF packet based communications.

# Requirements

1. Packet duration of roughly 20 seconds.
1. 128 and 256 bit payload options.
1. Tx capable of running on a reprogrammed RS41 STM32F100 (24 MHz with 64K flash, 8K RAM)
1. Tx modulation mFSK (e.g. Si4032 PLL controlled by STM32F100), 271Hz shift between tones.
1. Can run at Fs=24 kHz with Rs=25 Hz (large Fs/Rs ratios can be challenging).
1. Complex sample input, e.g. Fs=24kHz gives +/-12kHz range.
1. Rx acquisition/demodulation at +/- 10 kHz from carrier frequency (assuming no other CW interferers in passband).
1. Ability to track frequency change of +/- 0.01*Rs Hz/s.  In the satellite/HAB use case we assume the bulk of satellite Doppler is corrected externally to demod, leaving just a small residual rate/change.
1. Rx MDS < -150 dBm

# Waveform Design

```
256 payload bits, rate 1/3 code
256*3 = 768 bit codeword
Rb=768/20=38.4 bits/s (20s codeword duration)

Code will start to decode valid packets at around coded BER=5E-3 (1 error per codeword), which is Eb/No=1.5dB on a PSK channel.
The raw error rate can be calculated as:
EbNodB=1.5; rate=1/3; EbNodBraw=EbNodB+10*log10(rate); EbNo=10^(EbNodBraw/10); BERraw = 0.5*erfc(sqrt(EbNo)) = 0.165
Using 4FSK we get a raw error rate of 0.165 at  Eb/No=1.5dB (coincidentally the same as BPSK coded Eb/No!)

MDS=Eb/No+10*log10(Rb)+NF-174
   =1.5+10*log10(38.4)+1-174=-155.6dBm
SNR=Eb/No+10log10(Rb/B)
   =1.5+10*log10(38.4/3000)=-17.4dB (3kHz noise bandwidth)
```
# Project Plan

| Milestone | Who | Achievement | Status |
| --- | --- | --- | --- |
| 1 | @darksidelemm   | Integrated C modem works down to Eb/No = 1dB for 4FSK IL < 0.5dB | |
| 2 | @darksidelemm | As per M1 with external Doppler correction, equivalent to a maximum rate of change of ~80Hz/s (max doppler at 436 MHz) | |
| 3 | DR | Octave simulation of Freq Est BER IL<0.5dB Fs=24000 Rs=25 dF=+/-0.01*Rs || 
| 4| DR | Octave simulation of complete modem BER IL<0.5dB Fs=24000 Rs=25 dF=+/-0.01*Rs || 
| 5 | DR  | C port of freq est, complex freq, mask algorithm, C unit test passes | |
| 6 | | integration of freq est with C modem, API selection of freq ests, existing tests pass  | |
| 7 | | Integrated C demod BER IL<0.5dB Fs=24000 Rs=25 dF=+/-0.01*Rs | |
| 8 | WGC | rate TBD n=128 and n=256 (k,n) LDPC codes developed | |
| 9 |  | LDPC code integrated into C framework, passes unit tests | |
| 10 |  | LDPC code integrated with 4FSK modem, PER<0.5 at raw 4FSK Eb/No=2dB  Fs=24000 Rs=25 dF=+/-0.01*Rs | |
| 11 | @darksidelemm | RS41 Tx developed and functional test at high Eb/No ||
| 12 |  | bench test MDS testing at -155dBm (!) ||
| 13 |  | OTA HAB flight test ||


darksidelemm:
A few notes on testing methodology:
- fsk_put_test_bits has a frame detection threshold. This threshold effectively sets an upper bound on the measurable BER.
- This frame detection threshold can be increased, but the frame size must also be increased, else frames are incorrectly detected, causing the BER calculation to be in error.
- Even with large frame size, setting a threshold > ~0.45 results in far too many false detections.
- I have settled on a frame size of 1000, and a threshold of 0.4 for now. 
- In low Eb/N0 cases, it is common that put_bits does not detect all frames in a test signal. In this case, the known number of sent bits is used to correct the result. Some grace is given for the first frame in the test signal, to allow the modem to 'spool up'.

In my test script, I am generating the test signals as real samples, then reading them into python, converting to complex with a hilbert transform, then applying noise and saving in c8 format, with signal levels normalised to 1.0.  csdr is used to convert the c8 format to complex s16 for feeding into fsk_demod. 
The use of the complex intermediate format is so I can easily apply doppler profiles for later analysis. 

Some initial performance observations based on this code:
- 100 baud 4FSK, with a 250 Hz tone spacing (minim available on the target transmit platform), has between ~0.5 dB implementation loss > 3dB Eb/N0, increasing below this. With the same 250 Hz tone spacing, 50 baud 4FSK has ~2dB implementation loss, and 25 baud 4FSK is essentially broken. See the following figure:
Screen Shot 2020-03-29 at 6 01 32 pm
- The low baud rate performance issues do not appear to be due to frequency estimation. For example, here are the frequency estimator outputs for 25 baud at 18 dB Eb/N0, at which we get a BER of ~0.25:
Screen Shot 2020-03-29 at 6 13 28 pm
- Suspect the issue might be in the timing estimator.

drowe67:
@darksidelemm good work :+1:  yes I reckon you might be right about the timing estimator.  Lets check if if it's a fixed time window, or normalised by Rs to a fixed number of symbols.

darksidelemm:
@drowe67 before I go too much further, could you please review my noise addition code? 
https://github.com/drowe67/codec2/blob/horusbinary_v2/octave/fsk_demod_BER_test.py#L184

Note that I am using complex samples throughout (applying a hilbert transform to the samples generated by fsk_mod). The comment about 'real' samples is regarding if I was passing real samples through this function (which is generally expecting complex samples). That isn't the case for these tests.

It's fairly critical that this bit of code is working correctly before we read too much into any results. I'm fairly sure it's OK (it's based off some octave code you wrote a while ago, and I've used it in other applications), but it's worth a check...

drowe67:
> @drowe67 before I go too much further, could you please review my noise addition code?

Sure - will review on my day off Friday :smile:

drowe67:
@darksidelemm Just checking it's the [add_noise](https://github.com/drowe67/codec2/blob/35f896927f6575ff95a31d1d8a0b5d9174cbd0b2/octave/fsk_demod_BER_test.py#L194) function I should look at?

Link above is for ```calculate_variance```

drowe67:
@darksidelemm I wrote a test to exercise your channel simulation function:
```
~/codec2/octave$ python3 fsk_demod_BER_test.py --test
nb_errors: 1183 ber: 0.012 ber_lower_limit: 0.011 ber_upper_limit: 0.015
AWGN channel simulation test PASSED!
```

drowe67:
Building up [fsk_test_freq_est.m](https://github.com/drowe67/codec2/blob/39fde2780929aa670d462ef63f0c4f7e298bd6e3/octave/fsk_test_freq_est.m) for the freq offset estimator to support optimisation for out low E/No operating points.  

Uses arbitrary values for Fs and Rs for now.

Runs in single Eb/No mode to give a time domain plot:
fsk_freq_est_single

Can also draw curves:
fsk_freq_est_curve

drowe67:
@darksidelemm me-thinks we may need some algorithm enhancements to get this freq offset estimator working.  One idea is to use some known information, like the tone spacing from the RS41.  We could then drag a comb type window over the FFT, and look for the maxima where all four tones pop up.  Current algorithm can handle **any** random tone spacing which might be overkill in many applications.

darksidelemm:
Yes - the current algorithm is actually quite neat in that it will work with *anything*, but yes - it can get confused easily (e.g. strong interferers). I wouldn't want to drop the current functionality, but I agree that some kind of comb-based approach would probably perform better.

drowe67:
@darksidelemm: In https://github.com/drowe67/codec2/pull/101/commits/3305834eabc29f1b99cb5ffd645253f5e58a81c3 is a candidate for a new freq offset estimator:

fsk_freq_est_single
fsk_freq_est_curve

Notes:
1. The freq estimator works on a 50 symbol window.  However depending on the data, there could be different number of each symbol in that window, so the lines for each tone may be of different heights, some easier than others to pick out of the noise.
1. The "mask" should ideally be the shape of DFT-ed hanning window, I'm using a rectangular shape as an approximation.  This might break for different Fs/Rs ratios, we shall see.
1. The spacing of tones in the "mask" assumes a constant known tone separation.  We'll need to test this assumption for real world transmitters.  Might be a good idea to simulate a worst case Tx too, e.g. insert some jitter in Tx tone freqencies. 
1. There's a start up transient due to the IIR smoothing (Beta=0.95) of the DFT mag spectra, note in the first plot above "mask" is pretty good after T=20 frames.
1. Note the "mask" method fails benignly with an offset of -Rs from ideal (just one tone spacing down).  We could try decoding the two top candidates of the "mask" freq est, see if the decoder converges.

darksidelemm:
Looks great David!

Assumption 3 should be valid, especially true for the target application (the RS41 transmitter), and of course true for signals generated by fsk_mod.

How do you mean fails benignly? As in, it locks on at +/- one tone spacing, and then won't lock on correctly after that?

drowe67:
> How do you mean fails benignly? As in, it locks on at +/- one tone spacing, and then won't lock on correctly after that?

Well if fails with an error of +Rs or -Rs, rather than some random estimate (see red line in top plot).  So we could try running the demod/FEC decoder on 3 candidate freq offset estimates, and see if we get a valid FEC decode.  This would effectively allow us to correct for the occasional freq est error at modest (due to our low bit rate) CPU cost.

The mask algorithm appears to return to the correct estimate immediately after any error, it doesn't get stuck for long.

drowe67:
@darksidelemm I have added the ability to measure BER, here are some results for Fs=8000, Rs=100:

The first plot is a "single" Eb/No point run, note after about 10 estimates (frames is a misnomer) it settles down, so I start measuring BER from there:
fsk_freq_est_single

Looks the like 10% error threshold isn't a great indication for BER performance, which is pretty good for the "mask" algorithm:
fsk_freq_est_errors
fsk_freq_est_ber

Next step is trying lower Rs, or higher Fs/Rs and see if it hangs on :+1:

drowe67:
@darksidelemm Better performance at low Rs:

fsk_freq_est_ber

I discovered we had disabled time-averaging of the DFT spectra, probably to get fast acquisition for digital voice applications.  Have now made this averaging an init-time parameter, that is set to "slow" for telemetry, and "fast" for the high bit rate modes.  This also improves the earlier "peak" algorithm.

However I think the combination of low Eb/No, low Rs, and high Doppler will be very hard to handle:
+ We currently look at a window of 50 symbols for timing est (2 seconds at Rs=25Hz)
+ To handle freq est at low Eb/No we need the DFT averaging above - which means a much wider window, perhaps 20 seconds.  We expect the freq offset to be constant in this window,
+ IIRC Peak Doppler is 100Hz/s, which is is 400% of Rs - every second.  This requires a very short time window for freq estimation.  It's much easier if Rs is higher, but then we can't get the low SNR performance we want.

So, while we can certainly develop a very low SNR telemetry mode for terrestrial use, I'm not sure how to make it work over satellite.  Some ideas:
+ The Doppler spends a lot of time pinned at +/- 10kHz, it's only the centre of the pass where it's changing fast.  However it's the centre where the SNR is maximum.
+ A clever freq tracking loop, or a coherent demod.  However this is very challenging and beyond the effort I'd like to put into this project.
+ Some sort of outboard freq correction for Doppler.  We could even try a variety of profiles, and see which one results in a decode. The Doppler is pretty well defined by physics, and CPU is cheap. IIRC that's how satellite based position location started.

darksidelemm:
I agree that there isn't much point trying to compensate for high speed doppler drift. This is a difficult problem by itself, and can be dealt with in another application.

I already have some code to apply doppler shifts on a sample-by-sample basis to an IQ file, it wouldn't be difficult to use the same code to do the reverse, and attempt to compensate for doppler from a balloon from a partially-known location. While it may not correct perfectly, it will get it close enough that the modem has less issues.

drowe67:
@darksidelemm If we assume (n the satellite use case) most of the frequency offset (and Doppler) will be corrected externally, we can consider a lower sample rate, and a lower Fs/Rs ratio.  At Fs=2000Hz, even the earlier "peak" algorithm works quite well at low Eb/No:

fsk_freq_est_ber

Note a complex bandwidth of 2000 Hz is about what a SSB receiver will delivery to the demod, so this is close to terrestrial balloon telemetry use case.

Thinking about a freq rate of change requirement to test against. Use cases:
1. Satellite, we assume most of the Doppler is removed, leaving a residual rate of change of TBC1 Hz/2
2. Balloon telemetry, temperature drift causing TBC2 Hz/s
3. Terrestrial HF/VHF.  Modern HF radios are stable to +/-1Hz, TBC3 Hz/s

billcowley:
Looks like you've made great progress on the frequency estimation.  Without reading all of the above, I agree that a mask (at the tone spacing) to select the actual frequency offset, should give better results than making decisions on individual tones.   In effect you are making 'soft' estimates via the mask correlation, which will reduce the final estimation error.  

A quick comment re the DFTs from line 202 in the estimation code: you are basically using the Welch method but with no overlap.  The results are slightly better with some overlap - 50% is usual.  One could also try modest zero padding - this might improve the resolution a little.  https://octave.sourceforge.io/signal/function/pwelch.html  https://en.wikipedia.org/wiki/Welch%27s_method

drowe67:
@darksidelemm @billcowley I've adjusted the FFT size to have bins 0.1Rs in width:

fsk_freq_est_ber

This has quite a lot of DFT averaging.  Haven't added Welch overlapping yet. Currently seeing what happens when I add a linear rate of change to the frequency.

drowe67:
Some curves (just "peak" method) that sweep the freq rate/change between 1% and +1% of Rs:

fsk_freq_est_ber_24000_25
fsk_freq_est_ber_8000_100
fsk_freq_est_ber_8000_10

As per the final graph, it gets more challenging as Fs/Rs increases, but we can compensate for that with lower Fs.  The top plot demonstrates we meet our requirements.

billcowley:
That looks good!  You could also add a demod setting for 'high Doppler/rapid channel variations' to reduce the frequency estimation duration ...

drowe67:
> That looks good! You could also add a demod setting for 'high Doppler/rapid channel variations' to reduce the frequency estimation duration ...

Thanks @billcowley - yes we have just such a mode, we call it "high bit rate mode", as it was initiailly designed for applications like digital voice that ran at higher bit rates that the telemetry.  It also uses smaller FFTs to fit on embedded platforms.

drowe67:
@darksidelemm I'm building up the full modem simulation.  

You recall we had some issues with BER counting at low Eb/No.  I've used the Binomial distribution to determine some sensible thresholds for the ```ber_counter()``` state machine.  Here is the PDF and CDF for the number of bit errors we are likley to get for a 512 bit frame, with a worst 30% BER (Eb/No 0dB ideal BER is 26%):
```
figure(3); clf; x=1:512; plotyy(x,binopdf(x,512,0.3), x, binocdf(x,512,0.3)); print("bino.png", "-dpng")
```
bino
Note that even with this super high BER, the range of likely error counts is rather narrow, due to the large frame.  If we have a good estimators and frame sync, then 50% of the frames tested will have < 153 errors (ie 0.3*512), so this is a good threshold for saying "we have frame sync". However its really unlikely to get more than 200 errors - we can say for sure the BER counter must be out of sync or the estimators have fallen over.

@billcowley I imagine this is not new to you!  But something I have only just got my head around.  Only took me 30 years of playing with modems :nerd_face:

drowe67:
OK so here's the results for the complete modem simulation:
fsk_modem_ber_24000_25
fsk_modem_per_24000_25
I've adopted a two state model for BER analysis in the raw, uncoded demod.  On some frames the estimators or BER counter completely lose it, we count them as a packet errors.  When we do have valid packets, we estimate BER.  Our operating point (coded PER=10%) is about Eb/No=1.5dB for 4FSK (coincidentally about the same as coded PSK Eb/No)  So we will lose about 20% of those packets through estimator issues, but out of what's left we can still FEC decode 90% of them.

I think that's acceptable for our use case.

billcowley:
Hi David,

The BER plot looks good.  If 0.25 Hz/sec corresponds to about 0.17 m/sec^2
- is that OK?  Guess it depends on how much the wind varies in speed and
direction.

I looked at some old Cisco projects we did, where the Doppler rate was over
6kHz / sec for S band ISLs.  Anyway, this might be useful re the FFT size/
averaging we discussed

*In passing we note a useful rule of thumb for frequency estimators which
must operate up to some maximum doppler rate. Assuming that we are
operating above the threshold SNR, the doppler rate times the observation
interval should be less than the FFT frequency resolution. With an FFT size
of N symbols, this gives*
[image: Screen Shot 2020-04-25 at 5.38.22 pm.png]

I've started a git repo for the LDPC codes - will let you know when it
contains something useful.

Cheers, Bill


*Bill Cowley*
*http://lowsnr.org <http://lowsnr.org>*

drowe67:
> Hi David, The BER plot looks good. If 0.25 Hz/sec corresponds to about 0.17 m/sec^2 - is that OK? Guess it depends on how much the wind varies in speed and direction.

Now that is an interesting question!  Something I had never considered before. Good catch Bill :+1: 

Thinking about sources of acceleration:
1. Periods of constant acceleration leading to large changes in velocity as a HAB hits different wind conditions.
1. Small random accelerations due to buffeting, swinging, rotation of the payload.
1. More violent accelerations of the payload during descent (fast rotation?).
 
Some of those might manifest as "noise" and (like Doppler spreading) rather than a linear change in frequency.

@darksidelemm - I'm wondering if some of the problems in telemetry during descent are due to Doppler shifts as the payload undergoes circular motion (i.e constant acceleration), rather than antenna pattern/fading/polarisation issues as we have suspected?

I wonder how this has affected 50 baud RTTY and 100 baud Horus binary?  Some demods (Fldigi) have wider filters which would help, however we have flight results that show Horus Binary works pretty well.

This one could be tricky - I currently use a lot of time averaging on the freq estimator to get good low Eb/No results.  Although that could be disabled when SNR is high (a likely case when the payload is in motion).  We could also raise the symbol rate.  

We probably need to set a requirement for this and test against it.

darksidelemm:
I have a very large set of 1-second sample rate radiosonde telemetry that can be used to estimate the worst-case acceleration around burst time. This probably won't show the small time-scale variations like David's points (2) and (3) however.

To measure these, I would suggest doing a small balloon launch with one payload transmitting a constant carrier, which is monitored throughout the entire flight. We would want to have receivers stationed at at least two locations representing:
- Listeners lower to the horizon (worst-case for x/y doppler)
- A listener (e,g, a chase team) directly below the balloon (worst-case for z-doppler).

Even if the doppler shift isn't an issue, it would still be an interesting dataset to gather, just to get an understanding of what happens to our transmitted signal during a flight.

billcowley:
With two (stationary) listeners orthogonal to the HAB (transmitting a CW signal or pseudo-random sequence),  plus careful time sync of the recordings, one could derive lots of info regarding the  HAB trajectory!   Hmm - that's quite a bit of effort.

drowe67:
@billcowley I've written some [Octave code](https://github.com/drowe67/codec2/blob/df414f72e4c27b64621281c3a0fe7a2fefdd4d48/octave/fsk_lock_down.m#L311) to look at the effect of different code rates.  

The major assumption is a packet durations of 20 seconds.

I guessed the raw BERs for each code based on similar codes we have played with in the past, e.g. we use a rate 0.5 code for FreeDV 700D (similar size) that works at about 8% raw BER.

Code Rate | Raw BER | 4FSK Eb/No | n,k | Rs | MDS |
1.00 | 0.002 | 8.0 | 256,256 |  6.4 | -156.9 |
0.80 | 0.040 | 4.5 | 320,256 |  8.0 | -159.5 |
0.50 | 0.080 | 3.5 | 512,256 | 12.8 | -158.4 |
0.33 | 0.160 | 1.5 | 768,256 | 19.2 | -158.7 |

I may have messed up the working, pls check if possible.  Getting some very low Rs values!  This will make frequency estimation very sensitive, as the bandwidth of each DFT-type filter will be very narrow.  One advantage of the lower code rates is Rs increases, making life easier for the frequency estimator.

billcowley:
David -   I will be able to look at these BERs etc tonight.  BTW I've started putting some codes in https://github.com/billcowley/Horus-FEC   but not much there yet.  I think I mentioned that these are fixed row-weight codes so the (C) encoder software will be the same as before. /Bill

billcowley:
David,     Note sure why the noise BW is fixed in your code - it should depend on the symbol rate?  Normally we would have SNR = Eb/No*Rs*k*m/B  where k is the code rate and m is the number of channel bits/symbol, B is the noise BW.  For PSK/QAM, Rs/B = 1, but for 4FSK it might be 1/4(?). So with m=2 for 4FSK, I guess this would give SNR = Eb/No *k/2

drowe67:
> David, Note sure why the noise BW is fixed in your code - it should depend on the symbol rate?

Yep - have seen a similar definition of SNR in the day job :smile: !

Guess I'm using the "Ham Radio" convention of SNR = (signal power)/(noise power in 3000 Hz), just to give us a comparison to other modes.  It's kind or arbitrary and I don't like it much or find it very useful.

I agree, Rs/B would be 1/4 for 4FSK, as you have 4 filters each of equivalent rectangular bandwidth Rs soaking up the noise.  Hmm, that's assuming the equiv rectangular noise bandwidth of each of those DFT filters is Rs.  I need to check that.

drowe67:
@billcowley I just tried https://github.com/billcowley/Horus-FEC and I think it worked OK! This figure came up:

fig1

@darksidelemm, pls give it a try:

1. You need the CML library installed somewhere on your machine, [instructions](https://github.com/drowe67/codec2/blob/master/octave/ldpc.m#L6).
1. Edit [this](https://github.com/billcowley/Horus-FEC/blob/master/test_ldpc4.m#L55) line to point it to your ```cml``` directory.
1. Run the simulation:
   ``` 
   $ octave --no-gui
   $ octave> more off
   $ octave> hfilename = 'H_128_384_23.mat'
   $ octave> test_ldpc4
   ```

billcowley:
@drowe67  : yes - that agrees with my plots :)   Increase the number of codewords to get smoother plots eg Ntrials = 20000
I'll put some rate 1/4 and rate 1/2 codes on the site soon.  I think your existing C encoder/decoder should still work - do you want me to play with any C software?

darksidelemm:
I'm planning on having a go at integrating the new codes into the C software sometime this weekend :-)

drowe67:
As above I was a little concerned with the sensitivity of the modem to phase/freq noise at low symbol rates (for example buffeting of the payload), so worked up a simulation.  I have a HF channel simulator that generates Doppler spreading samples, by filtering IQ noise samples at a certain "spreading" bandwidth.  So I just use the phase part of that (we don't want amplitude fading) to simulate "spreading" the energy of each tone:

fsk_modem_ber_spread_8000_25

I think this is reasonable, we wouldn't expect to see an impact on performance until significant amounts of energy are spread outside of the filter bandwidth, so we start to see an effect at 5Hz spreading with Rs=25.

billcowley:
Sounds about right David.   Why the outlier value of the red plot?    How many independent channel simulations were used for each Eb value?

drowe67:
Not sure about that outlier, that simulation doesn't save all the information.  Sorry i'm not clear on what an "independent channel simulation" is? The spreading sequence is generated anew for each Eb point, and there were about 25k bits per Eb point.

billcowley:
Ideally, generate a separate channel realisation each time you add a set of noise samples to a codeword.  If you only have one channel simulation for each Eb value, then you may have a particularly bad channel and that dominates the degradation.   I guess this happened for the red curve at 8dB? Give me a call on hangouts if you want to discuss further - cheers.

drowe67:
Here are some results for a larger rate/change in frequency offset (5% of Rs or 1.25Hz at Rs=25).  The goal is to make sure the demod doesn't break down completely.  While the implementation loss is pretty big, at Eb/No=6dB we start getting frames the FEC can handle (less than 15% raw BER). 

Given this wavefom is so sensitive, I imagine most of the time it will be operating with Eb/No >> the 1.5dB operating point.

fsk_modem_per_8000_25
fsk_modem_ber_8000_25

billcowley:
I've added a few more K=128 results to the git repo -- I'll do something similar for K=256.  With no error floor, perhaps you can avoid CRC checks?

billcowley:
Guys, just added some K=256 codes to my repo, plus summary results.

darksidelemm:
Thanks Bill!

I've added some information on the proposed packet structures here: https://github.com/projecthorus/horusbinary/wiki/6-Packet-Format-Upgrades

I think we can live with the CRC taking up 2 bytes, but I think its definitely worth testing the codes that don't have an error floor and seeing how they go.

drowe67:
I've made some progress on the C version of the modem :nerd_face: Here is is running at ```Eb/No=2dB SNR=2 + 10*log10(2*25/3000) = -15.8dB```.  Note the `cohpsk_ch``` channel simulator reports a 3dB lower SNR due to single sided/double sided noise thing. 

```
~/codec2/build_linux/src$ bits=512; tx_packets=20; rx_packets=18; tx_tone_sep=270; Rs=25; ./fsk_get_test_bits - $(($bits*$tx_packets)) $bits | ./fsk_mod 4 8000 $Rs 1000 $tx_tone_sep - - | ./cohpsk_ch - - -13 --Fs 8000 --ssbfilt 0 --complexout -f -3000  | ./fsk_demod -c -p 8 --mask $tx_tone_sep -t1 --nsym 100 4 8000 $Rs - - 2>stats.txt | ./fsk_put_test_bits -b 0.25 -p $rx_packets -f $bits -q -
SNR3k(dB): -18.75 C/No: 16.0
[0019] BER 0.163, bits tested   9728, bit errors 1584 PASS
```
The pass threshold is 90% packets through (10% PER).

@darksidelemm note it's running with a -ve freq offset (complex valued input), and the freq ests are quite smooth:
```
python ../../octave/plot_fsk_demod_stats.py stats.txt
```
c_mask_freq_est

darksidelemm:
Have updated fsk_demod_BER_test.py to work with the latest updates and calling conventions.

Here's the BER performance for 100, 50, and 25 baud, with a 48 kHz sample rate and a fairly broad estimator range (100-23000 Hz). The test signal was at +2 kHz, with a 270 Hz tone spacing. The mask estimator was used.

Screen Shot 2020-05-13 at 8 50 09 pm

darksidelemm:
Same for 2FSK (Same settings as the 4FSK run, just with 2FSK):
Screen Shot 2020-05-13 at 10 10 28 pm

darksidelemm:
Have added a Wenet sample and some ctests for it, with the new mask estimator enabled and disabled. Currently the test using the 'classic' frequency estimator fails. 

It may be frequency estimator related. Here is the estimator outputs for the mask estimator version:
Screen Shot 2020-05-14 at 9 01 03 pm

... and classic estimator:
Screen Shot 2020-05-14 at 9 00 43 pm

drowe67:
@darksidelemm nice work with https://github.com/drowe67/codec2/pull/101/commits/21796e79384959298d5c575bdfbdb327fab109c8 :+1: Very useful addition.

Tjoppen:
Here's an idea which I've brought up on IRC: since the preamble is known, we should be able to correlate it to the input signal to dig out packets which the current implementation is missing. It would entail mixing the input with a negative frequency version of the preamble. That way any packets using that preamble should show up as some nice peaks. As a bonus any CW interferers get smushed out. The more bits of the packets which are known, the better this method would work.

drowe67:
@Tjoppen thank you yes I use this method for acquisition on the OFDM modem. Currently we plan to use binary correlation of the Unique Word - our calculations indicate this should find most of the packets.

Tjoppen:
@drowe67 What do you think about using soft bits in the UW detection? I had some issues with the binary UW implementation in master. But that might of course be due to the worse frequency in the modem in master :)

drowe67:
@Tjoppen yes Mark (@darksidelemm ) and I have discussed various ways of UW sync incudling correlation/soft decision. In fact, he is building up that part of the new waveform (UW frame sync) right now so we shall soon see if it's robust enough :grin: 

I've started using the [Binomial distribution](https://github.com/drowe67/codec2/pull/101#issuecomment-618684911) to set UW thresholds.  Finally feels like I'm "engineering" it...

What application have you been using the modem for?  Yes freq estimation at low SNRs is much improved now.

Tjoppen:
@drowe67 Cool, my thoughts are along the statistics lines as well.

At the moment I'm mostly experimenting, but the general idea is QRPp packet radio and/or some kind of APRS replacement. For QRPp my goal is to get something out of my HackRF that can be decoded from 80 km away. I know WSPR + HackRF TX works on 80m as far as France, from northern Sweden. Output power is around 10-15 dBm.

JeroenVreeken:
If you want an APRS replacement you might want to have a look at FPRS.
(http://dmlinking.net/fprs.html) It is a binary replacement for APRS.

drowe67:
> If you want an APRS replacement you might want to have a look at FPRS.
> (http://dmlinking.net/fprs.html) It is a binary replacement for APRS.

@JeroenVreeken so I'm looking into your test and example for ```freedv_data_channel.c``` (it's been a while so I'm coming up to speed).  It's a very powerful protocol :+1: 

Couple of questions:
1. How would this integrate with existing Ham data applications, e.g. [Pat](https://getpat.io/) Or [BPQ32](http://www.cantab.net/users/john.wiseman/Documents/BPQ32.html)?  or do you see it as a replacement/alternative for existing protocols?
1. Is it a problem that the modems it runs on don't have any FEC?
1. Some of the groups I am talking too (e.g. Winlink) already have there own protocol, e.g. ARDOP for HF, they just want to have an API the send receives/blocks of data.  For that should we use the low level data functions you have written in ```freedv_api.c```?

drowe67:
@Tjoppen your experiments are similar to what @darksidelemm and myself and planning.  On a good LOS path 80km should be no problem on 10mW.

Tjoppen:
@JeroenVreeken I'm concerned with the network/physical layer, not the application layer. DML looks interesting however.

@drowe67 Yeah, depending on baud rate of course. The ideas we've been floating locally involve sensor networks transmitting as slow as 1 baud, and AMPRnet. I'm leaning toward getting IPv4 and IPv6 working, with a very wide range of baud rates. Like 1 baud up to 1 Mbaud. I've had promising results cobbling together TUN/TAP with the FSK modem on Linux. Basically turns your radio/SDR into a network card. My rough goal is getting ssh and mosh working over it, so that it can be used to control machines like repeaters over the air.

Winlink unfortunately seems to only be for e-mail, and ARDOP is awfully US centric and steeped in legacy stuff. HFIP also exists, but I wasn't particularly impressed with that either.

JeroenVreeken:
I have been doing a similar thing with the freedv data channel as well.
My eth_ar project (http://dmlinking.net/~pe1rxq/eth_ar.git/) contains a program called 'freedv_eth'. Among other things it creates a tap device and any ethernet packets that do not contain voice or command data go straight to the data channel. 
(It can do a bit more, including controlling a repeater)
Basicly any protocol that can use ethernet as a lower layer works on any freedv mode that has a data channel

JeroenVreeken:
@drowe67 Having FEC is not necessary as there is still a CRC per packet, but it will help ofcourse.
The data channel code just provides a layer that looks like ethernet, so PAT and likes should be able to run on top of it without much effort. If they need some kind of acknowledge packets they can build on top of it with just a few bytes extra.

drowe67:
@Tjoppen @JeroenVreeken The TAP/TUN sounds very cool.  I'm also very interested in a "turn key" IP link using these modems on VHF/UHF.  This has applications for Ham data, emcon and developing world "TV white space" connectivity.  At the low end this could be RpiTx (with LPF) and a RTL-SDR, and some copper wire soldered to a Pi for an antenna.  After that, some sort of SDR with Tx capability.

@JeroenVreeken I am intrigued by the possibilities of your data channel protocol but also feel we need a low level interface into the FreeDV modems so people can use their existing protocols.  will work on that in another PR. It appears there aren't many groups developing open source HF/VHF modems so access to our modems might be the "sweet spot" here - i.e. the largest benefit we can offer.

Tjoppen:
> @Tjoppen @JeroenVreeken The TAP/TUN sounds very cool. I'm also very interested in a "turn key" IP link using these modems on VHF/UHF. This has applications for Ham data, emcon and developing world "TV white space" connectivity. At the low end this could be RpiTx (with LPF) and a RTL-SDR, and some copper wire soldered to a Pi for an antenna. After that, some sort of SDR with Tx capability.

This is very much in line with my thoughts. Having some decent way of dealing with different RX and TX capabilities would be the trickiest issue I think. That and channelization. For example, with an RTL-SDR in direct sampling mode the RX side will cover 20m down to 2200m. Gnuradio could be used for channelization into multiple named pipes to which the same TUN/TAP device is connected. Most of the pieces exist, they just need gluing together.

One major hurdle I've encountered so far is latency. In my experimenting with my FT-817 the latency from pinging the TUN device to the PTT activating was close to a second, no doubt due to lots of buffering.

drowe67:
@Tjoppen @JeroenVreeken sounds good.  I'd be interested in working with you on such a project.  Probably best we continue the discussion outside of this PR though :smile: Pls feel free to email me directly or we could use the codec 2 mailing list (or anywhere else you'd like).

Tjoppen:
@drowe67 Email sent

drowe67:
@darksidelemm @billcowley This PR is getting pretty big and I want to sync up with the rest of codec2, which has also had a lot of refactoring. So I'll merge this PR and kick off a new PR to continue this fine project :1st_place_medal:

Issue #102, opened 2020-04-28
problem with ofdm.c


dave-g8kbb:
Apologies if this is the wrong place to raise this.
I think there may be a problem with ofdm.c. James Ahlstom's quisk opens two receiver channels by calling freedv_open() twice. Quisk currently uses version 12, and this works fine. In version 13 however, ofdm.c introduced some static variables, for example tx_uw_syms. When quisk closes the first receiver channel, ofdm_destroy() calls FREE() to free up the memory allocated. When the second channel is closed, the call to ofdm_destroy() causes a memory exception which crashes quisk. Quisk is a python application that calls codec2 as a DLL via a python extension dll.
Should these variables not be part of the OFDM structure?
Dave

drowe67:
@dave-g8kbb thanks for your report.  Yes I can see how problem might crop up if there are two instances created of the modem, something we've not encountered until now.

@srsampson what do you think?

I'm happy to review a PR on fixing this one if either of you want to work on it.

dave-g8kbb:
Hi @drowe67 I'm happy to help in any way I can. I'm no expert on the internals of codec2 though. Also first time I've raised a query like this. What would you like me to do?
Best regards
Dave

srsampson:
The theory is, that with the SM1000 we don't want to allocate the filter memory as it is too large. So we moved the allocate out of ofdm_create(). The reference in ofdm_destroy() is an error then. freedv_api should handle the calls. I removed it from ofdm_destroy().

drowe67:
@srsampson is working on this on in https://github.com/drowe67/codec2/pull/103, will continue conversation there.

srsampson:
I created a new pull and just moved the unique word allocations as discussed.

Not sure if that resolves anything, but it does pass compile and tests.

drowe67:
Nice work @srsampson :+1: 

@dave-g8kbb would you be able to test Steve's branch in your application please?

dave-g8kbb:
@drowe67 I downloaded the branch, and it rebuilt without errors. I patched into quisk and it no longer crashes when switching modes.
@srsampson many thanks for the prompt work.
I'll carry on today retesting the complete system including some on-air tests - will let you know how it goes.

Best regards and once again many thanks.
Dave

srsampson:
Thanks Dave, hope all works fine!

Steve

dave-g8kbb:
[...]

PR #103, opened 2020-04-28
Remove tx filter deallocate from ofdm_destroy

srsampson:
Fix for https://github.com/drowe67/codec2/issues/102, to support multiple parallel instances.

### DIFF
--- a/src/codec2_ofdm.h
+++ b/src/codec2_ofdm.h
@@ -91,7 +91,7 @@ void ofdm_set_phase_est_enable(struct OFDM *, bool);
 void ofdm_set_phase_est_bandwidth_mode(struct OFDM *ofdm, int val);
 void ofdm_set_off_est_hz(struct OFDM *, float);
 void ofdm_set_sync(struct OFDM *, int);
-void ofdm_set_tx_bpf(struct OFDM *, bool);
+void ofdm_set_tx_bpf(bool);
 void ofdm_set_dpsk(struct OFDM *ofdm, bool val);
     
 void ofdm_print_info(struct OFDM *);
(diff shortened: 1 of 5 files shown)


srsampson:
I knew that was going to fail, but I'm trying to think of a work-around.

drowe67:
@srsampson I think the filter states are fine inside ```struct OFDM```, they don't get allocated when used on the SM1000.

The malloc/free code for the filter is tricky, but working correctly I think.  It could perhaps use some refactoring.

I think the issue @dave-g8kbb suggested is these:

```
static complex float *tx_uw_syms;
static int *uw_ind;
static int *uw_ind_sym;
```

get malloced/freed.  So on applications that create two instances of the OFDM modem, they dissapear when one instance is closed.

So maybe we need to move them inside ```struct OFDM``.

srsampson:
@drowe67 I wonder if all the statics can be moved into the structure?

That way the ofdm pointer returned from one create has no affect on other creates.

drowe67:
> @drowe67 I wonder if all the statics can be moved into the structure?
> 
> That way the ofdm pointer returned from one create has no affect on other creates.

@srsampson yes that would be ideal, I recall we were heading in that direction in stages.  IIRC the some of the static variable are used as "constants" that replaced #defines in earlier versions.

Suggest as a first step, just move any dynamically allocated variables into the struct.  If two instances of the same modem parameters are creates, that should allow multi-instance operation.

drowe67:
@srsampson another reason for the filter states being inside the struct is to support multiple instances, as each filter will have different data flowing through it.  So with regards to the filter,  I think what we have in ```master``` is fine.

PR #104, opened 2020-04-29
Move allocation pointers to ofdm struct

srsampson:
Resolve problems when multiple instances are created


srsampson:
Looks like user report is positive. Request merge. Thanks!

PR #110, opened 2020-05-14
QAM and Packet Data for OFDM

drowe67:
Getting interested in open source data modes, so lets build up the Octave support for:

1. QAM16
1. Data packets that are multiples of modem frames.

# QAM16

Initial Octave uncoded modem working:
```
ofdm_tx("test.raw", "QAM16", 10,10)
ofdm_rx("test.raw", "QAM16")
Rs=50.00 Nc=33 Tcp=0.004 Nbitsperframe: 528 Nrowsperframe: 4 Ntxtbits: 4 Nuwbits: 20 bits/s: 5280.0
BER2.: 0.0017 Tbits: 40656 Terrs:    71
```
This waveform as pilot rate of 8000/50/5 = 10Hz, so can track +/-5Hz Doppler spread (much better than 700D/2020).

I've built up a LDPC code simulation for QAM64:
```
octave:9> qam16=1; ldpcut
```
qam64_scatter
qam64_504_396
Es/No is easier for me to use for these higher order constellations.  I think this means a PER < 10% at:
```
 SNR = Es/No + 10*log10(Rstotal/B) = 12.5 + 10*log10(50*32/3000) = 12.5 - 2.5 = 10dB
```
... in a 3000 Hz noise bandwidth. The symbol rate Rstotal is the total number of symbols per second, given by the symbol rate per carrier times number of carriers.  SNR=10dB is pretty good - e.g. analog SSB isn't much fun at 10dB.  We can expect to lose a few dB from synchronisation overheads and implementation loss, plus 4-6dB dB for a HF channel.  This suggests a real world SNR of 16dB over a HF channel, for a payload data rate of 5280*396/504 = 4148.6 bits/s.

# Data Packets

First pass uncoded modem running:
```
octave:31> ofdm_tx("test.raw", "data",1);
Rs=50.00 Nc=17 Tcp=0.004 5 Ns: 2 Np: Nsymperframe: 85 Nbitsperpacket: 272 Nsamperframe: 960 Ntxtbits: 4 Nuwbits: 10 bits/s: 1360.0
PPPPPPPPPPPPPPPPPPP
 .........U....... 
 .U........U...... 
 ..U........U..... 
 ................. 
PPPPPPPPPPPPPPPPPPP
 ................. 
 ................. 
 ................. 
 ................. 
octave:32> ofdm_rx("test.raw", "data");
<snip>BER..: 0.0000 Tbits:   816 Terrs:     0
```

# TODO

(list of ideas, don't all have to be done in this PR)
1. Review simulation, make sure Es=1, this is a bit tricky for QAM.
1. Try APSK16.
1. Develop and try a longer code (e.g. (4 sec)(32 carrier)*(50 symb/s)*(2 bits/symbol)=12800 bit code, could be worth 2dB.
1. Make sure QAM works with changing amplitude, i.e. pilots recover amplitudes.
1. Use QPSK symbols for UW demapping to make UW detection more robust.
1. ~~Octave unit test for QAM64 constellation~~ ```ldpcut.m```
1. Integrate QAM64 and LDPC in Octave modem simulation
1. C port
1. ~~Design a better waveform with increased PS rate, CP.~~
1. Ways to achieve faster sync for data applications to avoid ARQ


drowe67:
@srsampson check this out :grinning: !  It wasn't as easy as I thought, but every time we try something new it means refactoring the modem so it's more flexible.  This time it was in the UW mapping area.

srsampson:
Wow, you are on top of it! I was reading another PDF about constellation patterns and a Chinese engineer came up with this one for over fiber. It's called a honeycomb constellation. Performance was greater all around.
```
Honeycomb Constellation

 0.7071 ( ring 1)
 1.3660 ( ring 2)
 1.7071 ( ring 3)
 2.3660 ( ring 4)

theta(i) = (pi / 2)*i = 0, 1, 2, 3 ( ring 1,3)
         = (pi / 2)*i + (pi / 4), i = 0, 1, 2, 3 ( ring 2, 4)
```
honeycomb

drowe67:
@srsampson OK that's interesting.  Perhaps we could test different constellations by measuring BER over of channel.

drowe67:
@billcowley Here is a WIP PR for QAM

octave/ldpcut.m
```diff
@@ -214,9 +214,20 @@
   mod_order = 16; bps = log2(mod_order);
   modulation = 'QAM'; mapping = ""; demod_type = 0; decoder_type = 0;
   max_iterations = 100; EsNo_dec = 10;
-
+  qam16 = [
+    1 + j,  1 + j*3,  3 + j,  3 + j*3;
+    1 - j,  1 - j*3,  3 - j,  3 - j*3;
+   -1 + j, -1 + j*3, -3 + j, -3 + j*3;
+   -1 - j, -1 - j*3, -3 - j, -3 - j*3];
+  qam16 = qam16/std(qam16(:));
```
billcowley, line 222: std divides the sum of squares by N-1; so I think v=qam16(:);  qam16 = qam16/ (sqrt(v'*v/16)) should be slightly more accurate

billcowley:
The APSK family are now often used instead of rectangular QAM eg see https://en.wikipedia.org/wiki/Amplitude_and_phase-shift_keying  especially for nonlinear channels.    Octave doesn't seem to include these yet (although matlab does).   The CCSDS 131.2 Blue book defines them  eg https://public.ccsds.org/Pubs/131x2b1e1.pdf - see section 5.2.3   The ring ratios seem to depend on the coding rate (see table 5-1) which is a shame.  I expect the performance differences are small, so maybe just use 3 as the ring ratio (for 16 APSK) between the 12 point ring and the 4 point ring.  I agree @drowe67 that would be useful to test via BER simulations.  Using APSK instead of rectangular may not matter much in multicarrier/OFDM cases.

drowe67:
Thanks @billcowley - I will try out the new constellations when I make my next pass at this code :smile:

billcowley:
Hi David,

re theoretical BERs for QAM and APSK:  there's not a lot around.   You can
try some of my old code below for rect 16QAM.  Note that even this code is
an approximation - it should be good for medium to high SNRs.  There are
quite a few BER and SER plots around for APSK, mainly from simulations.
The easiest approach might be to fit some polynomial models to APSK
simulation results and create a function to evaluate BER or SER at a given
Eb value. That should be fairly easy - would you like me to do that?
Cheers, Bill

---------------
function p = ber_qam(ebn0)
% Calculate the bit error rate (BER) for square 16QAM in AWGN
%   given the Eb/N0 in dB,  ebn0 can be a scalar or vector
%   (assuming coherent detection, uncoded)
%   [section 5.3 Webb and Hanzo text]

e = 4*10.^(ebn0/10);    %  Es/N0 vector in linear
b2 = qfn(sqrt(e/5));
b1 = (qfn(sqrt(e/5)) + qfn(3*sqrt(e/5)))/2;
p = (b1+b2)/2;
-----------------------------

function tail=qfn(a)
% Usage: tail=qfn(a)
% where: tail=area under the tail of the normal dist. from a to inf.
%        for zero mean, unit variance distribution
%
% If no argument is given, plot Q(x) for x = 0 to 5

% use erfc instead of 1-erf to avoid truncation errors!     April 2010

fact = 1 / sqrt(2);
if exist('a')

    % tail = 0.5 * ( 1 - erf(a * fact));
    tail = 0.5 * erfc(a * fact);
else
    x=(0: 0.1: 6); semilogy(x,  0.5*( erfc(x * fact)));
    title('Q function plot');
    xlabel('x'); ylabel('Q(x)');
end
--------------------

drowe67:
@billcowley Thanks I'll try the earlier code you have posted above first, lets see how that goes :+1:

octave/ofdm_lib.m
```diff
@@ -10,8 +10,235 @@
...
+    DDDDDD    |     |      |
+    DDDDDD    Ns    |      |
+    DDDDDD    |     |      |
+   PPPPPPPP  ---------     |
+
+   Freq------------------>
+
+   Time flows down, freq across
+#}
+
```
drowe67, line 40: @srsampson ... first steps in making this modem run with multiple pilots/modem frame.  Will need to move very carefully when I start tweaking the phase estimation :grimacing: 

This will let us have much longer modem frames (seconds on HF), and also gets us close to a single carrier modem for VHF and above, where we might have a modem frame of 100 symbols with pilots 4 symbols apart.

drowe67:
@srsampson Doh I got part way through the Nf/Ns mods and realised I have this all wrong.  Should be:
1. modem layer outputs a stream of bits, and operates same as it does now, on blocks of Ns symbols "PDDD".
1. Next layer is a framer that does UW sync and FEC, much like the VHF modes.
1. So to handle much longer packets of data, I need to abstract out the frame assembly, state machine, and FEC.  Good news is this is much easier than the path I was currently on (multiple pilots per modem frame).  Time to roll back the recent commits.  But I might keep the ascii art, quite like that :smile:

drowe67:
@srsampson - I've made some progress in have modem "packets" longer that the "modem frame".  Check out the "Data Packet" heading at the top of this PR, it even draws an ascii-art picture of the packet format.

This is also one step closer to single carrier operation, but that also requires some new code in acquisition.

srsampson:
Your art is simpler than mine, just for historical (hysterical) purposes:
```
S = Sync Word (Spread Out on Different Carriers) No FEC
T = Data bits (4 varicode bits, 2 symbols) No FEC
The rest of the QPSK symbols are interleaved LDPC (FEC) and Vocoder bit-pairs

|<---------------------------------------- BPSK Pilots --------------------------------------->|
+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
| 00 | 01 | 02 | 03 | 04 | 05 | 06 | 07 | 08 | 09 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 |
+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+

     |<-------------------------------- QPSK Subcarriers -------------------------------->|
     +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
     |    |    |    |    |    |    |    |    |    | S  |    |    |    |    |    |    |    |
     +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
     |    | S  |    |    |    |    |    |    |    |    | S  |    |    |    |    |    |    |
     +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
     |    |    | S  |    |    |    |    |    |    |    |    | S  |    |    |    |    |    |
     +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
     |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |
     +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
     |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |
     +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
     |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |
     +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
     |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    | T  | T  |
     +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
```

drowe67:
@srsampson that looks really good! Is it automagically generated?

srsampson:
> @srsampson that looks really good! Is it automagically generated?

No, which makes yours many times more valuable.

srsampson:
A data mode would be very popular. If the constellation is wide enough it could even fill the FM 5 kHz channel. Although multi-mode (SSB) rigs are pretty common now on 2 metres, I believe they would be limited to less than 3 kHz BW. In any case, the spectrum would not be anywhere near DC like the old 9600 modems, thus the digital input could be used to a greater effect.

drowe67:
@srsampson @billcowley I'm starting to get some good results.  The "data" waveform has a payload data rate (throughput) of roughly 1000 bits/s or 7500 bytes/minute (the standard Winlink unit).  The Codec 2 modem doesn't have ARQ yet.  At 10% PER I assume ARQ + other overheads results in a 20% hit on bandwidth, so I'm modelling the codec2 throughput as 0.8*7500 = 6000 bytes/min.

| Channel | Doppler (Hz) | Delay Spread (ms) | SNR3k  | Codec2 | ARDOP | VARA |
| --- | --- | --- | --- | --- | --- | --- |
| AWGN | - | - | 3 | 6000 | 2000 | 6000 |
| MPP | 1 | 2 | 12 | 6000 | unusable | 2500 |
| MPD | 2.5 | 4 (see note) | 13 | 6800 | n/a | n/a |

The last three columns are the bytes/minute compared for three modems.

Still very early days for HF data, some notes
1. Waveform details in the [spreadsheet](https://github.com/drowe67/codec2/blob/dr-qam/doc/modem_codec_frame_design.ods), I used 100 packets for these results.
1. The channel names, ARDOP and Vara results from [this report](https://www.winlink.org/sites/default/files/downloads/a_winlink_digital_mode_performance_comparison_based_on_the_ionos_hf_vhf_channel_simulator_-_may_18_2020.pdf). Although I cheated on the MPD channel, using 4ms instead of 5, which is breaking the modem for some reason I haven't looked into yet.
1. Need to check I haven't made an error and am getting results "too good to be true".
1. Run some of these files over the air, and see if I can actually send packets over real HF channels.
1. Need a waveform that works in the 5dB SNR range, as that's more typical of interstate paths here.
1. @billcowley I'm currently using the "longest" code I have - the H2064_516 code you developed for the Wenet SSTV system.  Despite the high code rate, it actually does OK, although the packet length is a bit short to ride over the fades (just less than twice the MPP Doppler spread).
1. @srsampson would you like to work with me to port some of this work into the C version of the modem?

Typical command line:
```
octave:114> ofdm_ldpc_tx('test.raw',"data",60,12,'mpp')
octave:114> ofdm_ldpc_rx('test.raw',"data")

**Comment by srsampson** [comment]
Is "make test" working for you? Mine hangs at test 6 with 100% CPU on 3 out of 4 cores. Using the dr-qam branch. I'm currently running Ubuntu 20.04 Octave 5.2.0

**Comment by drowe67** [comment]
> Is "make test" working for you? Mine hangs at test 6 with 100% CPU on 3 out of 4 cores. Using the dr-qam branch. I'm currently running Ubuntu 20.04 Octave 5.2.0

Yes 66/66, and Travis seems to like it too :point_up: . But you may have hit a corner case.  I haven't added any tests recently, just made sure the existing ones keep working.  It's kind of nice how the ctests prod you in that direction.

**Comment by srsampson** [comment]
> > Is "make test" working for you? Mine hangs at test 6 with 100% CPU on 3 out of 4 cores. Using the dr-qam branch. I'm currently running Ubuntu 20.04 Octave 5.2.0
> 
> Yes 66/66, and Travis seems to like it too  . But you may have hit a corner case. I haven't added any tests recently, just made sure the existing ones keep working. It's kind of nice how the ctests prod you in that direction.

I can make it to:
```
NC=26 octave-cli --no-gui -qf tofdm.m
```
After that it hangs. Oh well, testing is for later, I'll look at your octave code. I notice it just uses QPSK, is that correct? I think I saw 2 bits in the modem spreadsheet.

**Comment by srsampson** [comment]
Just a quick debug, it seems to blow up at line 90 in tofdm.m
```
  tx_log = [tx_log ofdm_mod(states, tx_bits)];
```
So I guess ofdm_lib.m must be having trouble with Ubuntu 20.04 with large complex arrays or something.

drowe67:
> Just a quick debug, it seems to blow up at line 90 in tofdm.m
> 
> ```
>   tx_log = [tx_log ofdm_mod(states, tx_bits)];
> ```
> 
> So I guess ofdm_lib.m must be having trouble with Ubuntu 20.04 with large complex arrays or something.

tofdm.m ilne 90 is comment in this branch Steve, or did you mean around line 90?

srsampson:
> > Just a quick debug, it seems to blow up at line 90 in tofdm.m
> > ```
> >   tx_log = [tx_log ofdm_mod(states, tx_bits)];
> > ```
> > 
> > 
> > So I guess ofdm_lib.m must be having trouble with Ubuntu 20.04 with large complex arrays or something.
> 
> tofdm.m ilne 90 is comment in this branch Steve, or did you mean around line 90?

Aha, yes, line 95 actually. I need my glasses! I looked at ofdm_lib and didn't see anything obvious. The fact that it runs on previous Ubuntu and Travis means it is my current versions being different.

I may built an older Ubuntu on my old machine, or maybe on the Raspberry.

drowe67:
@billcowley well your functions for QAM16 BER worked pretty well :grinning: !
qam16_uncoded_ber
Really useful, as it gives us some confidence that we are on the right track.  I also used your normalisation term above for the constellation power - now the average symbol power is right on 1 - thanks :+1: 

Still to try APSK, and modify the OFDM modem library to perform amplitude estimation so we can use QAM for real.

drowe67:
@srsampson @billcowley I'll merge this PR now, so we don't get too far out of sync with master and to make Steve's C port easier.  Will then recreate a new PR when next I work on this.  

Working on amplitude estimation and correction for QAM next.  We already have the pilots for phase estimation.

PR #114, opened 2020-05-21
Removing statics from ofdm modem

drowe67:
1. Would like to remove the statics from the ofdm modem that are used as constants that define the "config" of the modem (symbol rate, number of pilots etc).  
1. These constants don't change after ```ofdm_create()``` is called, they were #defines in the first version of the modem.
1. Move them into the ofdm modem states so we can create instances of the modem with different configurations.
 1. Have also simplified the call to ```ofdm_create()``` in the default case, you can just pass in a NULL ptr:
   ```
   ofdm = ofdm_create(NULL);
   ```


drowe67:
@srsampson I've just made a few first steps but ```ctest -R test-OFDM``` passes.  Pls tell me what you think :thinking: of this approach.  Happy to discuss other ways of doing this.

src/ofdm.c
```diff
@@ -84,11 +84,11 @@ static const int8_t pilotvalues[] = {
    1, 1, 1, 1,-1, 1,-1, 1
 };
 
-/* static variables - used as storage for constants that are unchanged after init-time */
+/* constants that define modem waveform, unchanged after init-time,
+   but may vary for each instance of the modem */
 
-static struct OFDM_CONFIG ofdm_config;
+#define ofdm_tx_centre (ofdm->config.tx_centre)
```
drowe67, line 91: So the trick is to make each of the current statics a macro like this, so the code reads the same.  I've just changed one as a starting point (feel free to change them all if you like this approach).

src/ofdm.c
```diff
@@ -167,13 +167,10 @@ struct OFDM *ofdm_create(const struct OFDM_CONFIG *config) {
...
-    /* Check if called correctly */
+    ofdm = (struct OFDM *) MALLOC(sizeof (struct OFDM));
+    assert(ofdm != NULL);
 
     if (config == NULL) {
-        return NULL;
-    }
-
-    if (config->nc == 0) {
         /* Fill in default values */
```
drowe67, line 174: This is the other change to the modem API - I felt this was more intuitive, and it simplifies the ```ofdm_create()``` calls

src/ofdm.c
```diff
@@ -818,29 +811,12 @@ void ofdm_txframe(struct OFDM *ofdm, complex float *tx, complex float *tx_sym_li
...
 
-struct OFDM_CONFIG *ofdm_get_config_param() {
-    return &ofdm_config;
-}
-
-int ofdm_get_nin(struct OFDM *ofdm) {
-    return ofdm->nin;
-}
-
-int ofdm_get_samples_per_frame() {
```
drowe67, line 829: The next few getters have to be changed when the stats are finally rm-ed.

src/ofdm.c
srsampson: I'm wondering if it creates confusion though. Maybe just change all the variables to the new structure reference style? The only advantage would be keeping the old style.

I can do those changes pretty simple in the editor.

drowe67: Yep, or make them upper case ```#define OFDM_TX_CENTRE``` so we know they are macros.  The idea was to have something that was easy to read for the constants, rather that two deference levels.

However if you're comfy with just changing them all to the form ```ofdm->config.foo_bar``` I'm OK with that :+1:

srsampson: I'll see if I can get it done this evening.

src/ofdm_internal.h
```diff
@@ -82,19 +82,55 @@ struct OFDM_CONFIG {
...
     int bps;   /* Bits per Symbol */
     int txtbits; /* number of auxiliary data bits */
     int ftwindowwidth;
+
+    int samples_per_frame;
 };
 
 struct OFDM {
+    struct OFDM_CONFIG config;
+
```
drowe67, line 98: I was wondering if we still need ```config``` embedded in the ofdm struct but I think we do - so we can return a ptr to it if someone wants to take a peek at the current config.  An alternative would be a getter that copies the config into a user supplier config struct.

drowe67:
@srsampson :+1: I think what you have done is better than my original approach :thinking:

src/ofdm_internal.h
srsampson: Hmm... A getter that copied the values into a user supplied struct might save memory in the Embedded version.

drowe67: Yes and maybe more intuitive than https://github.com/drowe67/codec2/blob/dr-ofdm-rm-statics/src/freedv_2020.c#L45

drowe67: But it's a fine point....what we have is pretty good :smile: I'm impressed with how easy this modem is to modify (e.g. this PR and the QAM work).  Shows we have a good design :+1:

drowe67:
@srsampson are you happy with this PR?  If so I'll merge shortly (after sorting out the conflict)

srsampson:
> @srsampson are you happy with this PR? If so I'll merge shortly (after sorting out the conflict)

Yes, I think it's ready for prime time!

PR #116, opened 2020-05-22
Data demos

drowe67:
1. Break out VHF packet data demos to separate programs from ```freedv_rx``` and ```freedv_tx``` to simply these examples.
1. Add raw data mode/demo to allow developers to use the codec2 modems with their own protocols (typically HF).
1. Made another pass at refactoring the freedv_api eco-system to push bit packing out to edges with HF modes.  VHF modes can still used packed frames internally.

@JeroenVreeken:

1. I've been studying your README_data.md :+1: , and made a few minor tweaks.
1. I've started using the nomenclature "VHF packet" to describe your fine packet system, to make it easy to distinguish from other data protocols that might use the codec2 modems, e.g. those from 3rd party developers.
1. Can you please take a look at ```freedv_mixed_tx.c``` and ```freedv_mixed_rx.c```?  Perhaps strip them down to the bare minimum to demonstrate mixed speech/VHF packet operation.  For example you only need the FSK modes, and a bunch of other options/BER testing aren't required, maybe add a ctest for them.
1. Feel free to push directly to this PR :smile:


README_data.md
```diff
@@ -123,6 +123,19 @@ When it is possible to determine activity in the voice signal (and it almost alw
...
+One simply uses the regular API calls for reception of speech samples. In addition the callback functions are used for data.
+There is one caveat though: when a data frame is received the API functions (like freedv_rx) will return zero as this is the amount of codec/voice data received.
+For proper playback silence (or comfort noise) should be inserted for the duration of a frame to restore the timing of the original source speech samples.
+An example of how this is done is provided in freedv_mixed_rx
+
+  ```
+   $ ./src/freedv_mixed_tx 2400A ../raw/hts1a.raw - | src/freedv_mixed_rx 2400A - ./hts1a_out.raw
+  ```
+
+
```
drowe67, line 138: @JeroenVreeken that's great!  Really simple demo :+1:

src/freedv_api.c
```diff
@@ -671,9 +671,11 @@ int freedv_bits_to_speech(struct freedv *f, short speech_out[], short demod_in[]
...
                     speech_out[i] = passthrough_gain*tmp[i];
             } else {
-                nout = f->nin_prev;                    
+	        /* Speech and modem rates might be different */
+	        int rate_factor = f->modem_sample_rate / f-> speech_sample_rate;
+                nout = f->nin_prev / rate_factor;
                 for(int i=0; i<nout; i++)
-                    speech_out[i] = passthrough_gain*demod_in[i];
+                    speech_out[i] = passthrough_gain*demod_in[i * rate_factor];
            }
```
drowe67, line 671: @JeroenVreeken Ok that's an interesting idea, it's simple decimator.  The code above for 2020 uses a decimating filter to knock out the high frequencies before we drop the sample rate.  So this code will run for 2400A and B.  :thinking:

JeroenVreeken:
Should we also update the library version to 0.10?
Since the codecrx and codectx functions were renamed the API is not compatible anymore.
As an alternative we could provide a wrapper for both functions so code using the old names keeps working.

unittest/tfreedv_mode_800XA.c
```diff
@@ -0,0 +1,133 @@
...
+    printf("%d Passed\n", codec_bits);
+
+    printf("freedv_get_n_bits_per_modem_frame() ");
+    int frame_bits = freedv_get_bits_per_modem_frame(f);
+    if (frame_bits != 56) {
+        printf("Expected 56 codec bits, got %d\n", frame_bits);
+	goto fail;
+    }
+    printf("%d Passed\n", frame_bits);
+
```
drowe67, line 93: @JeroenVreeken these tests are testing constants, is there any conceivable way they could fail? I am wondering if they do anything useful. Also you could just do them in one line with asserts:
```
assert(freedv_get_bits_per_modem_frame(f) == 56);
```

unittest/tfreedv_mode_800XA.c
```diff
@@ -0,0 +1,133 @@
+/*---------------------------------------------------------------------------*\
+
+  FILE........: tfreedv_mode_800XA.c
+  AUTHOR......: Jeroen Vreeken
+  DATE CREATED: 24 May 2020
+
+  Tests specific for mode 800XA
+
```
drowe67, line 8: Perhaps call this tfreedv_800XA_rawdata.c of similar? It's not actually testing the 800XA (voice) mode, just the rawdata API of this mode.

unittest/tfreedv_mode_800XA.c
```diff
@@ -0,0 +1,133 @@
...
+    printf("freedv_rawdatatx()/freedv_rawdatarx() ");
+    int frames = 0;
+    int fails = 0;
+    {
+        short mod[nom_samples * 10];
+	/* Note: A codec frame is only 3.5 bytes!
+	   so th fourth and eight bytes will be half empty!
+	 */
+        unsigned char payload[8] = { 0x11, 0x22, 0x33, 0x40, 0x55, 0x66, 0x77, 0x80 };
```
drowe67, line 102: :grin: yes this drove me nuts with all that packing and unpacking - especially with the HF modes that have multiple codec frames for every voice frame!  Anyone using 800XA for data transfer will need to manually pack and unpack bytes.

unittest/tfreedv_mode_800XA.c
JeroenVreeken: Some of them might be a bit pedantic, but others like bits per modem frame are not pure constants, they are a calculated based on the bits per codec frame.
One thing I learned with unittests over the last few years is to especially test the stupid stuff. Its the first that gets broken :)

unittest/tfreedv_mode_800XA.c
JeroenVreeken: I was actually planning on expanding them further to also test the voice part (atleast to some degree) it was just the rawdata that I needed first.

unittest/tfreedv_mode_800XA.c
drowe67: Yes, good point, some are computed so that makes sense.

drowe67:
> Should we also update the library version to 0.10?
> Since the codecrx and codectx functions were renamed the API is not compatible anymore.

Good point :+1: we should bump the library version, we usually do that here: https://github.com/drowe67/codec2/blob/dr-data-demo/src/freedv_api.c#L56

Where is ```library version``` constant you mentioned? (I get lost myself will all the stuff in codec2 :smile: )

JeroenVreeken:
It is in the top level CMakeLists.txt

These two lines (starting at line 46):
set(CODEC2_VERSION_MAJOR 0)
set(CODEC2_VERSION_MINOR 9)


O 05/24/2020 09:37 PM, drowe67 wrote:

drowe67:
OK - so in this case bumping the FreeDV API version (with a note in the comments) is adequate.

drowe67:
@JeroenVreeken was there anything else you would like to add to this PR?  If not, the I'll take a final look then merge....

JeroenVreeken:
Not directly data related.
I think it is good to extend the unittest further with other parts of the api, but that can be done later.

drowe67:
@JeroenVreeken I renamed a few of the tests to indicate they are just testing raw modes at this stage. Also shuffled the order of the tests an bit so I can test all API stuff together e.g. ```ctest -R  test_freedv_api```.

drowe67:
@JeroenVreeken do you have any links your projects where you are using these FreeDV features?

PR #120, opened 2020-05-28
mainentance cleanup

drowe67:
1. Removed some unused files.
1. Updated files that are ```make install```-ed
1. ctest for 2020 memory leaks, fixed a few issues there.
1. Sort through list of header files that are ```make install```-ed.
1. ```.travis.yml``` now tests with LPCNet and does a test build of ```freedv-gui``` to catch API breaking.
1. A freedv 2020 AWGN ctest.


drowe67:
@hobbes1069 I'd like to adjust the files that get installed.  I figure that all we really need to install is libcodec2 and headers required to use it.  The executables are really only for developers who will be building from source anyway.  What do you think?

Can you pls tell me if the variable ```CODEC2_PUBLIC_HEADERS``` in ```src/CMakelists.txt``` is the list of all the ```.h``` files that will be installed?

hobbes1069:
Well, I put the binaries in the -devel package for Fedora. We could add a switch instead? We'll have to put a conditional around all the "add_executable"s.

Something like:
`option(BULD_BINARIES "Build and install development binaries." OFF)`

hobbes1069:
Yes, anything tagged as a public header gets installed. Do we need to adjust those?

drowe67:
> Well, I put the binaries in the -devel package for Fedora. We could add a switch instead? We'll have to put a conditional around all the "add_executable"s.
> 
> Something like:
> `option(BUILD_BINARIES "Build and install development binaries." OFF)`

Yes that would work.  However I'd like to take a look at which binaries we are including and why.  Some of the current ones are pretty obscure and most have been left out already (the list looks like it's based on where we were around 2015).

drowe67:
> Yes, anything tagged as a public header gets installed. Do we need to adjust those?

Yes I think so, maybe just the codec_* and freedv_*

drowe67:
In general, it would be good to understand the use cases for codec2 - when installed.

hobbes1069:
Well, as far as "in a package" there's really two use cases:

1. Most typically a -devel package is the minimum needed to build against that library/package (e.g. freedv & lpcnetfreedv).
2. To actually do development work with codec2. If this doesn't make sense (because so much is done from git/master instead of tagged releases) then perhaps we don't need to include them in the package.

drowe67:
@hobbes1069 I'm also concerned about namespace polution - e.g. we are installing kiss_fft headers, and generic complex number structs like ```comp.h```.  Thinking maybe everything should be ```codec2_*.h``` or ```freedv_*.h```

hobbes1069:
Not as worried about that since we're installing to /usr/include/codec2.

drowe67:
> Not as worried about that since we're installing to /usr/include/codec2.

OK - so if another ```comp.h``` exists we won't get a namespace type clash?

drowe67:
@tmiw I've messed with the ctests quite a lot recently, I was wondering if (when convenient) you could pls try the ctests in this PR on OSX?  I've removed a bunch of those "close force" calls as they stopped me using the scripts for regular (non Travis) debugging.  However now we are using text mode DISPLAY for ctests I'm, hoping they run OK on OSX....

tmiw:
@drowe67, I have one segfault:

```MacBook-Pro:build_osx mooneer$ ctest -V -R test_fsk_horus
UpdateCTestConfiguration  from :/Users/mooneer/devel/codec2/build_osx/DartConfiguration.tcl
Parse Config file:/Users/mooneer/devel/codec2/build_osx/DartConfiguration.tcl
UpdateCTestConfiguration  from :/Users/mooneer/devel/codec2/build_osx/DartConfiguration.tcl
Parse Config file:/Users/mooneer/devel/codec2/build_osx/DartConfiguration.tcl
Test project /Users/mooneer/devel/codec2/build_osx
Constructing a list of tests
test 46
    Start 46: test_fsk_horus

46: Test command: /bin/sh "-c" "cd /Users/mooneer/devel/codec2/octave; DISPLAY="" octave --no-gui -qf fsk_horus.m"
46: Test timeout computed to be: 1500
46: Generating 3 test packets
46: SNRdB meas:  4.0
46: VK5ARG-3,24,04:03:36,-34.875041,138.543884,9,32,6,1616,08 CRC OK
46: VK5ARG-3,24,04:03:36,-34.875041,138.543884,9,32,6,1616,08 CRC OK
46: VK5ARG-3,24,04:03:36,-34.875041,138.543884,9,32,6,1616,08 CRC OK
46: Received 3 packets
46: warning: gh_manager::do_close_all_figures: some graphics elements failed to close.
46: fatal: caught signal Segmentation fault: 11 -- stopping myself...
46: /bin/sh: line 1: 17531 Segmentation fault: 11  DISPLAY="" octave --no-gui -qf fsk_horus.m
1/1 Test #46: test_fsk_horus ...................   Passed    6.44 sec

The following tests passed:
	test_fsk_horus

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   6.45 sec
MacBook-Pro:build_osx mooneer$ 
```

Unfortunately it incorrectly shows as passing if all of them are run, so there might need to be some work done to fix failure detection.

tmiw:
So, I think if we include [FAIL_REGULAR_EXPRESSION](https://cmake.org/cmake/help/latest/prop_test/FAIL_REGULAR_EXPRESSION.html?highlight=regular_expression) for each of the tests with something like "caught signal", we should be able to catch future segfaults. Not sure if that should be part of this PR or a separate one, though.

hobbes1069:
We should not unless a project linked to two projects that both provided comp.h and both locations were in the include path. Probably better to fix in the long run but perhaps not an immediate issue.

drowe67:
@tmiw thanks - OK looks we are close. Odd that test fails, when other, very similar tests pass (e.g. the one above).

1480c1:
@drowe67 is there a particular reason why `uint8_t` was replaced with `u_char`? I see in https://github.com/drowe67/codec2/commit/72660c2a1aaa8f6d852b2b20428d1aaf954cad36 that you mention `an earlier compiler`, but do you know which one? Currently, compilation is broken on mingw-w64 gcc since `u_char` is [non-standard](https://stackoverflow.com/questions/1918934/is-u-char-a-standard)


```c
gcc -DHORUS_L2_RX -DINTERLEAVER -DRUN_TIME_TABLES -DSCRAMBLER -I. -I../src -mthreads -mtune=generic -O2 -pipe -Wall -Wno-strict-overflow -std=gnu11 -O3 -MD -MT src/CMakeFiles/freedv_data_raw_tx.dir/freedv_data_raw_tx.c.o -MF src\CMakeFiles\freedv_data_raw_tx.dir\freedv_data_raw_tx.c.o.d -o src/CMakeFiles/freedv_data_raw_tx.dir/freedv_data_raw_tx.c.o   -c ../src/freedv_data_raw_tx.c
../src/freedv_data_raw_tx.c: In function 'main':
../src/freedv_data_raw_tx.c:103:5: error: unknown type name 'u_char'; did you mean 'char'?
  103 |     u_char bytes_in[bytes_per_modem_frame];
      |     ^~~~~~
      |     char
```

drowe67:
@1480c1 thanks for pointing this out. TBH I was half asleep and puzzled about that one myself :smile: Happy to change it back to ```uint8_t```. Pls feel free to raise a PR or I'll do it myself in a day or so.

drowe67:
@1480c1 https://github.com/drowe67/codec2/pull/123

PR #121, opened 2020-05-30
Detect segfaults in Octave tests

tmiw:
Added logic to CMakeLists.txt to detect segfaults. Additionally, I updated one test to use octave-cli due to it faulting on OSX.

@drowe67, let me know if this needs any edits.


drowe67:
@tmiw Thanks for your PR.  You know I think I've seen it detecting segfaults on ctests without additional logic (on Linux), however I will check that and get back to you. I'd like to avoid to have to set up test properties for every test if possible (many of them trigger a fail simply on a non-zero exit code).

Using octave-cli is a good idea for the mystery test that is failing on OSX.  I've moved Travis back to Xenial as a "lowest common denominator".

drowe67:
I tried inducting a segfault in a regular C program ```fsk_demod.c```:
```
*((int*)0x12345678) = 1;
```

and the ctest system picked it up:
```
60: Test command: /bin/sh "-c" "/home/david/codec2/build_linux/src/fsk_get_test_bits - 10000 | 
                            /home/david/codec2/build_linux/src/fsk_mod 2 8000 100 1200 100 - - | 
                            /home/david/codec2/build_linux/src/fsk_demod -l 2 8000 100 - - | 
                            /home/david/codec2/build_linux/src/fsk_put_test_bits -p 99 -q -"
60: Test timeout computed to be: 1500
60: Segmentation fault (core dumped)
60: [0000] BER 0.500, bits tested      0, bit errors    0
60: FAIL
```

@tmiw so not sure why ctest on OSX doesn't pick up segfaults in Octave? I wonder if segfaults in C programs get trapped on OSX?  

At this stage I'm reluctant to merge the signal trapping code, but am OK with the octave-cli part.

tmiw:
@drowe67, is it possible it's something related to how the tests are being executed? I do remember some tests actually being detected as faulting without the workaround but I didn't see any specific pattern.

In any case, I'll revert the CMakeList.txt changes.

drowe67:
@tmiw were you going to leave the ```octave-cli```fix in for the one test that failed on OSX?  I understand that is enough to get a pass on OSX?

Yeah I'm a bit puzzled as to why that one test has a problem :thinking:

tmiw:
@drowe67, oops, was a bit overzealous with git checkout before. I readded the octave-cli fix.

Anyway, I haven't found anything online regarding ctest having similar issues on OSX. Might need further investigation at some point.

drowe67:
@tmiw Coincidentally this week I moved the horus code into a separate repo, so the test that was bombing on OSX has been removed!  So closing this is no changes required.  Thanks for your help :+1:

PR #117, opened 2020-05-23
Project Lockdown

drowe67:
This PR continues our fine work in https://github.com/drowe67/codec2/pull/101:
1. An octave debug utility for tricky FSK modem signals
1. Moving HAB code out of codec2 into it's own repo https://github.com/drowe67/hab
1. We didn't actually get very much Project Loickdown work done!  A future PR perhaps :smile: 

# Project Lockdown

A weak signal (-155dBm @ 10% Packet Error Rate) telemetry waveform/protocol.

This PR continues our fine work in https://github.com/drowe67/codec2/pull/101.

Our goal is to develop a FSK/LDPC telemetry system that can decode telemetry packets at a very low MDS on a sateillte AWGN + Doppler channel.  Use cases:
1. Recovery of HAB payloads on the ground where their LOS signal is blocked from terrestrial stations, but may be detectable from LEO satellites passing overhead.
1. Low MDS Telemetry for HAB missions.
1. Low SNR Terrestrial HF/VHF/UHF packet based communications.

# Requirements

1. Packet duration of roughly 20 seconds.
1. 128 and 256 bit payload options.
1. Tx capable of running on a reprogrammed RS41 STM32F100 (24 MHz with 64K flash, 8K RAM)
1. Tx modulation mFSK (e.g. Si4032 PLL controlled by STM32F100), 271Hz shift between tones.
1. Can run at Fs=24 kHz with Rs=25 Hz (large Fs/Rs ratios can be challenging).
1. Complex sample input, e.g. Fs=24kHz gives +/-12kHz range.
1. Rx acquisition/demodulation at +/- 10 kHz from carrier frequency (assuming no other CW interferers in passband).
1. Ability to track frequency change of +/- 0.01*Rs Hz/s.  In the satellite/HAB use case we assume the bulk of satellite Doppler is corrected externally to demod, leaving just a small residual rate/change.
1. Rx MDS < -150 dBm

# Waveform Design

```
256 payload bits, rate 1/3 code
256*3 = 768 bit codeword
Rb=768/20=38.4 bits/s (20s codeword duration)

Code will start to decode valid packets at around coded BER=5E-3 (1 error per codeword), which is Eb/No=1.5dB on a PSK channel.
The raw error rate can be calculated as:
EbNodB=1.5; rate=1/3; EbNodBraw=EbNodB+10*log10(rate); EbNo=10^(EbNodBraw/10); BERraw = 0.5*erfc(sqrt(EbNo)) = 0.165
[...]

PR #122, opened 2020-05-30
Horus API Refactor

darksidelemm:
- Simplify Horus API for freedv compatability (remove Rs argument in horus_open)
- Add new horus_open_advanced call with more options.


darksidelemm:
Note that the tests relating to building freedv-gui are passing because the master branch has the horus sections define'd out. 

I've raised a PR against freedv-gui ( https://github.com/drowe67/freedv-gui/pull/59 ) , which i've tested against this branch.

src/fsk.c
```diff
@@ -628,10 +628,11 @@ void fsk_demod_core(struct FSK *fsk, uint8_t rx_bits[], float rx_sd[], COMP fsk_
     modem_probe_samp_f("t_f_est",fsk->f_est,M);
     #endif
     float *f_est;
-    if (fsk->freq_est_type)
+    if (fsk->freq_est_type){
         f_est = fsk->f2_est;
-    else
+    }else{
         f_est = fsk->f_est;
+    }
```
drowe67, line 636: I don't really have any strong conventions here, for single lines lack of braces it can save some whitespace.

src/horus_api.c
```diff
@@ -68,16 +68,44 @@ int8_t uw_horus_rtty[] = {
   0,0,1,0,0,1,0,1,1,0
 };
 
-/* Unique word for Horus Binary */
+/* Unique word for Horus Legacy Binary */
 
-int8_t uw_horus_binary[] = {
+int8_t uw_horus_legacy_binary[] = {
     0,0,1,0,0,1,0,0,
```
drowe67, line 74: Lets discuss names of the protocols .... legacy/new doesn't leave room for expansion should we develop a 3rd or 4th version.  Maybe sequential letters horus_binary_a/horus_binary_b or numbers horus_binary_v1/v2?

src/horus_api.c
```diff
@@ -68,16 +68,44 @@ int8_t uw_horus_rtty[] = {
...
+    assert((mode == HORUS_MODE_RTTY) || (mode == HORUS_MODE_BINARY) || (mode == HORUS_MODE_BINARY_256BIT) || (mode == HORUS_MODE_BINARY_128BIT));
+
+    if (mode == HORUS_MODE_RTTY){
+        // RTTY Mode defaults - 100 baud, no assumptions about tone spacing.
+        return horus_open_advanced(HORUS_MODE_RTTY, HORUS_RTTY_DEFAULT_BAUD, -1);
+    } else {
+        // Placeholder until we have more definition of the new modes.
+
+        // Legacy Horus Binary Mode defaults - 100 baud, 270 Hz Tone spacing (RS41HUP Transmitter)
+        return horus_open_advanced(HORUS_MODE_BINARY, HORUS_BINARY_DEFAULT_BAUD, HORUS_BINARY_DEFAULT_TONE_SPACING);
```
drowe67, line 99: Hmmm, hard coding the protocol defaults to specific Tx hardware?  Maybe this should be an init-time value?

src/horus_api.h
```diff
@@ -35,20 +35,75 @@
...
+#define HORUS_BINARY_NUM_BITS                   360   /* fixed number of coded bits */
+#define HORUS_BINARY_NUM_PAYLOAD_BYTES          22    /* fixed number of bytes in binary payload     */
+#define HORUS_BINARY_DEFAULT_BAUD               100   /* Default baud rate of 100 baud */
+#define HORUS_BINARY_DEFAULT_TONE_SPACING       270   /* Default tone spacing of 270 Hz */
+
+// Settings for Horus Binary 256-bit mode (LDPC Encoding, r=1/3)
+#define HORUS_BINARY_256BIT_NUM_BITS            768
+#define HORUS_BINARY_256BIT_NUM_PAYLOAD_BYTES   32
+#define HORUS_BINARY_256BIT_DEFAULT_BAUD        100   /* Default baud rate of 100 baud */
+#define HORUS_BINARY_256BIT_DEFAULT_TONE_SPACING 270   /* Default tone spacing of 270 Hz */
```
drowe67, line 55: We have Rs around 25Hz in https://github.com/drowe67/codec2/pull/117

src/horus_api.h
```diff
@@ -35,20 +35,75 @@
...
+
+// Settings for Horus Binary 128-bit mode (LDPC Encoding, r=1/3)
+#define HORUS_BINARY_128BIT_NUM_BITS            384
```
drowe67, line 58: So is this just the FEC codeword, excluding the UW?

src/fsk.c
darksidelemm: This was because I was doing some debugging in this area trying to figure out if the mask estimator was working correctly. I can revert if necessary.

src/horus_api.h
```diff
@@ -35,20 +35,75 @@
...
-#define HORUS_MODE_RTTY              1
+/* Horus API Modes */
+#define HORUS_MODE_BINARY            0  // Legacy binary mode
+#define HORUS_MODE_BINARY_256BIT     1  // New 256-bit LDPC-encoded mode
+#define HORUS_MODE_BINARY_128BIT     2  // New 128-bit LDPC-encoded mode
+#define HORUS_MODE_RTTY              99 // RTTY Decoding
+
+
+// Settings for Legacy Horus Binary Mode (Golay Encoding)
+#define HORUS_BINARY_NUM_BITS                   360   /* fixed number of coded bits */
```
drowe67, line 46: Rather than having to explain what protocol #defines refer too with a comment, lets make it clean in the #define name - see comments above re protocol naming.

src/horus_api.c
darksidelemm: I figure the newer modes are still in flux for now, hence Legacy for the known-working mode. The newer mode names can be changed. 

I've left all the defines as HORUS_MODE_BINARY for the legacy mode for now.

src/horus_api.h
```diff
@@ -35,20 +35,75 @@
...
 struct MODEM_STATS;
 
-struct horus *horus_open  (int mode, int Rs);
+/*
+ * Create an Horus Demod config/state struct using default mode parameters.
+ * 
+ * int mode - Horus Mode Type (refer list above)
+ */
+struct horus *horus_open  (int mode);
+
```
drowe67, line 76: Does this behave as per V1 (peak estimator, so assumptions about Tx shift?)

src/horus_api.c
darksidelemm: Since there's no way of entering this kind of setting into freedv, we need to have default somewhere. This is the point of the simplified horus_open function - it uses these defaults.
horus_open_advanced provides the option to either use the default (by using -1 for Rs or tx_tone_separation), or providing a value.

src/horus_demod.c
```diff
@@ -180,7 +195,7 @@ int main(int argc, char *argv[]) {
 
     /* end command line processing */
 
-    hstates = horus_open(mode, Rs);
+    hstates = horus_open_advanced(mode, Rs, tone_spacing);
     horus_set_verbose(hstates, verbose);
```
drowe67, line 199: So do we need a ctest that exercises horus_open()?

src/horus_api.h
darksidelemm: Nope. 
I've made the default settings for the legacy horus binary mode take advantage of the new mask estimator with the known shift (270 Hz) of the only transmitter of this mode that we have in service - the RS41. 
The defaults for the new modes are subject to change.

src/horus_demod.c
darksidelemm: Maybe. Horus_demod doesn't use it. 

I'd say it's fairly low risk however.

src/horus_api.c
drowe67: ... so lets put some thought into it and sort the names out in this PR.

src/horus_demod.c
darksidelemm: Also freedv-gui isn't using it either, as I'd still like to enable some user-customization of the options (baud rate, mask/peak estimator)

src/horus_api.h
darksidelemm: I figure this is all subject to change. I would like to make use of the larger frame size at a higher baud rate as well as the lower rate. This would mean improved performance and expanded functionality while retaining the current update rate. I guess the selection of the 'robust' version (e.g. lower baud rate) could be somewhere else? Apart from the baud rate all the other settings will be the same.

src/horus_api.h
darksidelemm: Good point, not sure if this should include the UW bits or not. Will investigate furthere.

src/horus_api.h
drowe67: This and the HW lock-in doesn't work for me - when convenient lets have a phone call to discuss this and several other aspects raised in my review.

src/horus_api.h
darksidelemm: Aha, further down the buffer is set to double the number of expected bits: https://github.com/drowe67/codec2/blob/044daa39249c2c94470f0f1e1ec973409f0de71a/src/horus_api.c#L177

This ensures there's enough space for a UW and a packet.

darksidelemm:
I've gone with:
V1 = Legacy 22-byte Golay Encoded mode.
V2 = New LDPC Encoded modes, with 256-bit and 128-bit sub-modes.

darksidelemm:
I can confirm that all the test files can be received using this branch. I was also able to transmit using all modes (except HorusB of course).

src/horus_api.c
```diff
@@ -98,19 +138,42 @@ struct horus *horus_open (int mode, int Rs) {
...
+        }
+        hstates->uw_len = sizeof(uw_horus_binary_v1);
+        hstates->uw_thresh = sizeof(uw_horus_binary_v1) - 2; /* allow a few bit errors in UW detection */
         horus_l2_init();
         hstates->rx_bits_len = hstates->max_packet_len;
     }
-   
-    hstates->fsk = fsk_create(hstates->Fs, hstates->Rs, hstates->mFSK, 1000, 2*hstates->Rs);
+    // TODO: Horus 256/128-bit modes here.
+
```
drowe67, line 170: OK this PR is a work in progress.  So shall we just keep it open until the lockdown project (new protocols) is finished?  I'm no interested in merging incomplete work.  You can still run it out of this branch if you need to.

src/horus_api.c
```diff
@@ -98,19 +138,42 @@ struct horus *horus_open (int mode, int Rs) {
...
+    // Create the FSK modedm struct. Note that the low-tone-frequency parameter is unused.
+    #define UNUSED 1000
+    hstates->fsk = fsk_create(hstates->Fs, hstates->Rs, hstates->mFSK, UNUSED, tx_tone_spacing);
+
```
drowe67, line 174: This is starting to look messy (also in fsk_modem.c). Think we need a better way of handling this case in the fsk.h API
+ maybe a FSK_UNUSED #define in fsk.h
+ a way to init the demod with an unused tone separation value when we are using the peak estimator.  The FSK modem API should trap the case of an FSK_UNUSED tone separation and mask being enabled.

I'll have a think about this one.

src/horus_api.c
```diff
@@ -275,7 +338,7 @@ int extract_horus_rtty(struct horus *hstates, char ascii_out[], int uw_loc) {
 }
 
 
-int extract_horus_binary(struct horus *hstates, char hex_out[], int uw_loc) {
+int extract_legacy_horus_binary(struct horus *hstates, char hex_out[], int uw_loc) {
     const int nfield = 8;                      /* 8 bit binary                   */
```
drowe67, line 342: maybe ```extract_horus_binary_v1``` ?

src/horus_api.h
```diff
@@ -35,20 +35,78 @@
...
+// Settings for Legacy Horus Binary Mode (Golay Encoding)
+#define HORUS_BINARY_V1_NUM_CODED_BITS             360
+#define HORUS_BINARY_V1_NUM_UNCODED_PAYLOAD_BYTES  22
+#define HORUS_BINARY_V1_DEFAULT_BAUD               100
+#define HORUS_BINARY_V1_DEFAULT_TONE_SPACING       270  // This is the minimum tone spacing possible on the RS41 
+                                                        // reference implementation of this modem.
+                                                        // Note that mask estimation is turned off by default for 
+                                                        // this mode, and hence this spacing is not used.
+
```
drowe67, line 53: I'm still uncomfortable with this. @darksidelemm the idea of an API is to abstract out hardware, not hard code it in. I feel this belongs in the calling application (demo program or freedv-gui). This feels more like a decision made for a specific application, than an API.  If that's the case (a one-off application with no general use case) - then lets talk about moving ```horus_api.*``` out of codec2 and into the Horus repo.

src/horus_api.h
drowe67: Fine.  But  the reason I spent a few weeks in re-factoring and improving the FSK modem -  is the low SNR/new protocol aspect - and that's where I'd like to focus this v2/3 protocols.  So can we make your different update rates an option - and default to previously agreed goals of the project?

src/horus_demod.c
drowe67: If there is no use case for it should it be removed?

src/horus_api.c
darksidelemm: Yep missed this one.

src/horus_api.c
darksidelemm: The aim of this PR was to try and bring codec2 and FreeDV into some sort of alignment so that they both build correctly, while giving some of the improvements of the updated FSK modem. I decided to try and bring some further improvements at the same time.

src/horus_api.h
darksidelemm: The point of this was to provide some sensible default value which can be used if nothing else was provided by the user. I thought we may as well make it something that might provide some improvement in the current use-case of this modem, and it made sense to keep all these parameters together in one place, rather than splitting it up across multiple files.

By your argument, should we even be specifying a baud rate here? That's could be considered an implementation specific choice too...

src/horus_api.c
darksidelemm: Yes, I just used your approach from fsk_modem here...

src/horus_api.c
darksidelemm: Given that we're not using the mask estimator for V1 (since I guess the Eb/N0 area the mask estimator brings improvements in doesn't help when using a golay code) I could just change all the freedv changes to just call horus_open with horus_open(HORUS_MODE_BINARY,100), and be done with it...

However, I figured it would be useful to try and get the horus_api into the expected state that it would be in once these new modes are developed, so that some of the extra functionality that might be helpful in freedv (selection of horus mode, input of tone spacing to help mask estimator, display of frequency estimator outputs) could be developed in parallel, and tested using the V1 modem, or even with RTTY decoding.

I could also just remove all the references to the V2 modems. This PR would still provide an improvement.

darksidelemm:
Odd. The failing test (test_fsk_lib) passes on my machine, and I haven't touched any of that code... 

From travis output:

```
57/71 Test #57: test_fsk_lib ...................................***Failed  Required regular expression not found.Regex=[PASS
]  2.65 sec
octave: X11 DISPLAY environment variable not set
octave: disabling GUI features
Fs: 8000 Rs: 100 df  0.00 EbNo: 6.00 ftx:   9 frx:   8 nbits: 3584 nerrs:  96 ber: 0.027
```

From my machine:
```
test 45
    Start 45: test_fsk_lib

45: Test command: /bin/sh "-c" "cd /Users/darkside/Dev/codec2-upstream/octave; DISPLAY="" octave --no-gui -qf fsk_lib_demo.m"
45: Test timeout computed to be: 1500
45: Fs: 8000 Rs: 100 df  0.00 EbNo: 6.00 ftx:   9 frx:   8 nbits: 3584 nerrs:  63 ber: 0.018
45: PASS
45:
1/1 Test #45: test_fsk_lib .....................   Passed    5.54 sec

The following tests passed:
	test_fsk_lib

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   5.55 sec
```

darksidelemm:
I guess I just got unlucky with the random noise added to the signal in fsk_lib_demo. Passed this time!

donmr:
I used to have random generators print their seed value so that we could 
force them to repeat the same data when a test found something.

drowe67:
Horus moved to https://github.com/drowe67/hab

PR #124, opened 2020-06-03
src/CMakeLists.txt: Remove horus_api.h

1480c1:
Fixes issue with cmake not finding the file during an install

https://github.com/m-ab-s/media-autobuild_suite/issues/1695


drowe67:
thanks @1480c1 :+1:

PR #125, opened 2020-06-06
FSK modem support for TAP/TUN

drowe67:
# Introduction

Some exciting work going on with codec2 modems and VHF/UHF IP links:

1. Tomas's project: https://github.com/Tjoppen/codec2/tree/tms_tun uses a Pi and RTLSDR.
2. @JeroenVreeken  is also doing some fine work with his VHF packet protocol and a fork of freedv-gui https://github.com/JeroenVreeken/freedv-gui that supports a TAP device over COTs radios.

I'm using this PR to support them, and as a place for ongoing discussion.  I'd like to see a  "100 kbit/s IP link" for VHF/UHF using codec2 modems and  SDR. Moving Ham radio past the 1995-era "9600 data port" paradigm to real time IP.

# Plan

1. @Tjoppen I'm using this PR to mess with your modem samples.  First problem is I screwed up the documentation of the P parameter.  Have added more explanation in this PR.
2. Over The Bench (OTB) 100 kbit/s 2FSK test using Rpitx and RTL-SDR at UHF.

### DIFF
--- a/octave/fsk_demod_file.m
+++ b/octave/fsk_demod_file.m
@@ -18,17 +18,20 @@
    octave:2> fsk_demod_file("fsk.cs16",format="cs16",8000,100,2)
 #}
 
-function fsk_demod_file(filename, format="s16", Fs=8000, Rs=50, M=2, max_secs=1E32)
+function fsk_demod_file(filename, format="s16", Fs=8000, Rs=50, M=2, P=8, max_secs=1E32)
   more off;
   fsk_lib;
   plot_en = 1;
 
-  states = fsk_init(Fs, Rs, M);
+  states = fsk_init(Fs, Rs, M, P);
   
   if strcmp(format,"s16")
     read_complex = 0; sample_size = 'int16'; shift_fs_on_4=0;
-  elseif strcmp(format,"cs16")
+  elseif strcmp(format,"cs16") || strcmp(format,"iq16")
     read_complex = 1; sample_size = 'int16'; shift_fs_on_4=0;
+    states.fest_fmin = -Fs/2; states.fest_fmax = Fs/2; 
+  elseif strcmp(format,"iqfloat")
+    read_complex = 1; sample_size = 'float32'; shift_fs_on_4=0;
     states.fest_fmin = -Fs/2; states.fest_fmax = Fs/2; 
   else
     printf("Error in format: %s\n", format);
@@ -57,7 +60,7 @@ function fsk_demod_file(filename, format="s16", Fs=8000, Rs=50, M=2, max_secs=1E
     nin = states.nin;
     if read_complex
       [sf count] = fread(fin, 2*nin, sample_size);
-      if sample_size == "uint8" sf = (sf - 127)/128; end
+      if strcmp(sample_size, "uint8") sf = (sf - 127)/128; end
       sf = sf(1:2:end) + j*sf(2:2:end);
       count /= 2;
       if shift_fs_on_4
@@ -68,7 +71,7 @@ function fsk_demod_file(filename, format="s16", Fs=8000, Rs=50, M=2, max_secs=1E
         end
       end
     else
-      [sf count] = fread(fin, nin, "short");
+      [sf count] = fread(fin, nin, sample_size);
     end
     rx = [rx; sf];
     
@@ -120,7 +123,9 @@ function fsk_demod_file(filename, format="s16", Fs=8000, Rs=50, M=2, max_secs=1E
     axis([-Fs/2 Fs/2 mx-80 mx])
     xlabel('Frequency (Hz)');
     length(rx)
-    figure(2); Ndft=2^ceil(log2(Fs/10)); specgram(rx,Ndft,Fs);
+    if length(rx) > Fs
+      figure(2); Ndft=2^ceil(log2(Fs/10)); specgram(rx,Ndft,Fs);
+    end
     figure(3); clf; plot(f_log,'+-'); axis([1 length(f_log) -Fs/2 Fs/2]); title('Tone Freq Estimates');    
     figure(4); clf; mesh(Sf_log(1:end,:)); title('Freq Est Sf over time');
     figure(5); clf; plot(f_int_resample_log','+'); title('Integrator outputs for each tone');
--- a/octave/fsk_demod_file.m
+++ b/octave/fsk_demod_file.m
@@ -122,7 +122,6 @@ function fsk_demod_file(filename, format="s16", Fs=8000, Rs=50, M=2, P=8, max_se
     plot(f, RxdB);
     axis([-Fs/2 Fs/2 mx-80 mx])
     xlabel('Frequency (Hz)');
-    length(rx)
     if length(rx) > Fs
       figure(2); Ndft=2^ceil(log2(Fs/10)); specgram(rx,Ndft,Fs);
     end
--- a/octave/fsk_demod_file.m
+++ b/octave/fsk_demod_file.m
@@ -14,7 +14,7 @@
 
    Same thing but complex (single sided):
    
-   $ ./fsk_get_test_bits - 1000 | ./fsk_mod 2 8000 100 1000 1000 - - | ./cohpsk_ch - fsk.cs16 -100 --FS 8000 --complexout
+   $ ./fsk_get_test_bits - 1000 | ./fsk_mod 2 8000 100 1000 1000 - - | ./cohpsk_ch - fsk.cs16 -100 --Fs 8000 --complexout
    octave:2> fsk_demod_file("fsk.cs16",format="cs16",8000,100,2)
 #}
 
@@ -22,9 +22,8 @@ function fsk_demod_file(filename, format="s16", Fs=8000, Rs=50, M=2, max_secs=1E
   more off;
   fsk_lib;
   plot_en = 1;
-  
+
   states = fsk_init(Fs, Rs, M);
-  states.fest_min_spacing = 50000;
   
   if strcmp(format,"s16")
     read_complex = 0; sample_size = 'int16'; shift_fs_on_4=0;
@@ -37,7 +36,7 @@ function fsk_demod_file(filename, format="s16", Fs=8000, Rs=50, M=2, max_secs=1E
   end
 
   fin = fopen(filename,"rb");
-  if fin == -1 printf("Error opneing file: %s\n",filename); return; end
+  if fin == -1 printf("Error opening file: %s\n",filename); return; end
   
   nbit = states.nbit;
 
@@ -120,8 +119,8 @@ function fsk_demod_file(filename, format="s16", Fs=8000, Rs=50, M=2, max_secs=1E
     plot(f, RxdB);
     axis([-Fs/2 Fs/2 mx-80 mx])
     xlabel('Frequency (Hz)');
-
-    figure(2); specgram(rx,Fs);
+    length(rx)
+    figure(2); Ndft=2^ceil(log2(Fs/10)); specgram(rx,Ndft,Fs);
     figure(3); clf; plot(f_log,'+-'); axis([1 length(f_log) -Fs/2 Fs/2]); title('Tone Freq Estimates');    
     figure(4); clf; mesh(Sf_log(1:end,:)); title('Freq Est Sf over time');
     figure(5); clf; plot(f_int_resample_log','+'); title('Integrator outputs for each tone');
--- a/octave/fsk_lib.m
+++ b/octave/fsk_lib.m
@@ -190,7 +190,7 @@ function states = est_freq(states, sf, ntones)
   states.mask = mask;
   
   % drag mask over Sf, looking for peak in correlation
-  bmax = st; corr_max = 0;
+  b_max = st; corr_max = 0;
   Sf = states.Sf; corr_log = [];
   for b=st:en-length(mask)
     corr = mask * Sf(b:b+length(mask)-1);
--- a/src/cohpsk_ch.c
+++ b/src/cohpsk_ch.c
@@ -33,12 +33,12 @@
 #include <math.h>
 #include <errno.h>
 
+#include "codec2_fdmdv.h"
 #include "codec2_cohpsk.h"
 #include "comp_prim.h"
 #include "noise_samples.h"
 #include "ht_coeff.h"
 #include "ssbfilt_coeff.h"
-#include "codec2_fdmdv.h"
 
 #include "debug_alloc.h"
 
@@ -266,7 +266,7 @@ int main(int argc, char *argv[])
 	                          Channel
 	\*---------------------------------------------------------*/
 
-        fdmdv_freq_shift(ch_fdm, ch_in, foff_hz, &phase_ch, BUF_N);
+        fdmdv_freq_shift_coh(ch_fdm, ch_in, foff_hz, Fs, &phase_ch, BUF_N);
 
         /* optional HF fading -------------------------------------*/
 
--- a/src/fsk.c
+++ b/src/fsk.c
@@ -111,7 +111,7 @@ static void fsk_generate_hann_table(struct FSK* fsk){
 
 \*---------------------------------------------------------------------------*/
 
-struct FSK * fsk_create_core(int Fs, int Rs, int M, int P, int Nsym, int tx_f1, int tx_fs)
+struct FSK * fsk_create_core(int Fs, int Rs, int M, int P, int Nsym, int f1_tx, int tone_spacing)
 {
     struct FSK *fsk;
     int i;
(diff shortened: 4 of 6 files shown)


drowe67:
@Tjoppen - so I have taken a look at your sample (Fs=160kHz, Rs=16kHz, M=2):
```
$ cd codec2/octave
$ octave --no-gui
octave:1> fsk_demod_file("~/Desktop/tomas-001.iqfloat",format="iqfloat",160000,16000,2,10)
```

Note I have set P=10 (2 is too coarse - but I didn't document that well previously!)  Here is a plot of the spectrum:
Screenshot from 2020-06-07 08-22-40

Here is what I see:
1. The tone spacing is too small, you need a minimum of Rs=16000 between tones.
1. It's quite band limited. I suggest you don't filter at all, just keep the full 160 kHz bandwidth. Unfiltered 2FSK has pretty wide side lobes that the demod might like to slurp up.
1. The usual RTLSDR DC line - suggest you move your signal away from that as it may confuse the freq estimator, e.g. tune a bit low (shifting the signal up), then set the freq estimator limits to avoid 0Hz.

Here's a signal I generated with similar parameters:
```
~/codec2/build_linux/src$ ./fsk_get_test_bits - 32000 | ./fsk_mod 2 160000 16000 8000 16000 - - | ./cohpsk_ch - fsk.iq16 -100 --Fs 160000 -f -16000 --complexout --ssbfilt 0
octave:173> fsk_demod_file("../build_linux/src/fsk.iq16",format="iq16",160000,16000,2,10)
```
I used my channel simulator to convert the real signal to a complex one and shift it down. Note to self - I need a complex output version of ```fsk_mod```.

Screenshot from 2020-06-07 08-38-40
Screenshot from 2020-06-07 08-40-37
Screenshot from 2020-06-07 08-40-46

1. So you can see the lines at +/- 8 kHz
1. Good freq est outputs (just using the default peak estimator)
1. The integrator outputs are widely spaced (good eye diagram)

Tjoppen:
@drowe67 Strange that the DC line is there, I'm purposefully tuning 100 kHz below the center frequency of the signal, followed by a frequency xlating FIR filter. I'll have to tinker a bit more in gnuradio-companion. Something like shifting  so that the signal is between 50 kHz +- 8 kHz and telling the estimator to look between 40-60 kHz.

I am actually using a separation of Rs, see my calls to fsk_create_hbr() .

Thanks for fixing the documentation, P does almost the of opposite what I thought it did :]

Tjoppen:
That "DC" line is just the 0 tone. It just happens to end up there through the xlating process. I don't know why your spectrum plot looks bandlimited like that. If I do a waterfall plot in gnuradio then the noise level is quite flat across the entire 160 kHz, except for the edges of course.

Cranking P up to 10 does seem to improve decoding significantly. I still get a few missed bits. But one thing stands out: the sample clock offset has quite a bit of drift. In one test it's as high as -1000 ppm, when transmitting some packets at 1600 baud. Could that be an oscillator issue on the pi?

Tjoppen:
I can add that with Fs=160k, Rs=1600, P=100 things are significantly better. I think the next thing to try with this is locking nin when a preamble is detected, until the end of the packet has been reached. Fixed packets might also be a good idea, or something like smaller "sub-packets".

drowe67:
> That "DC" line is just the 0 tone. It just happens to end up there through the xlating process. I don't know why your spectrum plot looks bandlimited like that. If I do a waterfall plot in gnuradio then the noise level is quite flat across the entire 160 kHz, except for the edges of course.

Reconciling the two plots might help find the problem.

Re timing slips, could be caused by many issues.  It needs to be sorted before building up the higher layers and serious use OTA.  However this is tough, I have been here before :thinking: 

This is how I tested a similar system (Rpi/RTLSDR) on HF: http://www.rowetel.com/?p=6340, we've also done similar Over The Bench (OTB) tests for the balloon telemetry systems.

The Wenet SSTV system works at 100 kbit/s using the same modem, but with a different Tx.  Hmm, [that test](http://www.rowetel.com/?p=5080) might be worth repeating with the Pi.  I'll take an action to work up such a test using a Pi in 1-2 weeks (I'm in the middle of some HF data work right now).  Perhaps I can come up with a stable and tested physical layer configuration to support your work and @JeroenVreeken using the Pi/RTLSDR hardware.  Or at least find out if there are any issues with that hardware.

At 400MHz, the path loss over 100km is 125dB.  So at 14dBm output power and 0dBi Tx and Rx antennas we would have Rx power of 14-125  = -111dBm. The Wenet system + rate 0.8 LDPC code has a MDS of around -112dBm @ 100 kbit/s, so that's a LOS range of 100km. Plus/minus better antennas/poorer paths.

Tjoppen:
Did you ever figure out *why* the Pi outputs at the wrong baudrate? I presume the balloon TX just turns the serial line into FSK in some analog way, so any change in the CPU clock translates directly to a change in the baudrate.

Is it possible that there's some interaction with sendiq, the PLL and the CPU clock on the Pi? I know when you install rpitx it disables some GPU throttling thing to get it more stable. The strange thing is that the carrier doesn't move much. Perhaps it's a temperature issue.

Another thing to test is to TX over an audio cable, from the Pi to my laptop. Might also check what an "analog" type of FM demodulator spits out, in gnuradio.

Tjoppen:
Using the FM detector block in gnuradio and dumping the result in a .wav file and analysing it in Octave reveals a peak at 805.75 Hz for a 01010101.. sequence at 1600 baud, or 7200 ppm (!). This is for an on-air signal received with an RTL-SDR.

The generated signal sent to sendiq on the pi is an FM square wave, bang-on 800 Hz at Fs=192k as measured in Octave (slope detection followed by FFT). My main suspicion is that librpitx is quantizing 192k to some other rate and not informing the user about it. Perhaps fskburst would work better, since the FT8 tools in rpitx use it.

drowe67:
> Did you ever figure out _why_ the Pi outputs at the wrong baudrate? I presume the balloon TX just turns the serial line into FSK in some analog way, so any change in the CPU clock translates directly to a change in the baudrate.

Yes, there was an analog FSK modulator.  I can't recall the reason, probably a divider ratio issue.

drowe67:
> The generated signal sent to sendiq on the pi is an FM square wave, bang-on 800 Hz at Fs=192k as measured in Octave (slope detection followed by FFT). My main suspicion is that librpitx is quantizing 192k to some other rate and not informing the user about it. 

Nice work @Tjoppen, that might explain missing bits, 7200ppm is quite high!  There may be another baud rate Rpitx can approximate better.

Rpitx also approximates the frequency by dithering between the two PLL values, which may be the cause a lot of phase noise at UHF.  In order to make it compliant with spectral purity regulations and generate a cleaner signal I'm thinking about modifying Rpitx to select a "nearest frequency" the PLL can generate cleanly and report it.

If we can't make RpItx work to our satisfaction (due to hardware limitations) we can then consider some sort of low cost FSK Tx hardware.  However I think we're pretty close, and the results I obtained at HF were very encouraging.  Such a neat solution if we can get it to work :smile:

drowe67:
> Another thing to test is to TX over an audio cable, from the Pi to my laptop. 

Yes an Over the Bench (OTB) (aka Over The Cable) test is essential and will shake out a lot of bugs.  You can see how we did it in those blog posts.  You need the Pi, some attenuators, coax leads,  and the RTL-SDR.  A spec-an is useful to check levels.

You add attenuation until you start to get packet losses, and compare that to what you would expect from theory.  In the equation above for Rx MDS I forgot that without a LNA the RTL-SDR has a noise figure of about 8dB (with it's RF gain at max).

drowe67:
> I added a rational resampler in the gnuradio signal chain, works well enough. No need to complicate the modem with that I suppose. But maybe there can be a "middle ground" solution. I feel like "dithering" the size of each symbol ought to be possible.

I have lots of tricks for this.  Suggest we get the bugs out of the basic link using convenient high level tool kits (e.g. GNU radio or for my work I often use ```csdr```), then we can swap in more compact/embedded signal processing blocks that do the same thing.

drowe67:
@Tjoppen I've started setting up an over the cable test.  As a first step I'm looking at the 2FSK spectrum, using the ```SimpleTestBurstFsk``` function in ```rpitx/src/librpitx/app/testrpitx.cpp```:

```
void SimpleTestBurstFsk(uint64_t Freq) {
	int SR = 10000;
	float Deviation = 20000;
	int FifoSize = SR;
	fskburst fsktest(Freq, SR, Deviation, 14, FifoSize);

	unsigned char TabSymbol[FifoSize];
	for(int i=0; i<=FifoSize; i++)
	  if (i % 2)
	    TabSymbol[i] = 1;
	  else
	    TabSymbol[i] = 0;
	    
	while (running) {
	  fsktest.SetSymbols(TabSymbol, FifoSize);
	  fsktest.stop();
	}
}
```

Here is what I see at 144.2MHz using the SR=10000:
rpitx_rs1E4

I think this should be two tones, and at low symbol rates (e.g SR=100), that is what I see.  I'm wondering if the sample clock has some dithering in it.  I'll post to the Rpitx group.

drowe67:
Another test, at 100 kbits/s, using a different API function:
```
  float frequency=144.5e6;
  int SampleRate=100000;
  fmmod=new ngfmdmasync(frequency,SampleRate,14,FifoSize);
  padgpio pad;
  pad.setlevel(7);// Set max power
	
  float VCOFreq = 0;
  while(running) {
    playtone(VCOFreq);
    if (VCOFreq == SampleRate*2)
      VCOFreq = 0;
    else
      VCOFreq = SampleRate*2;
    fmmod->SetFrequencySamples(&VCOFreq,1);
 }
```

Same issue with extra lines, however when I using a wider resolution bandwidth the extra lines do form the FSK tones.  I'm wondering if this is timing jitter.  However the FSK may work OK, just at a few dB less that optimal as the tone energy is spread out a bit:

rpi100

drowe67:
OK I repeated my HF experiments from [a few years ago](http://www.rowetel.com/?p=6340) but at 10kbit/s and 144.5MHz.  First generate 60 seconds (600,000 bits), then "play" out of rpitx:
```
~/rpitx/src $ ~/codec2/build_linux/src/fsk_get_test_bits - 600000 | ~/codec2/build_linux/src/fsk_mod_ext_vco - 2fsk_10000.f 2 --rpitx 20000 10000
~/rpitx/src $ sudo ../freedv 2fsk_10000.f 144500000 10000
```
Note, ```rpitx/src/freedv.cpp``` hacked to remove the 10x sampling thing.  I also modified fsk_mod_ext_vco to make the low tone 0 offset, as the 2*20000 doesn't fit in a signed short. Might need a repo to put this code.

I then recorded 10s of the signal using gqrx (raw record feature) and an Airspy SDR, all running on my laptop.  Tuned the *hardware* freq to 144.490MHz, so the sampled tones sit at 10 and 30kHz. Resampled from 3Msps iqfloat to 100ksps iq16 with:
```
cat gqrx_20200619_233846_144490000_3000000_fc.raw | csdr fir_decimate_cc 30 0.05 HAMMING | csdr convert_f_s16 > rpitx_rs1e4_fs1e5_003.iq16
```
Then I ran Octave utility to get a good look at the signal:
```
octave:84> fsk_demod_file("~/rpitx_rs1e4_fs1e5_003.iq16",format="iq16",Fs=100000,Rs=10000,M=2,P=10,max_sec=3)
```
It looks fine.  The f1/f2 levels are a bit out, but that might be the resampler roll off.  Clock offset 50ppm, better than a lot of Txs I have seen. 

Screenshot from 2020-06-20 09-19-27
Screenshot from 2020-06-20 09-19-21
Screenshot from 2020-06-20 09-19-15

However connecting all this together is unwieldy, need to work out how to automate so I can do some MDS testing on the bench and make sure the physical layer is behaiving.  It's a bit wide to get through the gqrx UDP feature which is real Fs=48kHz (so 24kHz bandwidth).

@Tjoppen :point_up:

drowe67:
OK good command line rx results at 10,000 bits/s with:
```
Fs=1200000; tsecs=5; rtl_sdr -g 1 -s $Fs -f 144000000 - -n $(($Fs*$tsecs)) | csdr convert_u8_f | csdr shift_addition_cc `python -c "print float(-500000)/1200000"` | csdr fir_decimate_cc 12 0.05 HAMMING | csdr convert_f_s16 | ~/codec2/build_linux/src/fsk_demod -c -p 10 2 100000 10000 - - | ~/codec2/build_linux/src/fsk_put_test_bits -
<snip>
[0491] BER 0.000, bits tested  49100, bit errors      0 errs:    0 
[0492] BER 0.000, bits tested  49200, bit errors      0 errs:    0 
[0493] BER 0.000, bits tested  49300, bit errors      0 errs:    0 
[0494] BER 0.000, bits tested  49400, bit errors      0 errs:    0 
[0495] BER 0.000, bits tested  49500, bit errors      0 errs:    0 
[0496] BER 0.000, bits tested  49600, bit errors      0 errs:    0 
```

To avoid the dreaded DC line I tune the RTLSDR 500kHz low, then shift the FSK signal down. When I run the Octave demod, results are similar to above.  Running the Rx on my laptop at this stage. 

Next problem is to reduce the Rx level so I can do a MDS test.  Getting a lot of coupling across the bench so can't attenuate the signal beneath -50dBm :satellite:

drowe67:
OK, I managed to attenuate the signal down and obtained results within a 1-2 dB of theory, around -120dBm for 1-2% BER :grinning:   

The theory is:
```
MDS = Eb/No +10log10(Rb) + NF - 174
         = 9 + 10*log10(1E4) + 6 - 174
         = -119 dBm
Eb/No = Eb/No in dB for 2FSK modem for 1% BER
NF = Noise Figure of RTLSDR which is about 6dB at full gain
Rb = bit rate 
```
Only trick was getting enough attenuation, and the gain the gain of the RTLSDR e.g.  ```rtl_sdr -g 60```.  If the gain is too low the extra quantisation noise adds to the LNA noise.

Further work:
1. Mess with the transmit side tools to support 100 kbit/s
1. Try a 100 kbit/s MDS test
1. Try the Rx on another Pi (used my laptop so far)
1. Better integration of the  various applications
1. A "tuning GUI" would be useful, a way to see the FSK signal, spot any obvious problems like interferers, tuning issues, overload

Tjoppen:
@drowe67 Great work! :) I've been talking a bit in #raspberrypi-internals on FreeNode, we've bounced around ideas like generating 8-bit PCM using the VPU in the Pi, for fancier waveforms than FSK, especially at samplerates that sendiq can't do. Getting rid of the dithering means diving into PLL details. Limits the set of possible TX frequencies and baudrates but should produce a cleaner signal.

drowe67:
@Tjoppen nice work learning about the Pi internals, I've taken a peak at RpiTx but still coming up to speed.  We have some QAM work in progress for the OFDM modem, in future would be great to use that on VHF/UHF links.

I'm not sure about the freq dithering now, that may have been an error.  I think it's a fractional PLL, so all done in HW.

Re the extra spectral lines in the 01010 signal above, still not sure.  I haven't had a response from the rpitx group or author yet. Perhaps I should look at a similar signal upconverted by a HackRF to compare.  The 2FSK signal when modulated by random data (e.g. test frames) looks better, and the bottom line is good MDS/BER results.  Will publish a blog post on these tests shortly.

Tjoppen:
> @Tjoppen nice work learning about the Pi internals, I've taken a peak at RpiTx but still coming up to speed. We have some QAM work in progress for the OFDM modem, in future would be great to use that on VHF/UHF links.

QAM would be cool, but of course to have it work we need to get any phase noise issues in rpitx ironed out.

> I'm not sure about the freq dithering now, that may have been an error. I think it's a fractional PLL, so all done in HW.

Yeah the Pi has a fractional PLL. I'm not sure how much of the inherent phase noise when Frac != 0 makes it on to the air. Limiting the PLL to integer mode would cut down on that but limit the choices of output frequencies.

I'm waiting for a Pi binary from _clever_ in #raspberrypi-internals to test on the spec an at the ham club. It's some VPU code that does modulation and spits out DAC bits on GPIO pins, to be passed through a little R2R DAC. He's been able to get the VPU to generate signals up to 85 MHz, so the 6m band might be within reach with this method. This means no messing with the PLLs, just normal DSP stuff instead. This sacrifices video output on the Pi, but it's not really necessary here.

diego0815:
I tried to get myself a bit more clever at rpitx because it really sounds too good.
But so far it seems there is no real solution for AM components above 3 bits, other than using a second pin to drive an amp external to the first pin. Also a really interesting approach, having a pin for the carrier plus angle modulation and another pin for amplitude control.

Some interesting discussions regarding this can be found here: https://groups.io/g/rpitx/topic/how_does_am_ssb_modulation/28976987?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,28976987

Right now this won't affect this project yet, as it's FSK, but maybe at a later stage.

Very cool project, thanks for developing in the open!

diego0815:
That VPU approach sounds very interesting, it reminded me of this DVB-T proof of concept using the ramdac of a VGA card: https://bellard.org/dvbt/
A pitty rpi has no VGA port.

drowe67:
Reopening as place for discussion

drowe67:
Ooops this got closed by mistake.  We can still chat about this project here.

PR #127, opened 2020-06-11
Compiler warnings

JeroenVreeken:
While compiling with a recent gcc I got some compiler warnings.
Mainly about string sizes that could potentially overflow.
Also the tfifo test uses pthreads which are not available on non UNIX like systems (e.g. a bare embedded platform.)


drowe67:
Thanks @JeroenVreeken - yes I saw those warnings when we tried Travis using a more modern version of Ubuntu.

Re the UNIX/pthread conditional - can you please give me an example of a system that could run these tests (ie support all the other shell stuff/cmake/file I/O) but not have UNIX/pthreads? We have a different set of tests for the stm32 (a bare metal system), but they are still hosted by a PC.

JeroenVreeken:
I don't have an example of a system that could run them, but I was trying to build for an ARM (actually a stm32 variant) and wanted to use the cmake system 'out of the box'.
This was the only file that did not build due to lack of headers. All others result in pretty much useless programs, but they do build and seem to be relativly 'clean' C.
So with this 'fix' it becomes easier to crosscompile the library, only a platform .cmake file is needed.

PR #128, opened 2020-06-11
Add comptx for 800XA

JeroenVreeken:
Add freedv_comptx program
Add tests for comptx/comprx


src/freedv_api.c
```diff
@@ -431,6 +438,17 @@ void freedv_datatx (struct freedv *f, short mod_out[]) {
     }
 }
 
+void freedv_datacomptx(struct freedv *f, COMP mod_out[])
+{
```
drowe67, line 442: Is this function tested in this PR?

src/freedv_comptx.c
```diff
@@ -0,0 +1,144 @@
...
+    short speech_in[n_speech_samples];
+    int n_nom_modem_samples = freedv_get_n_nom_modem_samples(freedv);
+    short mod_out[n_nom_modem_samples];
+    COMP mod_out_comp[n_nom_modem_samples];
+
+    /* OK main loop  --------------------------------------- */
+
+    while(fread(speech_in, sizeof(short), n_speech_samples, fin) == n_speech_samples) {
+        freedv_comptx(freedv, mod_out_comp, speech_in);
+	for (i = 0; i < n_nom_modem_samples; i++) {
```
drowe67, line 128: So just one line changed from ```freedv_tc.c```?  Hmm, I guess the trade off is simplicity of the example versus too may modes and options.  However on balance I feel a ```--usecomplex``` mode for ```freedv_tx.c``` might be a better approach.

src/freedv_api.c
JeroenVreeken: No unittest yet, but I do use it with freedv_gui

src/freedv_api.c
```diff
@@ -381,8 +382,14 @@ void freedv_comptx(struct freedv *f, COMP mod_out[], short speech_in[]) {
...
-    if(FDV_MODE_ACTIVE( FREEDV_MODE_2400A, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_2400B, f->mode)){
-        codec2_encode(f->codec2, f->tx_payload_bits, speech_in);
+    if ((FDV_MODE_ACTIVE( FREEDV_MODE_2400A, f->mode)) || (FDV_MODE_ACTIVE( FREEDV_MODE_2400B, f->mode)) || (FDV_MODE_ACTIVE( FREEDV_MODE_800XA, f->mode))){
+        /* 800XA has two codec frames per modem frame */
+        if(FDV_MODE_ACTIVE( FREEDV_MODE_800XA, f->mode)){
+            codec2_encode(f->codec2, &f->tx_payload_bits[0], &speech_in[  0]);
+            codec2_encode(f->codec2, &f->tx_payload_bits[4], &speech_in[320]);
+        }else{
+            codec2_encode(f->codec2, f->tx_payload_bits, speech_in);
+        }
```
drowe67, line 392: Couple of concerns here:
1. We need to test voice modes carefully, e.g. actually run them manually and listen to the output audio.  Testing voice voice is a bit hard to fully automate :thinking: The ctests get us most of the way, e.g. counting frames gives the demod a good work out.
1. I don't think 2400B can operate with  complex modulator or demodulator samples, it operates in the baseband domain of analog FM modulators/demodulators by design.  I probably need to write a README_freedv.md to talk about this sort of detail,and  describe the waveforms more fully.

drowe67:
Thanks @JeroenVreeken - I think we can make 2400A/800XA work with a complex interface, but not sure about 2400B. 

I've also come across the need for a complex valued output ```fsk_mod.c``` as well.  Currently I'm using ```cohpsk_ch``` to do the real->complex conversion for some tests.

src/freedv_api.c
JeroenVreeken: I added a usecomplex flag to freedv_data_tx and freedv_data_rx and created testcases for them

src/freedv_comptx.c
JeroenVreeken: Indeed, a usecomplex flag was easier

src/freedv_api.c
```diff
@@ -431,6 +438,17 @@ void freedv_datatx (struct freedv *f, short mod_out[]) {
     }
 }
 
+void freedv_datacomptx(struct freedv *f, COMP mod_out[])
+{
+    assert(f != NULL);
+    if (FDV_MODE_ACTIVE( FREEDV_MODE_2400A, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_2400B, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_800XA, f->mode)) {
+        short out[f->n_nom_modem_samples];
+        freedv_tx_fsk_data(f, out);
+	for (int i = 0; i < f->n_nom_modem_samples; i++)
```
drowe67, line 447: @JeroenVreeken sorry about the delay in looking at this.  I saw the ctests had failed so didn't take a look until now.  Pls feel free to ping me if I'm slow in responding to an open PR.

```freedv_tx_fsk_data()``` returns real samples, so making a complex version is not as simple as just filling in the real part.   That would generate a two sided spectrum.  While you could write a ctest that passes, it's still a two sided signal and not complex/single sided.

I think this is possible to use the complex FSK modulator function for 800XA and 2400A but it's not meaningful for 2400B which operates on baseband real samples either side of an analog FM modem.

JeroenVreeken: Looking at it again I indeed think this pull request is wrong...
I actually got confused by freedv-gui a bit as it used comptx for everything except 800XA....
The exception is actually correct, but it should do it for more modes....

JeroenVreeken:
fixing the wrong problem in the wrong place...

PR #129, opened 2020-06-13
4FSK LLR estimation

drowe67:
Developing algorithms for 4FSK LLR estimation, in order to build 4FSK/LDPC modems.

## Instructions

1. 2FSK sample rate simulation:
   ```
   octave:4> fsk_cml_sam
   octave:5> fsk_llr_plot
   ```

1.  4FSK sample rate simulation:
    ```
    octave:4> M=4; fsk_cml_sam
    octave:4> fsk_llr_plot
    ```

## Future Work

1. ~~```SNRLin/v_est``` estimators in ```fsk_lib.m``~~
1. ~~Octave simulation of 4FSK modem with LDPC, compare BER to these simulations.~~
1. Optional OTA test with octave simulation to flush out any real world issues early.
1. ```SNRLin/v_est``` estimators in ```fsk.c```
1. Unit test for ```SNRLin/v_est```
1. Port CML C routines for ```SoMap/DemodFSK```.
1. Unit test for C 4FSK -> LDPC decoder, compare to these simulations.
1. Build/test real world C waveform that integrates frame sync/4FSK modem/LDPC, compare to these simulations.


octave/fsk_cml_sam.m
```diff
@@ -0,0 +1,275 @@
...
+    arx_symb = rx((s-1)*Ts + (1:Ts));
+    for m=1:M
+        r= sum(exp(-j*w(m)*(1:Ts)) .* arx_symb);
+        rx_pows(m)= r * conj(r);
+        rx_filt(s,m) = abs(r);
+    end
+    [tmp symbol_index] = max(rx_filt(s,:));
+    rx_sig_pow = rx_sig_pow + rx_pows(symbol_index);
+    rx_pows(symbol_index)=[];
+    rx_nse_pow = rx_nse_pow + sum(rx_pows)/(M-1);
```
drowe67, line 115: So the noise est is the sum of all the other filter energies apart from the chosen (max) symbol?

octave/fsk_cml_sam.m
```diff
@@ -0,0 +1,275 @@
...
+for s=1:Nsymbols
+    arx_symb = rx((s-1)*Ts + (1:Ts));
+    for m=1:M
+        r= sum(exp(-j*w(m)*(1:Ts)) .* arx_symb);
+        rx_pows(m)= r * conj(r);
+        rx_filt(s,m) = abs(r);
+    end
+    [tmp symbol_index] = max(rx_filt(s,:));
+    rx_sig_pow = rx_sig_pow + rx_pows(symbol_index);
+    rx_pows(symbol_index)=[];
```
drowe67, line 114: Is this the same as ```rx_pows(symbol_index)=0```?

octave/fsk_cml_sam.m
```diff
@@ -0,0 +1,275 @@
...
+        % use the SD amp estimates 
+        if plt==1, plot_pdf(v_est, sig_est);  end
+        llr = mfsk_sd_to_llrs(rx_filt(st:en,:), demapper, v_est, sig_est);
+    end
+    
+    if Ltype==4,
+        % use CML demod:   non-coherent; still seems to need amplitude estimate  
+        symL = DemodFSK(1/v_est*rx_filt(st:en,:)', SNRlin, 1);     
+        llr = -Somap(symL);  % now convert to bit LLRs 
+    end
```
drowe67, line 205: OK so these are the CML functions we would need to get going In C.  I think I already have a version of Somap. I have taken a look at the estimators for v_est and SNRLin - so no Genie estimators left - cool :slightly_smiling_face:

drowe67:
@billcowley I just tried:
```
    octave:4> fsk_cml_sam
    octave:5> fsk_llr_plot
```
Worked well :smile: and I obatined this plot:

Screenshot from 2020-07-02 07-26-32
 
Is the title correct?  an alternative would be to use ```legend``` to describe each plot if it is displayed.

billcowley:
Yes David, I removed all the genie estimates.  The CML doc indicates 
that amplitude estimates should not be required - but it seems they are 
(this is the "v_est"). CML doesn't need the other pdf parameter 
("sig_est" - BTW not the STD) - but my LLR routine needs it.

billcowley:
I started with legends, but then the "res" array started to contain too 
many different types of plots! The title should be correct but this plot 
routine is very simple at pres and can be much improved ....

billcowley:
@drowe67 I've updated the coded FSK simulations.   A brief summary:    fsk_cml.m is a symbol level simulation which can do M=2 to 16 and offers 3 different LDPC options.  This code executes quite quickly compared to the full simulation. It also plots the BER simulation results -- e.g these results only required about a minute on an old laptop (simulating 100 codewords, with length N=2048, K=1536,  so rate 3/4). 



fsk_cam_example

Two other RA code options are available - both quite short (K=256 data bits), either rate 1/3 or rate 1/2.  I might be able to improve the 1/2 rate performance as the present version seems to have an error floor.    @darksidelemm  Mark - are you interested rate 1/2, K=128 codes as well?  

For the sample level simulation (ie 16  samples/symbol with David's modem functions) - see fsk_cml_sam.m  which can replace David's original fsk_llr.m.   This simulation handles M=2 or M=4 and at present offers the same 3 LDPC options.  If offers up to 4 methods of generating the LLRs (controlled by "Ltype").  The first of these is David's original 2FSK method (black plot below) - which certainly is a very good approximation in the M=2 case. I wrote the 2nd option (green) by modelling the 2 or 4 FSK SD (Ricean) pdfs. The 3rd option is a rough HD model with fairly poor performance (blue plots below).   The magenta plots are from the 4th option: CML's SD to LLR method. This is a lot faster than mine.     No genie estimates are required in this simulation (for any Ltype value). 

fsk_cam_sam_example

Note that these results used the short rate 1/3 code.   These plots can be generated by running the fsk_llr_plot.m script which reads the "res" array after using fsk_cml_sam.m The number of bits simulated here was fairly low so the low-BER results are not reliable. Note that all these plots ran over the same Eb/N0 range, but if no errors are recorded, the plot is not shown.  For example the M=4 green and magenta options are probably offering similar performance -- many more independent error events are needed to get a clear picture.

darksidelemm:
Hi Bill!
If the rate r=1/2 codes perform nearly as well as the r=1/3 codes (as the CML docs suggest they might), then I'm definitely interested - it will mean less on-air time. Definitely interested in a K=128 code too, for 'short' packets.

octave/fsk_cml_sam.m
billcowley: Yes

billcowley:
> Hi Bill!
> If the rate r=1/2 codes perform nearly as well as the r=1/3 codes (as the CML docs suggest they might), then I'm definitely interested - it will mean less on-air time. Definitely interested in a K=128 code too, for 'short' packets.

The (standard) structure of a Repeat Accumulate code makes it hard to get the optimum degree distribution. Hence for some rates the performance of this RA method is not ideal.  We need RA for the uC implementation as it very easy easy to encode.  We don't need it for lots of other applications, but we now have lots of coding effort invested here! 
So I found a 2005 paper by the Newcastle group suggesting a different approach called w3IRA codes - this might help fix some error floor issues.  The RA structure is slightly changed, with a little more encoding effort - still fine for a uC.  This group have done great LDPC research of the last 20 years.  I've asked and they will check if some short codes and/or matlab design code is available to assist us.  Will keep you posted.

drowe67:
@billcowley - I'm just back from Wilpena and catching up.  Fine business with the plots above, and code tweaks.  Your LLR routines look very close to CML.  16FSK does indeed look inviting when compared to 2FSK :slightly_smiling_face: 

I've had a go at building an Octave simulation of a 4FSK modem that integrates your work with the ```fsk_lib.m``` modem, which includes freq and timing estimation and can run with real world off-air signals.  The choice of sample rate (Fs) and Symbol rate (Rs) was arbitrary. I have included the uncoded BER curves (adjusted for information rate) to see if the core uncoded modem is behaving:

fsk_lib_ldpc

To run the simulation:
```
octave:35> fsk_lib_ldpc_demo
```
You can adjust ```num_frames``` to make it run faster.  It only works with integer Eb/No steps at present.  

Lets compare with ```fsk_cml_sam```:
```
octave:43> Ctype=2; M=4; fsk_cml_sam
octave:44> fsk_llr_plot
```
plotname
At 6dB the results look close, both with coded BER about 4E-3

Next step is breaking this up into tx/rx scripts so we can test over the air :satellite: We'll also need to think about up and down conversion to the sample rates of our SDRs.

billcowley:
@drowe67    Good to see your new routine agrees pretty well :)  The coded BER can be quite steep - hence I was running some of the sims with 1/4 dB steps to compare the waterfall regions.

drowe67:
@billcowley it looks like we have more or less completed the Octave side of this project.  I'm inclined to merge at this stage and close this PR.  We can open a new PR for the Further Work.

If there is anything else you'd like to commit pls let me know otherwise I'll merge.

billcowley:
> @billcowley it looks like we have more or less completed the Octave side of this project. I'm inclined to merge at this stage and close this PR. We can open a new PR for the Further Work.
> 
> If there is anything else you'd like to commit pls let me know otherwise I'll merge.

Go ahead David.    BTW I had intended to delete fsk_llr_test.m -- I think those functions are included in fsk_clm_sam.m

drowe67:
> Go ahead David. BTW I had intended to delete fsk_llr_test.m -- I think those functions are included in fsk_clm_sam.m

:+1:  I'll also post to the Codec 2 mailing list and link your notes above - several other people were interested in this problem.

Would it be OK if I link to your notes on this PR from my blog as well?  That will make Google pick it up and provide a resource for others.  Alternatively - pls feel free to add a few notes/link on your blog and I can link to that!

octave/fsk_cml.m
```diff
@@ -0,0 +1,132 @@
...
+
+        if Npad; codeword = [codeword   zeros(1,Npad)]; end 
+        Tx = Modulate(codeword, S);
+        
+        kn = sqrt(1/(2*Eslin));
+        Rx = Tx + kn * (randn(size(Tx)) + j*randn(size(Tx)));
+        
+        SNRlin = Eslin;  % Valenti calls this SNR, but seems to be Es/N0
+        symL = DemodFSK(Rx, SNRlin, 2);    %demod type is nonCOH, without estimate amplitudes
+        bitL = Somap(symL);
```
billcowley, line 92: @darksidelemm @drowe67    If required, to save memory, these CML routines can also be called one symbol at a time:
 SNRlin = Eslin;  % Valenti calls this SNR, but seems to be Es/N0
	[NNr, NNc]= size(Rx);    
	bitL= zeros(1,NNc*bps); 
	for n42 = 1:NNc
	    symL = DemodFSK(Rx(:,n42), SNRlin, 2);    %demod type is nonCOH, without estimate amplitudes
            bitL((n42-1)*bps+[1:bps]) = Somap(symL);
	end

PR #132, opened 2020-07-08
All modes use continuous data buffers for rawdatarx()/rawdatatx().

JeroenVreeken:
Conversion functions are provided to get byte aligned codec frames.
freedv_mixed_tx and freedv_mixed_rx were changed to use the new functions.
Unittest for 700D rawdata was added.


drowe67:
@JeroenVreeken thanks for submitting this for consideration.  I have some concerns:

1. As I mentioned on the list - using the raw data interface with current voice oriented waveforms (like 700D) is an interim step until much more practical waveforms are available.  We will not be dealing with codec to modem frame mapping in future for data.  The primary use case of the raw data API  is HF data - you already have VHF data well covered.

1. I am hesitant about adding extra code/tests and several extra API functions to support operation of the raw data API with legacy voice oriented waveforms like 700D.  We should only add API functions when there is a clear need.  Small is good.

1. Who is the likely audience for the new API functions, will they see any use?  (Alan's use case is not strong enough for me)

1. As a compromise, perhaps we could (i) limit this to the FSK modes and (ii) remove the 700D specific code/tests (iii) remove or limit the new API functions from this PR?

So I guess I am asking - just because we can do this - should we?

JeroenVreeken:
For me the biggest reason is consistency...
But also a personal reason: I use the rawdata interface in my own repeater code as well. (it was the reason for adding the original codecrx/codectx functions. Until now I myself  only used if for 2400B and 6000, but the code itself is generic. For that kind of usage access to codec frames is needed 

Some of the API extensions  (like the nr of byte functions) I added for convenience thinking it would be better than doing +7 /8 all throughout the code. I don't mind removing them again, but they are handy to make sure buffers have the right size.

drowe67:
@JeroenVreeken - yes I hear you re those conversion functions.  They were driving me crazy inside the API as well :smile:   However I want be careful not to optimise the API too much for a single use case.  For example I recently removed the Project Horus code from Codec 2 as I realised I was coding/maintaining a single application.

I also agree with consistency, but feel that can be addressed with proper data modes (in particular for HF).  The current HF modes are not very useful for data due to their small payloads, and several other factors.

However i feel we're close with option 3 above - happy to move a bit in your direction as you've contributed a lot to this project!  Have just seen you latest commit https://github.com/drowe67/codec2/pull/132/commits/ce95f077bf30656726a2f022b0239d2d97b79ed8 - thanks.  Next lets take a look at the new API functions and see if we really need them - or if they would be better placed in the application that uses them.

PR #141, opened 2020-07-29
Begin data mode conversion

srsampson:
Continue development of OFDM modem C code to support HF data modes, which have been prototyped in Octave.

This work started in https://github.com/drowe67/codec2/pull/126 and https://github.com/drowe67/codec2/pull/138.

## Demos

1. Uncoded C Tx, Octave Rx:
   ```
   $ cd codec2/build_liunx
   $ ./src/ofdm_mod --mode datac1 --in /dev/zero --testframes 5 --verbose 1 > test.raw
   octave:66> ofdm_rx("../build_linux/test.raw","datac1")
   ```

1. Uncoded C Tx & Rx:
   ```
   $ ./src/ofdm_mod --mode datac1 --in /dev/zero --testframes 20 | ./src/ofdm_demod --mode datac1 --out /dev/null --testframes --verbose 2
   ```

1. Coded C Tx & Rx:
   ```
   $ ./src/ofdm_mod --mode datac1 --ldpc --in /dev/zero --testframes 20 | ./src/ofdm_demod --mode datac1 --ldpc --out /dev/null --testframes --verbose 2
   ```

## TODO

1. ~~C Tx to Octave Rx on first data mode.~~
1. Shoud we change  `ofdm_mod(struct OFDM *ofdm, COMP *result, const int *tx_bits)` to `ofdm_mod(struct OFDM *ofdm, COMP *result, const uint8_t *tx_bits)`?  I'm seeing a few char->int conversions about.
1. ~~Reconcile variable names between Octave and e.g. Nbitsperframe in ofdm_lib.m, which is equivalent to bitsperframe in ofdm.c, but different to Nbitsperframe in ofdm_mod.c~~
1. run stm32 tests (have had one pass at this, but do a final check)
1. ~~Clean up uncoded error counting in coded/uncoded case - uncoded case could perhaps use `count_uncoded_errors()`~~
1. Consider removing `rx_bits[]` output from `ofdm_demod()` and friends as I think we only ever use the `ofdm->rx_np[]` output.

PR #142, opened 2020-08-04
Data modes for OFDM modem

drowe67:
Continue development of OFDM modem C code to support HF data modes, which have been prototyped in Octave.  This PR continues on from https://github.com/drowe67/codec2/pull/141

## Demo - OFDM modem level

1. Error free packets at -4dB SNR:
   ```
   ./src/ofdm_mod --mode datac3 --ldpc --in  /dev/zero --testframes 60 --verbose 1 | ./src/cohpsk_ch - - -20 --Fs 8000 |    ./src/ofdm_demod --mode datac3 --ldpc --out /dev/null --testframes -v 1
   <snip>
   SNR3k(dB): -3.54 C/No: 31.2 PAPR: 10.4 
   BER......: 0.1082 Tbits: 36096 Terrs:  3905 Tpackets:    47
   Coded BER: 0.0000 Tbits: 12032 Terrs:     0
   ```
50 packet were sent, however I think a few were lost at the start (and maybe end) due to buffering/sync delays.  We can probably improve that, e.g. send a couple of dummy packets at the start so the demod get's modem sync early.  Might be useful to add sequence numbers to test payload data so we can work out what's missing.

## Demo - via FreeDV Raw Data API

1. No noise test frames:
   ```
   ./src/freedv_data_raw_tx DATAC1 /dev/zero - --testframes 10 | ./src/freedv_data_raw_rx DATAC1 - /dev/null --testframes
   ```
1.  Send some text:
    ```
    echo 'Hello World                    ' | ./src/freedv_data_raw_tx DATAC3 - - | ./src/freedv_data_raw_rx DATAC3 - -
    ```

## Todo

1. ~~configurable Unique Word and `datac3` mode.~~
1. ~~Data mode notes in `README_ofdm.md`.~~
1. ~~PER reported by ofdm_demod.~~
1. ~~First FreeDV API data mode.~~
1. ~~Add CRC.~~
1. ~~README_data.md documentation.~~
1. ~~Run stm32 unit tests.~~

## Further work for subsequent PRs

1. Adjust levels of all waveforms to have the same peak power. Try clipping to manage PAPR as per recent blog post research.
1. Add simulations for different HF channels, e.g. slowly varying channels, make sure it doesn't break/can get sync/histogram of acquisition time.  Experiment with longer codes to overcome fading. Consider better data state machine if required.
1. OTA real world tests to SDRs to test waveform.
1. Uncoded and coded QAM16.
1. Simplest possible demos in C and Python.


srsampson:
Is FreeDV 2020 31 carriers or 37?

The README_ofdm.md says 37 with example but the spreadsheet docs say 31.

drowe67:
Yes you're right it's 31, they are nicely summarised [here](https://github.com/drowe67/codec2/blob/master/src/ofdm_mode.c#L41)

drowe67:
@srsampson in this PR I refactored the FreeDV API code a little more (in freedv_700.c) to handle different OFDM modem configurations without new code.  So the same functions that handle 700D, also handle datac1/c2/c3.  So it's getting easier all the time to build up new modem waveforms and expose them at the FreeDV API level.

Am thinking about adding a 700E mode atm - two 40ms codec frames, short (112,56) LDPC code, and 5 symbols between pilots/6ms CP to handle fast fading.  So half the latency & sync time of 700D too.  Would have slightly poorer AWGN perf (more sync overhead) but better multipath perf.

srsampson:
Wow, yea 700E would be a killer. 6ms more better :-)

PR #145, opened 2020-08-29
Enable codec2 library to be built by non-C99 compilers. 

srcejon:
These changes enable the library to be built with non-C99 compilers (E.g. MSVC and some embedded compilers).

Changes include using the pre-processor to switch between variable-length arrays and using alloca. An abstraction layer around the use of complex float datatype was also added.

Currently I've only changed the code to enable the library to be built, not the demo applications. (But this is enough to get FreeDV working in SDRangel.)


drowe67:
@srcejon wow you have done a lot of work, I appreciate that's a big job and admire your skill. Congratulations on getting the code to run on SDRAngel.

I am thinking about this PR, and have some reservations:
1. I'm not sure I want to support non-C99 standard compilers like MSVC.  
1. There are good reasons for using C99 and standardising compilers - it makes cross-platform development easier and we don't need #ifdefs/abstraction.
1. We already have ways of building/running on Windows, e.g. we can build libcode2.dll which can be linked with non-C99 applications.  Is it possible to use this approach for applications like SDRAngel?
1. The changes will need to be supported into the future, which adds an extra support load (to me mainly).  I have no interest in any Windows-specific support work.
1.  It complicates coding for all future work, as different platforms/targets have to be considered and handled - every PR would need to be built and tested for both C99 and non-C99 targets.
1. I'm not sure we could run the ctests on the non-C99 build.
1. The changes would have to be tested/debugged on the stm32 platform.

I don't mind if you want to fork the code for non-C99, just not sure if this major change in the design in Codec 2 is something I'd like to merge.

I will think about it some more :thinking: 

@srsampson do you have any thoughts?

srsampson:
I think your suggestion is correct. This needs to be a fork, and not the master.

hobbes1069:
I could be convinced either way but my primary concern with PR's of this magnitude, is it can't be a merge and disappear. Someone needs to help actively maintain the changes.

srcejon:
RE the existing MingW build, that appears to be 32-bit only, whereas SDRangel is 64-bit. It may be possible to build in a different environment such as MSYS2, but that didn't work out of the box.

srsampson:
While I personally think the Variable Length Array is a beautiful thing, I realize it has been deprecated by most of the world. I think the alloca() with pointers is knife-cuts to Java and high-level language programmers who spend time in a pointerless world. A few years ago, the manic-dictator from Portland decreed that VLA in the Linux Kernel was a communist conspiracy. It was declared a security hole, and produced ugly machine code on all processors. I believe Linux is now VLA free.

srsampson:
In reviewing the complex number code, I don't see anything that is any worse than the current code. I think the "a + b" complex addition is not that better than "complexf_add(a,b)" which is reminiscent of the old days. Maybe re-wire kiss_fft to use the same syntax, and graduate from COMP structures everywhere.

So that's the major changes here: complex number rewire and VLA deprecation.

srcejon:
I suspect alloca has the same security and performance issues as VLAs. An alternative way to implement it, that should be more secure and result in smaller code, would just be to use statically sized arrays of the maximum size. This has the downsize of using more stack of course, but embedded targets could more easily determine how much stack is needed as it can be determined at compile time. The main problem for me with that though, would be working out what the max size should be, as I'm not familiar with the code.

donmr:
Static allocation also avoids any possability of fragmentation.  If you
want you can check that the current size is OK at key points on the code,
like the places you would put an alloc.

drowe67:
> RE the existing MingW build, that appears to be 32-bit only, whereas SDRangel is 64-bit. It may be possible to build in a different environment such as MSYS2, but that didn't work out of the box.

For freedv-gui, we build 32 and 64 bit versions of the libraries, however we use cross compilers running on Fedora Linux.

drowe67:
> While I personally think the Variable Length Array is a beautiful thing, I realize it has been deprecated by most of the world. 

I understand Linux has removed VLAs from the kernal, are there any other examples of them being removed/not recommended?

They make my life so much easier for numerical code - in fact I tend to remove mallocs for arrays when I see them in Codec 2.  I haven't run into any practical problems using VLAs - unlike malloc/alloc which are often a source of memory leaks.

drowe67:
> Static allocation also avoids any possability of fragmentation. If you want you can check that the current size is OK at key points on the code, like the places you would put an alloc.

@donmr - do you have any examples of fragmentation, say in Codec 2?  I've heard of the problem (albeit years ago) when using malloc/free a lot but never experienced myself.

I imagine VLAs are just allocating memory from the stack, so hard to imagine how fragmentation could occur.

donmr:
I don't have and specific examples.  I worry that when the application level
changes mode (or other settings) that the old space which is free'd may be
surrounded by other allocations and thus not reusable for the new mode.

I'm not talking about VLAs just an advantage of static allocation as was
also recently suggested.

srsampson:
> > While I personally think the Variable Length Array is a beautiful thing, I realize it has been deprecated by most of the world.
> 
> I understand Linus has removed VLAs from the kernal, are there any other examples of them being removed/not recommended?
> 
> They make my life so much easier for numerical code - in fact I tend to remove mallocs for arrays when I see them in Codec 2. I haven't run into any practical problems using VLAs - unlike malloc/alloc which are often a source of memory leaks.

When I read some articles about C++ it seems to be stated that VLA's were just a C99 thing, and not adopted by other standards. But I think GCC has implemented it in G++. In the case of the kernel they pretty much just used some MAX value that was big enough, rather than alloca(). I think the problem with VLA and alloca() is that you can get a stack exception, but I assume that's the case with picking a max value as well. Maybe not as many, or not as random??

srcejon:
> I don't have and specific examples.  I worry that when the application level
> changes mode (or other settings) that the old space which is free'd may be
> surrounded by other allocations and thus not reusable for the new mode.
> 
> I'm not talking about VLAs just an advantage of static allocation as was
> also recently suggested.
> …

Fragmentation is a problem when you allocate memory from the heap (via malloc). Memory allocated from the stack (via VLAs, alloca or auto arrays) will not suffer from fragmentation.

To clarify, I didn't mean the arrays should use static allocation (i.e. be declared with 'static') - but use a static size (i.e. fixed, size).

srcejon:
> In the case of the kernel they pretty much just used some MAX value that was big enough, rather than alloca(). I think the problem with VLA and alloca() is that you can get a stack exception, but I assume that's the case with picking a max value as well. Maybe not as many, or not as random??

VLAs/alloca are in some ways better than malloc for memory that only needs to be allocated within a single function, as you don't need to worry about calling free or fragmentation and it should be faster as well. However, unlike malloc, the allocation will never fail, so it's possible for a program to just crash randomly when it runs out of stack (if the arrays are larger than the page size, you might not get a stack exception - and some embedded targets don't have MMUs). Also, they're less secure, as it gives more ways for an attacker to modify the stack (E.g. by changing parameters to the function, they can adjust the stack pointer to whatever they want).

Fixed size arrays are even faster than alloca (often for free) and so code size can be smaller. They are more secure as they don't allow the calling function to change the size of the stack frame. The disadvantage is that they use more stack, but this usage is predicable at compile time (with -fstack-usage), which isn't the case with VLAs/alloca (as the size is only known at runtime).

srcejon:
> When I read some articles about C++ it seems to be stated that VLA's were just a C99 thing, and not adopted by other standards. 

Yep - they were made optional in the latter C11 standard (as was support for complex.h).

db4ple:
Hi, 

##VLA vs. alloca vs. fixed size arrays:

We can safely assume that in places where the VLA is in fact of constant size like in
```C
int K = 25;
int array[K];
```
the compiler will notice and there is no difference to `int array[25];` In case of enabled intra-unit optimization e.g. gcc's Link Time Optimizations, this can even work for array sizes passed as function parameters. 

In terms of potential stack overflow, all approaches (alloca, VLA, fixed size array) basically suffer from the same problem: If the stack is not big enough, you will get into trouble (read "undefined behavior").  Of course, it is a little easier to add up numbers if all are easily known. Unless we do run static code analysis, this is more of theoretical value.

In many cases the benefit automatic cleanup if getting out of scope outweighs the risk of stack overflow. In many embedded system memory configurations, heap (malloc) and stack (alloca) may collide without being noticed (this is for instance the case for our SM1000 code on the STM32). 

##Back to original discussion: 

Of course, it is a shame that MSVC doesn't support the complex data type and VLA. 
- I don't like the #ifdef'd solution to switch between VLA and alloca because it is a "One Software - Two Approaches required to solve the same problem" issue. For one recurring "problem" we should have only one solution which works in any case. I.e. fully switch to alloca or leaving it out entirely. Otherwise programmers may get confused, and maintainers have more work to check that rules have been followed (potentially with help from our CI). 
- In a similar way I don't like the fact that MSVC requires us to step back from native data types compilers have a clue about and are also a little easier to read in some places, but the impact is not as massive as the alloca I hope.

Keeping two branches as a long term approach seems to be put a lot of burden on the maintainer of the derived MSVC/NON-C99 branch, if larger reorganization happens on the main C99 branch. For now, possible, in the long run no really feasible unless the maintainer of the NON-C99 branch is willing to do the work, I would say. So we should try to get an accepted single source approach out of it or leave it unmerged.

@srcejon: Would a 64bit codec2 DLL build using mingw64 be a feasible solution for SDRangel (build in the same way we build the 32bit DLL for FreeDV), i.e. can this be used with MSVC?

hobbes1069:
> @srcejon: Would a 64bit codec2 DLL build using mingw64 be a feasible solution for SDRangel (build in the same way we build the 32bit DLL for FreeDV), i.e. can this be used with MSVC?

It's actually already done, in a way, it's just that it's built as part of a CMake subproject within the FreeDV MinGW builds, but no reason it can't be done separately.

From Fedora (probably should be added to the readme):

```
$ sudo dnf groupinstall "Development Tools"
$ sudo dnf install cmake
$ sudo dnf install mingw{32,64}-filesystem mingw{32,64}-binutils mingw{32,64}-gcc{,-c++} mingw{32,64}-crt mingw{32,64}-headers mingw{32,64}-pkg-config

$ git clone https://github.com/drowe67/codec2.git
$ mkdir codec2-win64 && cd codec2-win64
$ mingw64-cmake ../codec2
$ make
```

Something like that should work. The MinGW line actually installs both 32 and 64bit capability, but I copied and pasted that from my Fldigi howto :)

db4ple:
If this works, it would quite nice as it would let us keep the code "as is" at the expense of a little more complexity when building SDRangle, since the codec2 lib has to be build with a different tools set, but nothing proper automation (e.g. with docker) cannot solve with relatively little pain.

srcejon:
A working DLL can be built on Windows with gcc with the following:

Install MSYS2: https://www.msys2.org/
In MSYS2 Minf-w64 64 bit shell:
pacman -S mingw-w64-x86_64-cmake   (pacman -S cmake will install a version that doesn't work)
cmake .. -G "MSYS Makefiles"
make codec2
Rename libcodec2.dll.a to libcodec2.a and link against that with MSVC.
Need to copy msys64\mingw64\bin\libgcc_s_seh-1.dll and msys64\mingw64\bin\libwinpthread-1.dll alongside libcodec2.dll

drowe67:
Thanks @srcejon for your PR, and everyone for the interesting discussion.  I am pleased you found a way to build a working 64 bit DLL.  For the reasons explained above, I'm not inclined to merge, as I'd prefer to standardise on C11 and obtain windows compatibility using some of the techniques discussed above.

PR #153, opened 2020-11-01
bug fix writing outside of array bounds

LieBtrau:
Signed-off-by: Christoph Tack <prog.send@gmail.com>

I found this out when trying to port codec2 to ESP32.


drowe67:
@db4ple :point_up:

PR #154, opened 2020-11-15
Mac universal binary support for Codec2

tmiw:
Since Apple's officially announced devices with Apple Silicon now, I figure it's a good time to create a PR for this (and FreeDV). Codec2 universal binary support seems pretty trivial so we'll start with that.

@drowe67, BTW, make test doesn't work correctly on ARM yet because I haven't been able to get an ARM version of Octave built via MacPorts. Looks like it somehow tries to build a universal version of icu (despite having a non-universal one installed already) and fails during destroot. Would running the tests on my x86_64 Mac be sufficient, you think?


drowe67:
@tmiw OK sounds like a cool project.  Oh dear, it's a pity Octave is not available.  Re tests, I guess Travis will tell us if any generic changes break anything on x86.

tmiw:
@drowe67, I'm sure that'll eventually be fixed as more ARM based Macs get sent out to customers. At least on x86, Codec2 and FreeDV still seem to work fine, so I'm fine with merging.

PR #156, opened 2020-12-22
Make native build compatible with cmake within cmake build

JeroenVreeken:
Some small changes to the native build for cross compiling.
With these changes it is possible to build codec2 as a native lib(s) in android studio and integrate it into android apps.
Unfortunatly the slightly 'special' generate_codebook native build needs a bit of help to get it to work.
- Use 'current' instead of top level dir as this might not be codec2 in case of a nested build.
- Use the cmake prefered tool instead of make (android studio uses ninja)
- Choose our own build dir (android studios choice is wildly different from regular linux build and might break to install/copy action)
- Specify generate_codebook as a build byproduct as ninja does not understand build time generated dependencies as well as make

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,31 +200,32 @@ On Ubuntu 17 and above:
  ```
 
 
-## Including codec2 in an Android project
+## Including Codec 2 in an Android project
 
 In an Android Studio 'NDK' project (a project that uses 'native' code)
-codec2 can be added to the project in the following way.
+Codec 2 can be added to the project in the following way.
 
-1. Add the codec2 source tree to you app (e.g. in app/src/main/codec2)
-   (e.g. as a git sub-module)
+1. Add the Codec 2 source tree to your app (e.g. in app/src/main/codec2)
+   (e.g. as a git sub-module).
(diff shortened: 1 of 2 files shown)


drowemyriota:
Hi @JeroenVreeken looks good.  Would it be possible/useful to add an example usage of this feature in the README.md, e.g. an example of building on another platform like android?

JeroenVreeken:
Sure, I am building a test android app first to see if it works. I want to try a few more things like pulling codec2 from github automaticly. I'll add a description of this and a way to build as soon as I figure it out.

drowemyriota:
@JeroenVreeken thanks for the `README.md` example.  Just let me know when you're ready to merge

drowe67:
@JeroenVreeken Ping! :point_up:

JeroenVreeken:
This is as good as it gets right now.
I can do some sanity checks from within my android app and it all seems to work.
I had hoped to hook it up to a real audio source as well, but figuring out how that subsystem works will take some more time...
Lets merge this now

PR #157, opened 2020-12-23
Tuning HF data modes and (8192,4096) LDPC code

drowe67:
1. Unit tested Bill's new (8192,4096) code in Octave :heavy_check_mark: 
2. Designed `datac1_2` HF data mode using this code, approx 4 seconds per packet, 980 bits/ (7400 bytes/min).  Details on spreadsheet in `doc` directory. :heavy_check_mark: 
3. Tested mode in Octave works 5dB (@ 10% PER) better than previous datatc1 mode on MPP/MPD :heavy_check_mark: 

   | Waveform | Channel | SNR | Npackets | PER |
   | datac1 | MPP | 10 | 61 | 0.06 |
   | datac1_2 | MPP | 5 | 142 | 0.09 |
   | datac1_2 | MPD | 5 | 143 | 0.06 
   | datac1_2 | AWGN | 1 | 29 | 0.00 |
4. Port LDPC codes to C :heavy_check_mark: 
1. Replace `datac1` mode in FreeDV API with `data1_c2` :heavy_check_mark: 
1. Debugged LDPC `encode()` with low weight rows. :heavy_check_mark: 
1. Debugged Octave <-> interoperaibility issues that crept in with 700E.  Added ctests to trap in future. :heavy_check_mark: 
1. C support for DVBS2 (qam16) code .:heavy_check_mark: 
1. Redesign datac3 -> datac3_2. :heavy_check_mark: 
1. C support for new 4 second long datac3_c code. :heavy_check_mark: 
1. debug sync issue with datac3_2 - possibly two positions where UW errors are zero. :heavy_check_mark: 
1. PER measurements exclude last tx frame used for latency. :heavy_check_mark: 

## Further work (next PR)
1. Unit test for sync time of each data mode.  Measure probability of first frame sync at rated SNR.
1. FreeDV API facility to add preamble to improve first frame sync probability.
1. First frame sync probability requirements met.
1. Adjust clipper and Tx BPF for target peak power and PAPR.


drowe67:
@billcowley new code works great :1st_place_medal: ... see table above

I'm currently porting to C, but have an issue with a high raw BER, this example has no noise but a high raw BER (which the decoder is coping with) but is obviously a bug.
```
$ ./src/ldpc_enc /dev/zero - --sd --code H_4096_8192_3d --testframes 10 | ./src/ldpc_dec - /dev/null --testframes --code H_4096_8192_3d --sd
Raw   Tbits:  81920 Terr:   9950 BER: 0.121
Coded Tbits:  40960 Terr:      0 BER: 0.000
      Tpkts:     10 Tper:      0 PER: 0.000
```
Not sure if I'm running out of stack or something like that.  Same example works for other codes.

Can you pls confirm the C [encoder](https://github.com/drowe67/codec2/blob/master/src/mpdecode_core.c#L72) should work OK?

billcowley:
> @billcowley new code works great 🥇 ... see table above
> 
> I'm currently porting to C, but have an issue with a high raw BER, this example has no noise but a high raw BER (which the decoder is coping with) but is obviously a bug.
> 
> ```
> $ ./src/ldpc_enc /dev/zero - --sd --code H_4096_8192_3d --testframes 10 | ./src/ldpc_dec - /dev/null --testframes --code H_4096_8192_3d --sd
> Raw   Tbits:  81920 Terr:   9950 BER: 0.121
> Coded Tbits:  40960 Terr:      0 BER: 0.000
>       Tpkts:     10 Tper:      0 PER: 0.000
> ```
> 
> Not sure if I'm running out of stack or something like that. Same example works for other codes.
> 
> Can you pls confirm the C [encoder](https://github.com/drowe67/codec2/blob/master/src/mpdecode_core.c#L72) should work OK?

@drowe67 That's great you have the 8k code going already!  The high raw error rate sounds odd - I can try to look into it but my C skills are minimal ;)    Do you get raw errors if the info bits are all zeros?  ie this should give the all zero codeword and if not it might be easier to see what's going wrong.

drowe67:
> @drowe67 That's great you have the 8k code going already! The high raw error rate sounds odd - I can try to look into it but my C skills are minimal ;) Do you get raw errors if the info bits are all zeros? ie this should give the all zero codeword and if not it might be easier to see what's going wrong.

Well I've traced it to `ldpc_enc` encoder program.  I've added some debug code to that prints the crc16 of the `ibit` and `pibit` arrays as it runs.  `pbits` appears to change on alternate calls to `encode()`, this is add as it shouldn't have any state:
```
./src/ldpc_enc /dev/zero /dev/null --code H_4096_8192_3d --testframes 4
Using: H_4096_8192_3d
Nframes: 4
crc16 ibits:0xba6e pbits: 0x1ad4
crc16 ibits:0xba6e pbits: 0xc5bc
crc16 ibits:0xba6e pbits: 0x1ad4
crc16 ibits:0xba6e pbits: 0xc5bc
written: 32768
```
With other codes the `pbits` CRC is always the same.

drowe67:
@billcowley another possibility is something when wrong in the Octave-> translation which we do with:
```
octave:313> ldpc_gen_c_h_file("H_4096_8192_3d.mat")
```

drowe67:
@billcowley OK it's throwing up an invalid `ind==0` in [encode line 82](https://github.com/drowe67/codec2/blob/9d33afa0dabead49c809b73cb6aa06bce9f65e7f/src/mpdecode_core.c#L82):
```
 ./src/ldpc_enc /dev/zero /dev/null --code H_4096_8192_3d --testframes 2
Using: H_4096_8192_3d
Nframes: 2
unused_data_bits: 0 ldpc->NumberParityBits: 4096
i: 6 p: 0 ind: 0
ldpc_enc: /home/david/codec2/src/mpdecode_core.c:82: encode: Assertion `0' failed.
Aborted (core dumped)
```

This would means we are reading `ibit[-1]` which would explain the problem.

drowe67:
@billcowley, when convenient, could you pls review [ldpc_gen_c_h.m](https://github.com/drowe67/codec2/blob/master/octave/ldpc_gen_c_h_file.m#L90) ?  We might have messed up one the row or col weight constants.

billcowley:
> when convenient, could you pls review [ldpc_gen_c_h.m](https://github.com/drowe67/codec2/blob/master/octave/ldpc_gen_c_h_file.m#L90) ? We might have messed up one the row or col weight constants.

OK I will try to take a look.   I also have the parity check matrix in alist format

billcowley:
> when convenient, could you pls review [ldpc_gen_c_h.m](https://github.com/drowe67/codec2/blob/master/octave/ldpc_gen_c_h_file.m#L90) ? We might have messed up one the row or col weight constants.

For some reason the max row weight is wrong in your .h file -- it should be 11.  The max col weight is right (13)

billcowley:
@drowe67  My comment above was wrong -- the max row weight of the full H matrix is 11 but if we omit the RA structure (as the cml encoder needs) then the max col wt is 9. I'll need to return to this debugging as time allows!

billcowley:
@drowe67 Has ldpc_gen_c_h.m changed recently?  I think H_rows and H_cols should just follow the same  named variables in code_param, as given by Mat2Hrows().   But the order seems to have changed ie instead of writing the "1" positions in the first row (eg H_rows(1,:) =   68         283         371   ...   )    and H_rows(2,:)=  31         327        3678 ...)  in the C code we have const uint16_t H_4096_8192_3d_H_rows[] = { 68, 31,    ...    which is the first "1" position in row 1, then row 2 etc.  If the code worked before, I'm unsure what is happening.

drowe67:
@billcowley - `git log` will show you it's history.  That utility was developed in 2019 for the FreeDV 2020 work, IIRC we used `ldpc_gen_h_file.m` before that. Wow I wonder if eh C versions of the recent codes are messed up? Must be subtle - they do seem to correct errors OK.

srsampson:
I added a cast to the LDPC structure. The default integer may be overwriting memory in the automatic variable in ldpc_dec.c

srsampson:
I'm rusty on my C syntax for this particular cast. revert, and will think about it.

srsampson:
I played with the LDPC structure syntax, but then got:
```
$ ./src/ldpc_enc /dev/zero - --sd --code H_4096_8192_3d --testframes 10 | ./src/ldpc_dec - /dev/null --testframes --code H_4096_8192_3d --sd

crc16 ibits:0xdd0 pbits1: 0x9cbd
written: 81920
xtotal iters 505
Raw   Tbits:  81920 Terr:  40800 BER: 0.498
Coded Tbits:  40960 Terr:  20120 BER: 0.491
      Tpkts:     10 Tper:     10 PER: 1.000
```
Which seems worse...

drowe67:
@srsampson thanks - but I think @billcowley has zeroed in on the problem as something to do with the Octave->C mapping. I'm not strong in this area, but I note that In Octave we use the CML library encoder, but in C we use a more efficient encoder that may require a special mapping.  

Both C and Octave use the CML library decoder.

One mystery is why the current codes implemented in C work so well :thinking:

It's Xmas day here :santa: so we'll probably get back to this one in a few days time.

billcowley:
> @srsampson thanks - but I think @billcowley has zeroed in on the problem as something to do with the Octave->C mapping. I'm not strong in this area, but I note that In Octave we use the CML library encoder, but in C we use a more efficient encoder that may require a special mapping.
> 
> Both C and Octave use the CML library decoder.
> 
> One mystery is why the current codes implemented in C work so well 🤔
> 
> It's Xmas day here 🎅 so we'll probably get back to this one in a few days time.

Yep - I'm sure there's a simple explanation but it will need to wait a few days in my case.  When I looked yesterday it seemed that the H_rows and H_cols were not right. In that case decoding might still work but performance would not be good.  So in due course we will  need to check that the C version can do as well as octave eg BER of 1e-5 at EbN0=1.5 dB.  Merry Christmas!

drowe67:
OK I think I have some idea on what's happening.  About 5 years ago I ported the CML `.mex` C code to be called directly from C.  However it's set up for Matlab style Matrices which are stored column wise - so in memory A[0][1] is next to A[1][0], whereas in C A[0][0] is next to A[0][1].  Another subtlety is when we generate the C version of H_rows in `ldpc_gen_c_h_file.m` we read out the matrix as a linear array and let the CML decoder functions interpret it as a column wise matrix.

However we are using the non-CML `encode()` routine, I'm not sure what conventions that function is using.  Biggest puzzle of all is why the LDPC codes seem to work as designed in C :thinking:

drowe67:
OK so the issue is with the larger codes not all entries of each row are populated:
```
octave:1> ldpc;
octave:2> load H_4096_8192_3d.mat
octave:4> init_cml('~/cml/');
octave:5> cp = ldpc_init_user(HRA, 'QPSK', 4, 'gray');
octave:6> cp.H_rows
ans =

     68    283    371   1626   2725   3315      0      0      0
     31    327   3678      0      0      0      0      0      0
     76   1528   3995      0      0      0      0      0      0
```

So I just modified `encode()` to stop summing the parity if a 0 was encountered in H_rows.  Works well now :+1:

src/mpdecode_core.c
```diff
@@ -79,7 +79,7 @@ void encode(struct LDPC *ldpc, unsigned char ibits[], unsigned char pbits[]) {
 
         for (i=0; i<ldpc->max_row_weight; i++) {
             ind = H_rows[p + i*ldpc->NumberParityBits];
-            par = par + ibits[ind-1];
+            if (ind) par = par + ibits[ind-1];
```
drowe67, line 82: @billcowley :point_up:

billcowley: Good show David!  Don't know why this did not cause a problem before since many of the codes use irregular weights. Having seen the H_rows and H_cols structures I'll try to look in to this area in a few days just to understand why it's working - from your comments yesterday, maybe two effects are cancelling each other out(?)   Does the C code now give the same performance as the octave?  ie steep waterfall up to 2dB then error floor slightly above 1e-6?

drowe67: Thanks @billcowley - I looked at a few other codes and they seem to use all entries on each row.  Yes the (8192,4096) code is performing as expected in the unit tests - 1.5 to 2 dB better than the shorter rate 0.5 codes :smile: 

```
$ ./src/ldpc_enc /dev/zero - --code H_4096_8192_3d --testframes 100 --sd | ./src/ldpc_noise - - 1.5 | ./src/ldpc_dec - /dev/null --code H_4096_8192_3d --testframes --sd
Uncoded PSK Eb/No simulation:
No    =  1.50 dB (1.41 linear)
Eb    =  0.00 dB (1.00 linear)
Eb/No = -1.50 dB (0.71 linear)
Using: H_4096_8192_3d
Using: H_4096_8192_3d
Nframes: 100
CodeLength: 8192 offset: 0
.........x.x............................x......................................................written: 819200
..measured double sided (real) noise power: 0.705781
x..total iters 9102
Raw   Tbits: 819200 Terr:  95516 BER: 0.117
Coded Tbits: 409600 Terr:     15 BER: 0.000
      Tpkts:    100 Tper:      4 PER: 0.040
```

This is the form we use for the unit testing, simple BPSK simulation - I think Eb/No is raw/uncoded.

At 2dB coded Eb/No it is getting 1% PER, not sure about the coded BER, not enough digits to see 1E-6!  Will fix that soon.

Currently working on integration with the FreeDV API data modes.

drowe67:
@billcowley @srsampson have some initial results in C, about 10% PER at 5-6dB SNR on a MPD channel (2Hz/4ms), same as Octave.  Down to about 1dB AWGN.
```
./src/freedv_data_raw_tx DATAC1 /dev/zero - --testframes 25 | ./src/cohpsk_ch - - -21 --mpd --raw_dir ../raw --Fs 8000 | ./src/freedv_data_raw_rx DATAC1 - /dev/null --testframes

cohpsk_ch: SNR3k(dB):     5.66  C/No....:    40.43
cohpsk_ch: peak.....: 19219.25  RMS.....:  9362.75   CPAPR.....:  6.25 

modem bufs processed: 1027  output bytes: 11220 output_frames: 22 
BER......: 0.0980 Tbits: 182432 Terrs: 17874
Coded BER: 0.0000 Tbits: 90112 Terrs:     0
```

This is what the simulated signal [sounds](http://rowetel.com/downloads/codec2/hfdata/201226_datac1_6dbsnr.wav) like :alien: 

Still some optimization that could be done, e.g. PAPR compression, EsNo tuning for the decoder, and the sync state machine will need some work to make it reliable on real world channels.

Here is a plot of some other HF data systems from some recent tests the Winlink guys performed.  This is actual throughput, so including ARQ.  The y-axis is payload bytes per minute, which is about 7400 for datac1. None of the modes tested in this report worked on MPD, so they only tested up to MPP (1Hz/2ms).  I'm not 100% sure they are using the same SNR measurement (e.g. noise BW), but their simulator is similar.

Screenshot from 2020-12-26 12-47-54

srsampson:
Spectrum has a step at -12 dB down or so 200 Hz wide on each side..

drowe67:
> Spectrum has a step at -12 dB down or so 200 Hz wide on each side..

That might be due to some mild compression. Turns out if we allow the signal to be a little wider than strictly necessary we can get a good PAPR improvement.  There's a trade off between PAPR and bandwidth.

drowe67:
@billcowley re next code I'd like to see what we can do at -ve SNRs and around 300 bit/s payload data rate. If we try to use most of the 2000Hz RF bandwidth, and set a 4 second duration for the codeword, it suggests a (8192,1024) code.  There's a waveform design in the `datac3_2` tab of [modem_codec_frame_design.ods](https://github.com/drowe67/codec2/blob/dr-packet/doc/modem_codec_frame_design.ods)

I'm not sure this realistic, e.g. can the code really operate at a raw BER of over 20%?  Also not sure the OFDM Modem will hang on at those SNRs, but I will simulate the uncoded modem shortly and find out.

billcowley:
Yes @drowe67 : this is the tradeoff!  Shannon showed that the minimum possible Eb/N0 (-1.6dB) requires infinite BW - or zero spectral efficiency.  It's nicely shown by the attached plot - we can approach the Shannon bound (black) and generally must use spectrally efficient modulations unless we try to operate near the lower LHS of the figure.  This plot also shows some of the tighter constellation bounds, plus a few of the early iterative coding schemes, mostly developed at ITR many years ago.   (Note these bounds assume AWGN!)  So let me know how you go with the sync threshold then we can pick one or two codes that might suit.  
spectral_efficency_bounds

drowe67:
@billcowley I've run some Octave simulations using the current (9 carrier) `datac3` waveform design in the spreadsheet as a first pass.  The demod is currently falling over at -7dB, it fails to get UW sync. We might be able to improve on that.  There is about 2.8dB of loss due to pilot symbols and cyclic prefix (see spreadsheet), plus I estimate 0.5dB of implementation loss.  I have adjusted the curves for that.

Octave lines:
```
uber=[0.0352 0.0448 0.0665 0.0918 0.120 0.1545 0.23]
snr=0:-1:-6
EbNodB=snr+2.8; uber_theory=0.5*erfc(sqrt(10.^(EbNodB/10)))
figure(1); clf; semilogy(snr,uber,';sim;r+-'); hold on; semilogy(snr,uber_theory,';theory;g+-'); grid; hold off; xlabel('SNR (dB)'); ylabel('BER'); title('Uncoded BER versus SNR'); print('-dpng','ber_snr.png')
figure(2); clf; semilogy(EbNodB,uber,';sim;r+-'); hold on; semilogy(EbNodB,uber_theory,';theory;g+-'); grid; hold off; xlabel('Eb/No (dB)'); ylabel('BER'); title('Uncoded BER versus Eb/No'); print('-dpng','ber_ebno.png')
```

ber_snr
ber_ebno

I tried the wider, 28 carrier waveform but it falls over much earlier:

| SNR | UBER |
| --- | --- |
| 0 | 0.1533 |
| -1 | 0.19 |
| -2 | 0.22 |
| -3 | 0.44|

Hmm, so a wider waveform means the uncoded modem is struggling.  We can't run the timing/phase estimators over longer time windows because the channel is not stationary (fast fading).  I guess the alternative for low SNR channels is a narrow (in bandwidth) signal with a higher code rate.  If we make the codeword duration long enough, it will ride over fades that are hopefully time varying.  A slow fading channel with a notch in our signal might kill us.

Hmm - how about we just spread a small number of carriers over a wide bandwidth, rather than packing them next to each other?  Lots of unused spectrum in between. Then an 8 carrier waveform with a (2048,1024) code gets us a packet duration of 3.6.seconds.

srsampson:
> how about we just spread a small number of carriers over a wide bandwidth
There must be some name for OFDM with unmodulated carriers

Like COHFDM maybe use a wider gap between carriers at one end than the other
```
^....^....^...^...^..^..^.^^
```

drowemyriota:
@srsampson good point we could make the carrier spacing non-linear to help with PAPR :+1:

billcowley:
@drowe67   The HF channel is difficult compared to AWGN!  Ideally you want wide frequency separation to make the fading independent and long codewords/ interleaving compared to the average fade duration. In practice, a) I doubt that much bandwidth is available b) with lots of BW, the sync strategy would need to handle independent  fading on sub-carriers, and  c) latency constraints exist.  In addition the "best" solution will be dependent on channel conditions which are variable. 
Your software is already pretty flexible.  It seems sensible to make these HF parameters as malleable as possible and test over a wide range of channels.   This might translate to selecting a small set of candidates that can all be handled by the common software framework?  [Happy to chat later in the day on Hangouts etc]

drowe67:
Hi @billcowley look forward to talking with you later this morning re :
(a) If we just have 10 (50 Hz wide) carriers in 2000 Hz we have a reasonable distribution in frequency.  A 2ms delay between paths delay would give us notches 500Hz apart.
(b) Timing est uses all carriers, phase and ampl est is done on each carrier indep via pilots, UW frame sync is based on UW symbols spread over all carriers.
(c) Unlike PTT speech, we can add significant latency for data, I figure a 4 second codeword will cover us for >=1Hz Doppler.  Slow fading channel tend to be less severe (closer to AWGN).

drowe67:
@billcowley can the built in CML codes like `wimax` and `dvsb2` be encoded with our encode() C function?  Otherwise I guess we can call the CML encoder function.  Those built in codes are handy as we can dialup the code rate and codeword length at init time.

Here is where I'm at with some HF data waveforms that cover a range of SNRs.  The AWGN/MPP SNRs are the 10% PER point for those channels.  Note the big jump in SNR Required for QAM16, that's that QPSK->QAM16 Eb/No hit I guess.  Price we pay for bandwidth efficiency.

| name | coded (payload) bits/s | AWGN | MPP/MPD |
| --- | --- | --- | -- |
| datac3_2 | 283 | -4 | 0 |
| datac1_2 | 980 | 1 | 5 |
| qam16_2 | 3135 | 10 | 15 |

I tried 4FSK_LDPC at the low rate - it worked well on AWGN (-5dB for 200 bits/s) but had big problems with multipath,  I guess that's because it has no protection from ISI.

billcowley:
@drowe67  Hmmm .. I've never tried changing the rate or codeword length with the CML encoder!  The codes must be eIRA which use double diagonal rows on the RHS of H.  The original C code I wrote for this RA encoding will only work with fixed N and rate R, but the CML seems to be more flexible -- I just tried a very quick test of CML encoding (without any BER measurement) with my 8k code. Note that N and R can not be varied arbitrarily - eg using 1024 bits gives a rate 1/5 code.  Anyway this may be very useful for codec2 as you could use the optimised codes from DVBS2 etc.  I think your encoder is a converted version of the CML C encoder(?) However the re-ordering issue of H_rows and H_cols we spoke about may create a problem(?)  

For a start I'll try some BER tests of the CML encoder with adjustments to the nominal K and R params.  

re the FSK tests above - what bandwidth did you use for these low rates?  For minimum BW I expected the fading to be flat; with extra BW the fading for each carrier might be flat but the LLRs metrics for AWGN will not be suitable.  There might have been some good work done at ITR on this issue -- I'll try to find a paper on that.

drowe67:
@billcowley Yes I'm using the built-in CML codes for the Octave simulations of the `qam16` and `datac3` modes, they work well. For the `qam16` mode I tweaked the [DVBS2 code rate](https://github.com/drowe67/codec2/blob/master/octave/ofdm_lib.m#L1472) until I obtained the multi-path performance I needed.  Very neat, here is how you can run the `qam16` mode with the DVBS2 code at rate 0.6:
```
octave:363> ofdm_ldpc_tx("test_qam16c2.raw","qam16c2",30);
octave:364> ofdm_ldpc_rx("test_qam16c2.raw","qam16c2");
```

Also I have unit tests for these codes in `octave/ldpcut.m`

IIRC you supplied Mark and I with the [encode()](https://github.com/drowe67/codec2/blob/master/src/mpdecode_core.c#L72) function some time ago, as it was simple enough to run on micro-controllers in the HAB use case.  It's probably been tweaked to handle the re-ordering.

However to support the built in codes, I could move the CML encoder function across to the codec2 C library.  The apparent re-ordering is so the CML C functions see the arrays in the same column wise order as they expect from Matlab. So they are in the correct order for the CML encoder.

For the FSK tests I used Rs=200, 4FSK, rate 0.5 code, and tried 200 and 400Hz tone spacing (800 to 1600 Hz wide signal).  Noise bandwidth was 3000 Hz.  As well as fading of each tone, there's also the multi-path echoes which we have no protection from.  At Rs=200, each symbol is 5ms long, so a 2-4ms echo will cause some problems when we get two consecutive symbols with the same tone.  Good point about the decoder metrics.

billcowley:
@drowe67    I had hardly used the DVB S2 and Wimax codes from CML, but now I see they are great for a certain range of N values. For complete flexibility you could certainly convert those 2 CML initialise routines for use in codec2.    They are generating the H matrices "as required" using the protograph designs in those standards.  Of course if you select a  small set of N values and rates, then just store the resulting H matrices from the octave functions and use ldpc_gen_...  

So it would be nice to find some good protograph designs for other N values eg 4k, 8k - but I haven't managed that yet.  Various code is available eg https://shubhamchandak94.github.io/ProtographLDPC/usage.html  but I don't think it provides the eIRA format we have used so far.  I'll explore this a bit further.  

re the FSK option -- see the paper I sent via email.

drowe67:
@billcowley thanks for the paper - interesting read.  I guess we would still need a strategy for handling ISI.  The idea of using FSK for low rates on HF is intriguing, I feel we would get similar performance to PSK.

Re the CML codes, I'll look into ways to get them running in C (either via a function call at init time live Octave or via fixed arrays).

drowe67:
Hi @billcowley I've managed to get the CML built-in DVBS2 code running with our existing  `encode()` function.  However it appears the Wimax built-in code requires the fancier CML encoder function, which can handle the auxillary P matrix.

As an alternative, if it's not too much trouble, would it be possible pls for you to develop a (2048,1024) code (or thereabouts, as long as it's a multiple of 8 data-bits) that can run with our existing encoder?  That would let me move ahead without the additional MEX->C porting step for the CML encoder.

I need a code of about (2048,1024) for the datac3_2 waveform, to get a packet of about 4 seconds duration (see spreadsheet mentioned above).  It would also nicely complete our library of LDPC codes.

billcowley:
@drowe67    I could certainly make one, but the WiMax will be better.    It should be OK for 1/2 rate: eg see the help for this WiMax init function and try [H_rows, H_cols, P] = InitializeWiMaxLDPC( 0.5, 2304, [] );     BTW  N=2304 is the largest value. (The 3rd param is needed for some rates.)    I haven't run the BER simulation for this code - it should work but let me know if you have any trouble.  For consistency it would be nice to have the full H matrix for this code ie we need the inverse of mat2Hrows.m.  That should easy enough - I can take a look tomorrow.  

BTW I think the 5G "NR" LDPC codes might suit well but I can not find an open version of their H matrices. Once again, they are optimised and will perform better than anything I can design.  They should be available to K=8448 and I think they are compatible with the CML encoder (not 100% sure).

drowe67:
@billcowley yes I've run the (2304,1152) WiMax code in [Octave](https://github.com/drowe67/codec2/blob/master/octave/ldpcut.m#L34), and it works well.  Issue is I think it requires a different encoder to what we currently have running in C.  However I may be mistaken - I'll call you tomorrow to discuss just to make sure my assumptions are correct on this one!

billcowley:
@drowe67     You are right - it does require a different encoder.   I've  plotted the H matrix that comes from the WiMax H_rows and it's entries extend slightly into the RHS of H (where we usually have the double diagonal structure).    I guess the help for LdpcEncode() gives the clue to how this can work:  

> codeword    = LdpcEncode( data, H_rows, [P])
>        Where:
>        codeword = the encoded codeword
>        data   = a row vector containing the data
>        H_rows = a M-row matrix containing the indices of the non-zero rows of H excluding the dual-diagonal part.
>        P      = (optional) z times z matrix used to generate the first z check bits for WiMax (default =[])
 
This is new to me!   Clearly the 3rd param is required in this case. I guess the easiest approach is just to replicate the CML encoder functionality.  It would be helpful to have a clear definition of the eIRA structure.  I expect the WiMax spec must discuss this as well but I haven't searched for that yet.

billcowley:
@drowe67 I've added a rate half, N=2048, RA code to my Horus-FEC repo which you might like to try.  It's called H_1024_2048_4f.mat

Here are some sample simulations from aff3ct, generated by 

> ./aff3ct -C "LDPC" -D "BP_HORIZONTAL_LAYERED" -i 100  --dec-h-path '/Users/bill/work/radio/cml_ldpc_lowsnr/dec2020_4or8k/H_1024_2048_4f.alist' -N 2048 -K 1024 -m 0 -M 3 -s 0.25
......

>      Es/N0 |    Eb/N0 ||      FRA |       BE |       FE |      BER |      FER ||  SIM_THR |    ET/RT 
>       (dB) |     (dB) ||          |          |          |          |          ||   (Mb/s) | (hhmmss) 
>  ----------|----------||----------|----------|----------|----------|----------||----------|----------
>       -3.01 |     0.00 ||      103 |    15068 |      103 | 1.43e-01 | 1.00e+00 ||    0.088 | 00h00'01   
>       -2.76 |     0.25 ||      104 |    13964 |      103 | 1.31e-01 | 9.90e-01 ||    0.093 | 00h00'01   
>       -2.51 |     0.50 ||      122 |    12831 |      105 | 1.03e-01 | 8.61e-01 ||    0.103 | 00h00'01   
>       -2.26 |     0.75 ||      158 |    11789 |      104 | 7.29e-02 | 6.58e-01 ||    0.126 | 00h00'01   
>       -2.01 |     1.00 ||      397 |    11030 |      103 | 2.71e-02 | 2.59e-01 ||    0.237 | 00h00'01   
>       -1.76 |     1.25 ||     1503 |    10500 |      103 | 6.82e-03 | 6.85e-02 ||    0.523 | 00h00'02   
>       -1.51 |     1.50 ||    10281 |     7936 |      100 | 7.54e-04 | 9.73e-03 ||    0.975 | 00h00'10   
>       -1.26 |     1.75 ||    24817 |     1399 |      101 | 5.51e-05 | 4.07e-03 ||    1.311 | 00h00'19   
>       -1.01 |     2.00 ||    28978 |      478 |      100 | 1.61e-05 | 3.45e-03 ||    1.571 | 00h00'18   
>       -0.76 |     2.25 ||    42607 |      421 |      100 | 9.65e-06 | 2.35e-03 ||    1.818 | 00h00'23   
>       -0.51 |     2.50 ||    56211 |      423 |      100 | 7.35e-06 | 1.78e-03 ||    2.033 | 00h00'28   
>       -0.26 |     2.75 ||    63366 |      416 |      100 | 6.41e-06 | 1.58e-03 ||    2.237 | 00h00'29   
>       -0.01 |     3.00 ||    88516 |      425 |      100 | 4.69e-06 | 1.13e-03 ||    2.526 | 00h00'35    

As before it's generated via a random approach, with some adjustment of column weights, rather than protograph which would give better results but requires a different encoder. This code still has an error floor but it's not too bad; at Eb/N0=4dB the BER is ~1e-6 and the FER is ~2e-4

drowe67:
@billcowley thanks for that code.  I've integrated it into the Octave and C HF `datac3_2` mode.  It works as well as the (2304,1152) Wimax code, e.g. 0dB SNR, MPP channel:
```
./src/freedv_data_raw_tx DATAC3 /dev/zero - --testframes 30 | ./src/cohpsk_ch - - -20 --Fs 8000 --mpp --raw_dir ../raw |  ./src/freedv_data_raw_rx DATAC3 - /dev/null --testframes
cohpsk_ch: SNR3k(dB):    -0.15  C/No....:    34.62
BER......: 0.1196 Tbits: 65792 Terrs:  7870
Coded BER: 0.0209 Tbits: 31744 Terrs:   663
Coded PER: 0.0645 Tpkts:    31 Tpers:     2
```
(and down to -4dB AWGN).

PR #158, opened 2020-12-24
Support for data demos

drowe67:
Support for @DJ2LS  TNC work https://github.com/DJ2LS/FreeDV-Socket-TNC/tree/dev/test

## Fixed freedv_set_sync() for data modes

The data mode state machine doesn't have any logic to lose sync once a burst ends, as this is problematic for HF channels where frames may be lost due to fades.  Instead it relies on knowledge from higher layers about when the burst of frames from the Tx ends.  For example both the Tx and Rx may know that a fixed number of frames will be sent, or that the frames must arrive in a certain time window.

Here is a demo of freedv_data_raw_rx resetting sync after `resetsync` frames have been received.  Three bursts of 3 frames are sent by the Tx.  The Rx resets the state machine after it receives 3 frames, so it is ready to sync up on the next burst.

```
~/codec2/build_linux/src/freedv_data_raw_tx DATAC3 /dev/zero - --testframes 3 --bursts 3 | ~/codec2/build_linux/src/freedv_data_raw_rx DATAC3 - /dev/null --resetsync 3 -vv | hexdump
<snip>
modem bufs processed: 171  output bytes: 270 output_frames: 9
``` 

This method will fail when a frame is lost.  A real world TNC could use a timer:
1. Start Rx running.  Modem is waiting for burst of frames to start.
1. Start a timer with a period equivalent to the taken to receive N frames.
1. When N frames arrive, or timer fires, reset modem state machine.  Modem is now ready for a new burst of frames.

### DIFF
--- a/src/freedv_700.c
+++ b/src/freedv_700.c
@@ -520,7 +520,7 @@ int freedv_comp_short_rx_ofdm(struct freedv *f, void *demod_in_8kHz, int demod_i
     ofdm_sync_state_machine(ofdm, rx_uw);
 
     if ((f->verbose && (ofdm->last_sync_state == search)) || (f->verbose == 2)) {
-        fprintf(stderr, "%3d nin: %4d st: %-6s euw: %2d %1d mf: %2d f: %5.1f pbw: %d snr: %4.1f eraw: %3d ecdd: %3d iter: %3d "
+        fprintf(stderr, "%3d nin: %4d st: %-6s euw: %2d %2d mf: %2d f: %5.1f pbw: %d snr: %4.1f eraw: %3d ecdd: %3d iter: %3d "
                 "pcc: %3d rxst: %s\n",
                 f->frames++, ofdm->nin,
                 ofdm_statemode[ofdm->last_sync_state],
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1248,7 +1248,7 @@ void freedv_set_carrier_ampl(struct freedv *f, int c, float ampl) {
 void freedv_set_sync(struct freedv *freedv, int sync_cmd) {
     assert (freedv != NULL);
 
-    if (FDV_MODE_ACTIVE( FREEDV_MODE_700D, freedv->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_700E, freedv->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_2020, freedv->mode)) {
+    if (freedv->ofdm != NULL) {
         ofdm_set_sync(freedv->ofdm, sync_cmd);
     }
 }
--- a/src/freedv_api.h
+++ b/src/freedv_api.h
@@ -64,7 +64,7 @@
 // peak (complex) sample value from Tx modulator
 #define FREEDV_PEAK             16384
 
-// Return code flags for freedv_*rx* functions
+// Return code flags for freedv_get_rx_status() function
 #define FREEDV_RX_TRIAL_SYNC       0x1       // demodulator has trial sync
 #define FREEDV_RX_SYNC             0x2       // demodulator has sync
 #define FREEDV_RX_BITS             0x4       // data bits have been returned
--- a/src/freedv_data_raw_rx.c
+++ b/src/freedv_data_raw_rx.c
@@ -47,12 +47,14 @@ int main(int argc, char *argv[]) {
     int                        mode;
     int                        verbose = 0, use_testframes = 0;
     int                        mask = 0;
+    int                        resetsync = 0;
 
     if (argc < 3) {
     helpmsg:
       	fprintf(stderr, "usage: %s [options] FSK_LDPC|DATAC1|DATAC2|DATAC3 InputModemSpeechFile BinaryDataFile\n"
-               "  -v or --vv      verbose options\n"
-               "  --testframes    count raw and coded errors in testframes sent by tx\n"
+               "  -v or --vv              verbose options\n"
+               "  --testframes            count raw and coded errors in testframes sent by tx\n"
+               "  --resetsync  Nframes    reset sync after Nframes received\n"
                "\n"
                "For FSK_LDPC only:\n\n"
                "  -m      2|4     number of FSK tones\n"
@@ -73,6 +75,7 @@ int main(int argc, char *argv[]) {
             {"Rs",         required_argument,  0, 'r'},
             {"vv",         no_argument,        0, 'x'},
             {"mask",       required_argument,  0, 'k'},
+            {"resetsync",  required_argument,  0, 's'},
             {0, 0, 0, 0}
         };
 
@@ -91,6 +94,9 @@ int main(int argc, char *argv[]) {
             break;
         case 'r':
             adv.Rs = atoi(optarg);
+            break;
+        case 's':
+            resetsync = atoi(optarg);
             break;
         case 't':
             use_testframes = 1;
@@ -178,6 +184,8 @@ int main(int argc, char *argv[]) {
             fwrite(bytes_out, sizeof(uint8_t), nbytes-2, fout);
             nbytes_out += nbytes-2;
             nframes_out++;
+            if (resetsync && (nframes_out >= resetsync))
+                freedv_set_sync(freedv, FREEDV_SYNC_UNSYNC);
         }
 
         if (verbose == 3) {


drowe67:
@DJ2LS :point_up_2:

DJ2LS:
@drowe67 Thanks for the Xmas present 🎅

I updated TEST_RX.py and the TNC. 
Seems to work!

I'm doing a "NO SYNC" after a successful frame has been received by checking it like this:
`
nbytes = freedv_rawdatarx(freedv, bytes_out, data_in)
if nbytes == freedv_get_bits_per_modem_frame(self.freedv)/8
`
so I can be sure, we have received a full frame and then force a resync. But because of the CRC16 check, the comparison between nbytes and bytes_per_frame shouldn't matter that much, I guess. 

I will do some more tests!

drowemyriota:
> so I can be sure, we have received a full frame and then force a resync. But because of the CRC16 check, the comparison between nbytes and bytes_per_frame shouldn't matter that much, I guess.

Yes it will only return `nbytes != 0` if the full frame has been received, and CRC matches.

DJ2LS:
Okay, I tested mode 10,11,12. 

A short explanation of the terminal output of my TNC test:
AUDIO LEVEL = Indicator for modulated data over the audio sinkhole which is coming in. 0 = no input
NIN = nin
SYNC = Sync state of freedv
13.1453534535 = Time stamp, which started with the program and is printed every time, the program loops through data


After a frame has been successful received, freedv_set_sync(UNSYNC) will be set. FreeDV seems to recognize this, as we can see directly after the received frame. But one step later, FreeDV switches back to SYNC state, which will be released after some more cycles. This behavior is equal for mode 10 and 12. For mode 11 its the same, but FreeDV seems to be stuck in SYNC state. 


Is it possible, to reduce the "lock" time until FreeDV releases the SYNC state? 
Or is there a reason, why the state machine switches back to SYNC state after a 



```
AUDIO LEVEL: 0
NIN : 880
SYNC: 0
AUDIO LEVEL: 4155
NIN : 880
SYNC: 0
AUDIO LEVEL: 4095
NIN : 352
SYNC: 0
AUDIO LEVEL: 4247
NIN : 880
SYNC: 1
13.02912449836731
AUDIO LEVEL: 4118
NIN : 880
SYNC: 1
....
AUDIO LEVEL: 3068
NIN : 880
SYNC: 1
14.180505752563477
AUDIO LEVEL: 0
NIN : 880
SYNC: 1
b'HALLO\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
14.181149959564209
AUDIO LEVEL: 0
NIN : 880
SYNC: 0
AUDIO LEVEL: 0
NIN : 880
SYNC: 1
14.412214517593384
AUDIO LEVEL: 0
NIN : 836
SYNC: 1
....
AUDIO LEVEL: 0
NIN : 880
SYNC: 0
AUDIO LEVEL: 0
NIN : 880
SYNC: 0


```

drowe67:
@DJ2LS OK I've taken a look at the more verbose internal logs using this example:
```
~/codec2/build_linux/src/freedv_data_raw_tx DATAC3 /dev/zero - --testframes 1 --bursts 3 | ~/codec2/build_linux/src/freedv_data_raw_rx DATAC3 - /dev/null --resetsync 1 --vv
```
I think this is happening:
1. Line 12 "BS" flags: It finds the first frame and returns it to us, and we manually unsync.  When sync is reset the modem starts looking for candidate signals.
1. It finds a candidate straight away (line 14) (might be previous audio data still in the buffer) and enters an internal "trial sync" state.  This is reported at the API level as "sync", which made sense for coded voice modes where we want to start decoding voice as soon as possible.
1.  After a few seconds it realises this is a bad candidate (L26) so resets itself back to "search" and looks for the next candidate.  
1. It finds a genuine candidate on L36, which is probably the next frame from the Tx.

A way to get higher resolution information on sync state is  `freedv_get_rx_status()`, this returns some bit-flags defined on L67 of `freedv_api.h`.

You're making me think some more about the modem state machine for data :thinking:  Genuine use cases help - thanks.  It's a tricky design issue so we might need a few iterations to get the design right.

At some point we should also work out why it finds a candidate straight away, e.g. we might need to clear some more states when we unsync.  With noise or no signal it shouldn't find candidates very often.

DJ2LS:
@drowe67 
I'm not sure if you already did it, but could you merge the commits to master ? Then I could use the default build process to update everything automatically with my default setup script.

DJ2LS:
Hi @drowe67 ,

I'm currently testing a lot with audio and sample rate conversion with some examples from the KISS TNC Project and my own. 
I'm still wondering why some things are working sometimes and why not. 

So it seems, that decoding modes 10, 11, 12 is working just randomly when using real audio in my personal case ( could be possible, that I'm wrong ). Sometimes its working, sometimes not. Could be possible this if 


- Do you have any advice for me, which data is best as preamble? I'm not sure, but could it be possible, that there's a difference between sending just 000000, 11111111, 123456789 or different chars? Sometimes it feels, that mode 700D for example is decoding a little bit better, if we don't have repeating input data. 

- My thoughts are to modulate some random data as preload and append the real modulated data to the preamble raw data. Does the CRC matter in case of using modes 10-12 for the preamble or is it enough if there's just raw data of the same mode?

- Does it make sense, if a preamble feature will be a part or feature of codec2? Maybe as  an additional parameter of freedv_rawdatatx() or just as a default part of the modulation functions?  Or do you think its best if this will be done by upper layers? 


Thanks 😃

drowe67:
Hi @DJ2LS,

It's hard to comment on your problem without being able to reproduce it.

My tests with the C demo programs indicate the modes sync up on the first frame when the SNR is high.   At lower SNRs sync can be delayed past the first frame.

In future I would like to add a preamble to improve first frame sync, probably a few "modem" frames in length (much smaller than the frames the API uses, which are based on the FEC codeword size). Yes, this would be a [FreeDV API function](https://github.com/drowe67/codec2/blob/master/src/freedv_api.h#L186), which is already implemented for the FSK_LDPC raw data mode.

I'm planning to write a test for preamble/sync time that runs many bursts on a simulated channel with noise and measures the probability of first frame sync.  About 99% would be a good target.  There will be some cases that are hard to handle (like deep fades).  In that case the ARQ will need to schedule re-transmission.

As we discovered a few weeks ago, we'll need to adjust the data modes, API programming models etc, to get these data modes into a form suitable for real world use.  I'm currently doing some work on the physical layer over on https://github.com/drowe67/codec2/pull/157.  I'll then take a look at the preamble and first frame sync.

DJ2LS:
@drowe67 ,
Yes, sorry for the not reproducible problems with the preamble and sync. I wish I could reproduce the problems a little bit better and give you better feedback!  
Maybe someone else is reading this thread and has an equal behavior.  

I will keep on testing, looking forward,  and having a look at the FSK_LDPC examples!

drowe67:
@DJ2LS some detailed use cases or prototype code might be a good start.  So we can start thinking about a suitable design for the FreeDV API functions to ease TNC development.

DJ2LS:
@drowe67 
I'm actually creating a new, small TX and RX program for testing with audio. I need to reorganize my tests a little bit....
I will publish and document them as soon as I found a best practice for the highest chance of a successful decoding.
I think a document with frame rates, buffer sizes and so on could be interesting as well. The last weeks I did some experiments with asynchronous multimode decoding which have been partially successful. But those experiments need some more time in future. They are the basics for the ARQ module.

Regarding to the freedv_api functions: I think the functions are good, if you know what you have to do. More detailed examples/documentation will improve this ( I think if I'm able to create some more nice code examples, improving the existing ones, this could be a good way )


Regarding to the preamble function:
`int  freedv_rawdatapreambletx      (struct freedv *f, short mod_out[]);`
I think, if this feature would be part of freedv_rawdatatx it would be nice. I'm not sure, If I'm right ( I still need to create a preamble test ) but the preamble could be necessary for every transmission?

Maybe some additional status could be interesting. 
Something like "FreeDV frames received, but no sync/....." or something like this. 

I need some more time to do the things above. Corona is not getting better here and we have a lot of snow outside, which is not that common here. So I will try to enjoy this a little bit before we have a harder lockdown. 
Just4Info: I renamed my Github Repository to [FreeDV-JATE](https://github.com/DJ2LS/FreeDV-JATE) --> Just Another TNC Experiment. This name fits a little bit better for what I'm doing and we have less confusion because of the KISS-TNC project.

DJ2LS:
Here are my test files:
https://github.com/DJ2LS/FreeDV-JATE/tree/PREAMBLE-TESTS

Am I right, that this are the following rx states as an integer?:

```

0 - NO SYNC
2 - SYNC
3 - TRIAL SYNC ( or trying to sync )
6 -  SUCCESSFULLY DECODED
10 - ?  
```

Is there a rx state for a bad CRC?

DJ2LS:
After some tests this afternoon I got the following results:
#### Mode 7:
- using the total modulated data as preamble seems to be working well for mode usage as a signalling frame. I think this could work fine :+1: 

#### Mode 12:
- using the first 100 bytes of modulation as a preamble increased the chance of receiving
- every 32 receiving cycles ( one cycle == NIN ) we get the integer 10 from freedv_get_rx_status() --> I'm doing a freedv_set_sync(UNSYNC) then
- This results in a chance of a succesfull decoding with maybe 75% --> But this is not statistical approved, since I only doing tests manually and not automated at this point
- If I'm sending while we don't have UNSYNC state, decoding fails


I was just testing a little bit where I thought it could be useful. If you need more specific information, please let me know. Maybe I can help. The total debugging output of all sync states is too long, I guess. So I'll wait for your request to send more accurate output.




#### A general question about using the data mode:
If we have a larger amount of data to send than one frame can handle, lets say 64bytes to send for mode 12 with 32 bytes per frame:
Should I send frame per frame by sending every "modulation package" which results in a small interruption between them and maybe an unsync/resync state, or should I stick them together to one big audio stream so we don't have an interruption between them?

drowe67:
> Regarding to the preamble function:
> `int freedv_rawdatapreambletx (struct freedv *f, short mod_out[]);`
> I think, if this feature would be part of freedv_rawdatatx it would be nice. I'm not sure, If I'm right ( I still need to create a preamble test ) but the preamble could be necessary for every transmission?

Yes, pls see the documentation on the [FSK_LDPC](https://github.com/drowe67/codec2/blob/master/README_data.md#fsk-ldpc-raw-data-mode) mode for an example.

> Maybe some additional status could be interesting.
> Something like "FreeDV frames received, but no sync/....." or something like this.
> EDIT: I think freedv_get_rx_status() is already doing this?

This function returns the [rx_status](https://github.com/drowe67/codec2/blob/master/src/freedv_api.h) flags.

drowe67:
> Here are my test files:
> https://github.com/DJ2LS/FreeDV-JATE/tree/PREAMBLE-TESTS
> 
> Am I right, that this are the following rx states as an integer?:
> 
> ```
> 
> 0 - NO SYNC
> 2 - SYNC
> 3 - TRIAL SYNC ( or trying to sync )
> 6 -  SUCCESSFULLY DECODED ( MODE 12 )
> 7 - SUCCESSFULLY DECODED ( MODE 7 )
> 10 - BAD DECODING 
> ```
> 
> Is there a rx state for a bad CRC? Or is it 10 ?

It's best to interpret them as flags, not integers (see link above for definition of each flag).

[Here](https://github.com/drowe67/codec2/blob/master/src/freedv_700.c#L461) is where the flags are set when the CRC is checked.  No bytes will be returned from `freedv_rawdatarx` unless the CRC is valid.

Please note the earlier voice modes have been deprecated for data use, the only modes I'm moving forward with are FSK_LDPC/DATAC1/DATAC2/DATAC3, as per the latest README_data.md.

The earlier voice modes have no CRC and frame size too short to be useful.  If you need a shorter frame for some reason - lets discuss that use case.

drowe67:
> After some tests this afternoon I got the following results:
> #### Mode 7:

Re these tests - with no noise (and no preamble) we should be able to get 100% sync on first frame, and 100% data throughput.  In future I'll write a test to check this at the C code level.

It would be useful to discuss use cases for the FreeDV raw data API so we can build up a set of tests, and tune the API to make it easier to use.

> If we have a larger amount of data to send than one frame can handle, lets say 64bytes to send for mode 12 with 32 bytes per frame:
> Should I send frame per frame by sending every "modulation package" which results in a small interruption between them and maybe an unsync/resync state, or should I stick them together to one big audio stream so we don't have an interruption between them?

You should send a burst of continuous frames, here is an example for the [FSK_LDPC](https://github.com/drowe67/codec2/blob/master/README_data.md#fsk-ldpc-raw-data-mode) mode.

BTW the number of bytes per frame has just changed, as I've been tuning the FEC codewords.  I'll update README_data.md now.

drowe67:
@DJ2LS just sent you an email with some ideas for Use Cases :slightly_smiling_face:

DJ2LS:
@drowe67 
I'm creating the preamble part actually, 
would it be possible, to get n_preamble with an own function like
`freedv_get_n_nom_modem_samples(freedv)`
instead of using it directly as part of the modulation function? 
`n_preamble = c_lib.freedv_rawdatapreambletx(freedv, mod_out_preamble);` ? 

Maybe
`freedv_get_n_preamble(freedv)`

drowe67:
Are you asking for a way to send user-defined preamble bits?  It's fairly important that we have a fixed preamble known to the Rx, in order to tune Rx sync algorithms.

DJ2LS:
I'm talking about the size of the preamble for buffer allocation. If there's an own function, we would keep it homogeneous to the other functions for buffer allocation and possibly we avoid problems with python and ctypes.
e.g.
`freedv_get_n_preamble(freedv)` which would return the needed buffer size for the preamble as int.
Actually the size of the preamble will be returned by 
`n_preamble = c_lib.freedv_rawdatapreambletx(freedv, mod_out_preamble);`
Do you know what I mean? Sorry, if I'm talking too unclear

drowemyriota:
Ahh OK.  Sure, that is a good idea.  At the moment I'm making sure the buffer space for the preamble is less than `freedv_get_n_tx_modem_samples` and assuming there is already a buffer allocated for that.

PR #159, opened 2020-12-25
Make H_2064_516 vectors same style as others

srsampson:
This vector was missing the (uint16_t) cast and the style was different, so I added the .c and .h files to duplicate the style of the others.


drowe67:
Hi @srsampson I think that was one of the first codes developed, and was generated by an earlier version of the Octave tool.

Another approach would be to generate the file using the current tool `ldpc_gen_c_h_file.m`, to avoid manual edits on machine generated files.

Also we need to have consistent names between C and Octave versions of the same LDPC code, and make sure the Octave simulations using that LDPC code still work .... not sure if there are ctests for that.

drowe67:
@srsampson and just to complicate things I just discovered some of the Octave data modes simulations that use this code are broken ... better let me fix them over in https://github.com/drowe67/codec2/pull/157 first!

srsampson:
Ah! Good point about generated code.

billcowley:
A good idea to have a use a standard tool to generate the generate the code files as I reckon some of the earlier codes can be improved in the waterfall region, and they can probably all be improved in terms of error floors.   Also @drowe67 what is the next priority for longer code designs - maybe N=4k at rate 3/4?

drowemyriota:
> A good idea to have a use a standard tool to generate the generate the code files as I reckon some of the earlier codes can be improved in the waterfall region, and they can probably all be improved in terms of error floors. Also @drowe67 what is the next priority for longer code designs - maybe N=4k at rate 3/4?

Good question :thinking: I'd like to improve the `datac3` mode (see spreadsheet  in codec2/doc).  My current rule of thumb is a 4 second FEC codeword period to comfortably ride over MPP channels (1 Hz Doppler).  A good target SNR would be 0dB (or a few dB negative) SNR on a MPP/MPD channel.  We might have to do some sums to work out what that means in terms of codeword length, code rate, and throughput bits/s.

Are there any guidelines for selecting the code rate for QPSK?  I recall there were for FSK.

drowe67:
@srsampson I've brought this up to date with master and added a ctest for the `H_2064_516_sparse` code.  So I'm happy.

Anything else you'd like to do before we merge?

srsampson:
No nothing here, merge away! :+1:

PR #161, opened 2020-12-30
cmake: add libraries needed by freedv-gui

ra1nb0w:
This is essentially an issue that should be discussed.
I updated codec2/freedv-gui macports packages to the latest commit and to build freedv-gui I need the following library on /opt/local/include (they are requested by `modem_stats.h` and `.h` in freedv-gui). This patch accomplishes this request but, for example,kiss_fft is an external project and it should not be copied.
What do you think?

PR #162, opened 2020-12-30
reduce header file requirements for freedv-gui

drowe67:
Alternative approach to https://github.com/drowe67/codec2/pull/161

### DIFF
--- a/src/freedv_api.h
+++ b/src/freedv_api.h
@@ -75,6 +75,8 @@
 #define FREEDV_SYNC_AUTO   1                 // falls out of sync automatically
 #define FREEDV_SYNC_MANUAL 2                 // fall out of sync only under operator control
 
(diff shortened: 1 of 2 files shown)


src/freedv_api.h
```diff
@@ -75,6 +75,8 @@
 #define FREEDV_SYNC_AUTO   1                 // falls out of sync automatically
 #define FREEDV_SYNC_MANUAL 2                 // fall out of sync only under operator control
 
+#define FREEDV_VARICODE_MAX_BITS  12         // max bits for each ASCII character
```
tmiw, line 78: Would this be better as a function? That said, I don't foresee this changing any time soon, either.

drowe67: Yes I did consider that, but arrived at the same conclusion.

PR #167, opened 2021-01-04
Use different SNR value to determine whether to open squelch.

tmiw:
@drowe67, this resolves drowe67/freedv-gui#81, where squelch must be set below 0dB for FreeDV to emit any audio for 800XA at all (regardless of the actual SNR).

### DIFF
--- a/src/fdmdv.c
+++ b/src/fdmdv.c
@@ -677,8 +677,8 @@ void generate_pilot_lut(COMP pilot_lut[], COMP *pilot_freq)
 	if (f >= 4)
 	    memcpy(&pilot_lut[M_FAC*(f-4)], pilot, M_FAC*sizeof(COMP));
     }
-   
-    // create complex conjugate since we need this and only this later on 
+
+    // create complex conjugate since we need this and only this later on
     for (f=0;f<4*M_FAC;f++)
     {
         pilot_lut[f] = cconj(pilot_lut[f]);
@@ -841,8 +841,8 @@ float rx_est_freq_offset(struct FDMDV *f, COMP rx_fdm[], int nin, int do_fft)
 	f->pilot_baseband2[j] = cmult(rx_fdm[i], prev_pilot[i]);
     }
 #else
-    // TODO: Maybe a handwritten mult taking advantage of rx_fdm[0] being 
-    // used twice would be faster but this is for sure faster than 
+    // TODO: Maybe a handwritten mult taking advantage of rx_fdm[0] being
+    // used twice would be faster but this is for sure faster than
     // the implementation above in any case.
     arm_cmplx_mult_cmplx_f32(&rx_fdm[0].real,&pilot[0].real,&f->pilot_baseband1[NPILOTBASEBAND-nin].real,nin);
     arm_cmplx_mult_cmplx_f32(&rx_fdm[0].real,&prev_pilot[0].real,&f->pilot_baseband2[NPILOTBASEBAND-nin].real,nin);
@@ -1157,13 +1157,13 @@ void down_convert_and_rx_filter(COMP rx_filt[NC+1][P+1], int Nc, COMP rx_fdm[],
         entire rx_fdm_mem array.  To downconvert these we need the LO
         phase referenced to the start of the rx_fdm_mem array.
 
-      
+
         <--------------- Nrx_filt_mem ------->
         nin
         |--------------------------|---------|
         1                          |
         phase_rx(c)
-     
+
         This means winding phase(c) back from this point
         to ensure phase continuity.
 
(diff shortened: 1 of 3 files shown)


src/freedv_api.c
```diff
@@ -855,7 +855,7 @@ int freedv_bits_to_speech(struct freedv *f, short speech_out[], short demod_in[]
            }
            else {
                /* sync is solid - decode even through fades as there is still some speech info there */
-               if (f->stats.snr_est > f->snr_squelch_thresh)
+               if (f->snr_est > f->snr_squelch_thresh)
```
drowe67, line 858: Thanks @tmiw :+1: 

This same function is used by all voice modes, so we'll need to make sure other modes are not affected. So far we haven't had any reports of similar issues on 1600, 700C/D/E, or 2020. 

OK so it looks like the other modes set `f->snr` like [this example for 700C](https://github.com/drowe67/codec2/blob/master/src/freedv_700.c#L291)

I think `f->snr` is set [here for 800XA](https://github.com/drowe67/codec2/blob/master/src/freedv_fsk.c#L575).
 
BTW the code above and the FSK modem were refactored mid-2020, which might explain why the bug just appeared.  Suspect it's an Eb/No to SNR scaling thing.  I'll look into it some more.

tmiw: @drowe67, I did some brief testing with 700D and squelch still seemed to work, FWIW. However, perhaps we can see where SNR should have gotten copied to f->stats for 800XA and make sure that happens instead?

drowe67: @tmiw IIIRC my general design idea was for all modes to set  `f->snr` during the freedv_rx() call, so the logic in `freedv_bits_to_speech()` can then handle squelch for all modes. The way the SNR is determined could vary across modems (e.g. some modems might not fill the stats structure). `f->stats` isn't actually filled in `freedv_comprx_fsk()` although I think it gets filled later when we grab extended info for the eye diagram - which might be why the initial fix happens to work.

I'll look into the SNR calculation for 800XA.

drowe67: @tmiw Ahh Ok I misread the initial patch above - yes I agree with your fix.  As per the comment above it's what I had intended in the design. Good work :100:

drowe67: @tmiw I'm running a few tests on squelch using the sample 700D file and the built-in Tools-Options noise generator.  However I'm getting some odd results, e.g. when I enable the noise, and press Resync : 700D won't resync properly and the audio stops.  This is odd, as I haven't had any reports of this behavior from off air signals.  Could be a configuration issue on my end.

I've run out of time to work on this today, will try again later this week.

tmiw: @drowe67, FWIW, I see the same issue using full duplex too, not via playing back the sample file (700D and 700E, haven't tried other modes). However, it doesn't seem related to this fix as I still see it even after backing out the change. Not sure what it could be offhand.

tmiw: @drowe67, I noticed a comment in freedv_set_sync() indicating that it needed to be called from the same thread that calls freedv_rx(). I made an update so that pushing the button sets a global variable that's then checked in txRxProcessing() before calling freedv_set_sync() and that seems to have fixed the issue for me. Should I check that into the same bugfixes branch as the scatter plot changes, or is a different branch/PR preferred?

drowemyriota: @tmiw thanks for looking into that.  Lol looks like I should read my own documentation!  Yes I agree it's unrelated to the changes in this PR.  Adding your change to the freedv bug fix PR would be great, thanks :+1:

tmiw: @drowe67, done. See drowe67/freedv-gui#85.

src/fdmdv.c
```diff
@@ -1956,6 +1956,9 @@ void fdmdv_simulate_channel(float *sig_pwr_av, COMP samples[], int nin, float ta
     float sig_pwr, target_snr_linear, noise_pwr, noise_pwr_1Hz, noise_pwr_4000Hz, noise_gain;
     int   i;
 
+    /* prevent NAN when we divide by nin below */
+    if (nin == 0) return;
+
```
drowe67, line 1961: @tmiw :point_up:

tmiw: @drowe67, works for me!

DG9BFC: any copiled version for me to try it out?!?

DG9BFC: in 141 devel 800xa mode did have a differet squelch (no idea who of your clever programmers made it but it worked) ... if you now have a compiled version with new snr calculation or whatever then send me a downloadlink PLEASE ... greetz sigi dg9bfc

drowe67: @DG9BFC no.  Please be patient.  We do this as volunteers and as a service to the Ham radio community.  A new release will be available in good time.

DG9BFC: cant wait to try it out hi hi ;-)
(pulling your leg a bit)
i found the snr thingy and i thought its not that difficult to solve (cant you just use the snr calculation that was in 141 devel??) ... ok ok it seems i have to wait ... sniff ...
if you have something to try out ... drop me a line , send me a note, make a signal fire or whatever you can do please :-)

PR #169, opened 2021-01-08
Add Debian packaging configuration in CMake.

tmiw:
Codec2 equivalent to what was done in drowe67/freedv-gui#87.

@drowe67, this might need a bit more work considering that Debian standard is to have a -dev package for stuff like include files (while this only generates one package for both the .so files and includes). I'm not sure offhand how to do that, though.


tmiw:
@maitbot, yeah, I did notice that LPCNet's CMakeLists.txt fetched a file from the internet during configuration which did cause issues for me in the past. A way to build without doing that (either specifying the path to the needed file(s) on the cmake command line or via some other mechanism) would be good, I think.

@drowe67, I wonder if it would be possible to dynamically generate the files inside lpcnet_191005_v1.0.tgz during the CMake configuration step. Do those change all that often?

drowe67:
> @drowe67, I wonder if it would be possible to dynamically generate the files inside lpcnet_191005_v1.0.tgz during the CMake configuration step. Do those change all that often?

That tarball holds the neural network model (a lot of floats).  It doesn't change often.  It's generated by a couple of days training on a lot of data so not the sort of process someone building the code could easily repeat.

tmiw:
> > @drowe67, I wonder if it would be possible to dynamically generate the files inside lpcnet_191005_v1.0.tgz during the CMake configuration step. Do those change all that often?
> 
> That tarball holds the neural network model (a lot of floats). It doesn't change often. It's generated by a couple of days training on a lot of data so not the sort of process someone building the code could easily repeat.

Hmm, okay, maybe just the results can be included in that case. 

BTW, I updated README.md with instructions on using cpack.

PR #171, opened 2021-01-17
HF Data Acquisition

drowe67:
Tuning and debugging FreeDV raw data mode acquisition (sync) using TNC use cases as a guide:
1. Developed many ctests, and `unittest/tnc*` test scripts that simulate operation over sound cards.
1. Developed a new burst mode acqusittion system based on pre and post ambles.


src/ofdm.c
```diff
@@ -1099,13 +1099,14 @@ static int ofdm_sync_search_core(struct OFDM *ofdm) {
...
+    if (ofdm->verbose > 1) {
         fprintf(stderr, "    ct_est: %4d foff_est: %4.1f timing_valid: %d timing_mx: %5.4f\n",
                 ct_est, (double) ofdm->coarse_foff_est_hz, timing_valid,
                 (double)timing_mx);
     }
 
-    if (timing_valid != 0) {
+    ofdm->timing_valid = timing_valid;
+    if (ofdm->timing_valid != 0) {
         /* potential candidate found .... */
```
drowe67, line 1110: @srsampson :point_up: I think this is a bug in the OFDM modem -  `ofdm->timing_valid` wasn't being cleared when we transitioned from `synced` to `search`, resulting in an immediate transition back to `trial`, even if the acquisition conditions weren't valid.  Not too important at the end of a voice over I guess.

drowe67:
@DJ2LS - I built up a C based test for the first use case I suggested - and found a bug :roll_eyes: Sorry about that - this might have been affecting your tests.

When convenient, pls try the command line above and see if you can run similar tests in your Python code.  Under these test conditions (high SNR) - we should get 100% sync, and receive every packet we send.

drowe67:
@DJ2LS I've made a first pass at adding a preamble to the data modes.  I works OK in moderate amounts of AWGN noise:
```
./src/freedv_data_raw_tx DATAC1 /dev/zero - --testframes 1 --bursts 100 --delay 2000 | ./src/cohpsk_ch - - -22 --Fs 8000 --raw_dir ../raw | ./src/freedv_data_raw_rx DATAC1 - /dev/null --resetsync 1
<snip>
cohpsk_ch: SNR3k(dB):     5.24  C/No....:    40.01
modem bufs processed: 5899  output bytes: 47430 output_frames: 93 
```
So 93/100 single packet bursts detected at 5dB SNR which is OK.  However still quite a bit of work to get this working reliably at lower SNRs (in streaming mode datac1 works down to 1dB).  I need to think more about use cases for a TNC:  

1. For example in this test, the Rx has no idea when the burst will come.  In practice, a receiving terminal will likely have a known time window where it will expect a burst to appear.  

1. The modem can support bursts of any number of frames, but it's unlikely we will transmit more than say 30 seconds at a time.  I think many current ARQ systems just send for about 4 seconds (a one frame burst).

DJ2LS:
And here is a test with data sent and received from Python but without an UNSYNC after every burst!
```
python3 TEST_TX.py --frames 1 --bursts 100 --mode 10 --delay 2000 | ./codec2/build_linux/src/cohpsk_ch - - -22 --Fs 8000 --raw_dir ../raw | python3 TEST_RX.py --mode 10
cohpsk_ch: Fs: 8000 NodB: -22.00 foff: 0.00 Hz fading: 0 nhfdelay: 0 clip: 32767.00 ssbfilt: 1 complexout: 0
cohpsk_ch: SNR3k(dB):     6.64  C/No....:    41.42
cohpsk_ch: peak.....: 22552.78  RMS.....:  9345.54   CPAPR.....:  7.65 
cohpsk_ch: Nsamples.:  3344000  clipped.:     0.00%  OutClipped:  0.00%
end of data
------------------------------
TOTAL RECEIVED BYTES: 51200
RECEIVED FRAMES: 100
```

DJ2LS:
> @DJ2LS - I built up a C based test for the first use case I suggested - and found a bug 🙄 Sorry about that - this might have been affecting your tests.
> 
> When convenient, pls try the command line above and see if you can run similar tests in your Python code. Under these test conditions (high SNR) - we should get 100% sync, and receive every packet we send.

Done!  But I need to improve the UNSYNC part on the receiver side after each burst to be homogeneous with your tests. Actually I don't have any packet loss, while your C test programs indicate a packet loss. 🤔

DJ2LS:
> @DJ2LS I've made a first pass at adding a preamble to the data modes. It works OK in moderate amounts of AWGN noise:

meanwhile I added the UNSYNC state after each burst on the RX side and added the preamble to the TX side for each burst. But for some reason the number of  successfull decoded frames decreased drastically even if I don't use "unsync"  🤔 I think I have to check my test programs tomorrow. Time for a small pause

drowe67:
> meanwhile I added the UNSYNC state after each burst on the RX side and added the preamble to the TX side for each burst. 

That's great progress - well done :+1: .  Given we are adding noise, and have different test frame data, I wouldn't be surprised if we get slightly different results (like a different number of packets received).  At high SNRs (low noise), the results should converge on 100%.  The 3rd parameter of `cohpsk_ch` sets the level of the noise power in dB, smaller is less noise - so "-100" would mean zero noise.

When adding the preamble, did you use the [freedv_rawdatapreambletx](https://github.com/drowe67/codec2/blob/dr-packet/src/freedv_data_raw_tx.c#L240) function? This is now working for the DATA modes. It's currently set to send two "modem frames" of data.  Each modem frame is 0.11s long for the data modes.  We concatenate modem frames to make the entire frame in order to get enough bits for the FEC codeword.  If you listen to the Tx signal, you can hear the preamble at the start.

There's still a lot of tuning needed to get reliable burst operation at low SNRs, I need to think about that.

DJ2LS:
> When adding the preamble, did you use the [freedv_rawdatapreambletx](https://github.com/drowe67/codec2/blob/dr-packet/src/freedv_data_raw_tx.c#L240) function? This is now working for the DATA modes. It's currently set to send two "modem frames" of data. Each modem frame is 0.11s long for the data modes. We concatenate modem frames to make the entire frame in order to get enough bits for the FEC codeword. If you listen to the Tx signal, you can hear the preamble at the start.
>
I solved the problem. It was on my RX side. Sorry

drowe67:
Hi @DJ2LS, I've had an attempt at building up a "service" script to run a "Test 1 - High SNR" over virtual sounds cards.  I used a similar service script approach for some similar VHF data work. 

I'm not sure how useful this will be, as similar tests can be run much faster than real time using pipes.  However I guess it gives the modem a work out over sound cards.

To use it:
```
cd codec2/unittest
./tnc1_high_snr.sh start 3
<wait 15 seconds or so for 3 packets to be sent and received>
cat tnc1.log
```
The script has options to run in the foreground, or the background as a "service", and will prompt you if the loopback ALSA driver is not installed.  I've run 100 packets over it OK (just come back after a coffee).

drowe67:
@DJ2LS - after another few hours of fighting bash, a ping script using two virtual sounds cards and three processes.  Turns out you can pipe the output of `freedv_data_raw_rx` straight into `freedv_data_raw_tx`:
```
./tnc4_high_snr_ping.sh start 10
<find out when we are finished>
./tnc4_high_snr_ping.sh status
<number of frames pinged back to us>
cat tnc4_1.log
```
This starting to get pretty complex, and hard to debug.  We might need t think about a better way of visualizing logs, so we can follow the protocol in action and follow how the two terminals are talking to each other.

drowe67:
@DJ2LS I have added a first pass `DATAC0` mode:
```
./src/freedv_data_raw_tx DATAC0 /dev/zero - --testframes 1 --bursts 100 | ./src/cohpsk_ch - - -24 --raw_dir ../raw --Fs 8000 | ./src/freedv_data_raw_rx DATAC0 - /dev/null --testframes --vv --resetsync 1
modem bufs processed: 1100  output bytes: 1386 output_frames: 99
```
Hopefully that's enough to get started.  As with the other modes, I'll need to improve the single frame burst sync properties to make it more robust.

drowe67:
I've been working on a new acquisition system that uses a pre-amble and post-amble at the start and end of a frame.  It's currently in Octave simulation form. Here are some single frame bursts for AWGN and MPP, both at 5dB SNR:

Screenshot from 2021-02-11 06-52-19
Screenshot from 2021-02-11 06-51-05

You can see what a mess the MPP channel makes.  Sometimes we find the pre-amble, other times the post-amble.  Using both increases the probability of detecting the burst, it's a form of time diversity.  If the probability of missing the pre-amble is P(fail)=0.1, then the probability of missing the pre and post-amble is P(fail)*P(fail)=0.01. If you find either one you can work out where the burst starts and start demodulating.  I've run a 100 burst tests at 5dB/MPP and obtained just one failure (5dB is where this datac1 mode falls over in terms of 10% PER).

@srsampson I'm re-working acquisition to handle the HF data case. It jointly estimates coarse timing and frequency.  It's a bit heavy on CPU but if we target x86s that's OK.

drowe67:
@DJ2LS @srsampson I've been working up some acquisition tests for the data waveforms.  Here are the probability of acquisition results for the MPP HF channel model.

ofdm_dev_acq_curves_datac0_hf
ofdm_dev_acq_curves_datac1_hf

`datac0` looks pretty sold, and `datac1` works well above 0dB where we need it to.  We want `datac0` to be very robust as it's used for ACK packets.

These tests are pretty CPU heavy and took a few hours - made my laptop hot! Each point runs 10 tests (not really enough) and as you can see I'm sweeping over a bunch of SNRs and frequency offsets.  Each burst has a random timing offset too (I add a few 100ms jitter when I generate the packet). I suspect the `datac1` dip at 10dB is a random fail - 10 tests at each point isn't really enough.  I have larger machine I might try where I can run longer tests.

DJ2LS:
@drowe67  Is there a way, I can assist by processing plots for you?

drowe67:
> @drowe67 Is there a way, I can assist by processing plots for you?

Thanks Simon but I think I'll have enough CPU if I move to my larger machine :slightly_smiling_face: Hopefully the C code will run much faster so we won't have a problem with real time operation.

drowe67:
@DJ2LS I've started running some longer acquisition simulations on my server and am getting better results, the plot below shows HF acquisition performance using 25 tests per simulation point.  It looks like the probability of acquisition is >=0.9.  That should be OK for our application.

I'll run some more simulations with more points on different fading channels, then think about how to get this going in C.

 @srsampson FYI the core routine is [here](https://github.com/drowe67/codec2/pull/171/files#diff-cf038c822f9a68e9e8f080c0cea7bb03e0847e70cfb360886b0e575f6c5d0e85R580), it's a more exhaustive version on the acquisition algorithms we have used in the past.  I'm a bit concerned about CPU load but we shall see if it's a problem in the real world.  The target platforms are x86 so plenty of CPU.

ofdm_dev_acq_curves_datac1_hf

srsampson:
I'll take a look. I like the idea of a preamble and postamble, although I thought maybe they were old technology for a bit. I guess dealing with pilots made me think that, but they certainly would give you fast acquisition, rather than scanning two frames for the start. oops, I'm blogging, ha.

drowe67:
> I'll take a look. I like the idea of a preamble and postamble, although I thought maybe they were old technology for a bit. I guess dealing with pilots made me think that, but they certainly would give you fast acquisition, rather than scanning two frames for the start. oops, I'm blogging, ha.

Hi @srsampson - yes I think we need the post/pre-amble for reliable single frame sync in the HF data use case.  Our pilot based system (essentially a small, repeating pre-amble) works well for the streaming data case, where we don't care if we lose the first few frames of voice.  Also data is not latency sensitive, so buffering several seconds until you find the post-amble is OK.

I pleased with how reliable the acquistion simulations are even a negative SNRs.

drowe67:
Latest and greatest P(acquisition) curves for the three waveforms.  I tweaked datac3 as it's initial performance wasn't so great.  They are all P(acquisition) > 0.9 for the range of SNRs we are interested in.  These are the results for the Octave simulation, some more work required to see if we get similar results when integrated into the C modem.

ofdm_dev_acq_curves_datac0_mpp_acq
ofdm_dev_acq_curves_datac1_mpp_acq
ofdm_dev_acq_curves_datac3_mpp_acq

drowe67:
@DJ2LS here's a first attempt at a new state machine/programming model.  Now you set the number of frames per burst (which we call N in our Google doc) using the new `freedv_set_frames_per_burst` API call, and the state machine will reset itself when N frames have been received.  So we don't need `freedv_set_sync`.  The default (frames per burst 0) is the previous streaming sync state (state machine requires manual reset), and you can still use `freedv_set_sync` if you want. See `freedv_data_raw_rx.c` for a programming example.

For example with N=1 frame per burst:
```
./freedv_data_raw_tx DATAC0 /dev/zero - --testframes 1 --bursts 4 | ./freedv_data_raw_rx DATAC0 - /dev/null --testframes --vv --framesperburst 1
```
with N=2:
```
./freedv_data_raw_tx DATAC0 /dev/zero - --testframes 2 --bursts 4 | ./freedv_data_raw_rx DATAC0 - /dev/null --testframes --vv --framesperburst 2
```
Or 100 frames with some AWGN noise:
```
/src/freedv_data_raw_tx DATAC0 /dev/zero - --testframes 1 --bursts 100 | ./src/cohpsk_ch - - -24 --raw_dir ../raw --Fs 8000 | ./src/freedv_data_raw_rx DATAC0 - /dev/null --testframes --vv --framesperburst 1
```
It's even doing OK on fading channels:
```
./freedv_data_raw_tx DATAC1 /dev/zero - --testframes 1 --bursts 100 | ./cohpsk_ch - - -24 --mpp  --raw_dir ../raw --Fs 8000 | ./freedv_data_raw_rx DATAC1 - /dev/null --testframes --vv --framesperburst 1

```
Please tell me if you feel this programming model is better for you, and helps with the lock up situations you have been experiencing.  My TNC1 and TNC4 unit test scripts seem to work OK with this model.

DJ2LS:
@drowe67 , thanks a lot, I will test the function tomorrow.
What happens, if we are loosing a frame during a burst? Lets say, we are sending a burst with N = 5 frames and we are loosing only frame no 4. Will there an unsync after frame no 5 or do we have to trigger an unsync manually in this case?

drowe67:
@DJ2LS good question! If frame 4 is lost (in the sense it doesn't decode correctly, a CRC fail, and no bytes are returned), it will still reset the state machine to the search state after frame 5.  Basically the state machine will always be reset to the search state after N frames.

DJ2LS:
@drowe67 I did some ARQ tests with `freedv_set_frames_per_burst`, today. In my case I called `freedv_set_frames_per_burst` after the first received frame, because I'm getting the value for N frames per burst from the ARQ header to keep the RX side dynamical.  This seems to be working as well - instead of placing it at the beginning like in your programming example. 

But as soon as I have to repeat a frame again via the ARQ protocol, because it got lost for example, I have to handle the sync state manually, again. 
In this case an unsync , triggered by a **postamble** would help a lot! 
A general unsync, triggered by the postamble, could keep the RX side more flexible.

DJ2LS:
@drowe67 I implemented the function into my test programs. 
For the test with HIGH SNR everything seems to be fine and I have a "pass" as well.

The other test programs with piping through cohpsk_ch need to be optimized by me. There I have some bugs which are affecting my Tests and with the current state I have only a "partial pass"

drowe67:
I've been building up an acquisition algorithm that takes blocks of `nin` samples frame by frame (like the real modem does), and has a small state machine to find the peaks when looking for the post and preamble, and includes thresholds.  So I'm getting closer to the real world algorithm we need for the modem.  I also found a bug in setting the SNR, I needed to account for the on/off time of the burst, unlike the previous streaming work where the Tx signal is present all of the time.

So here are the latest probability of acquisition curves for the three modes.  I didn't sweep across frequency as the previous simulations suggested performance was uniform across frequency offsets (so the frequency offset is fixed at 0Hz).  There were 50 trials per simulation point.  

ofdm_dev_acq_curves_fbf_datac0
ofdm_dev_acq_curves_fbf_datac1
ofdm_dev_acq_curves_fbf_datac3

Some notes and observations:
1. The "operating point" of `datac1` is 5dB on a MPP channel, and `datac3` 0dB.  That's where those waveforms get 10% PER (assuming they have acquired OK).  So we want acquisition to work OK a bit beneath those SNRs, just to be sure. `datac0` is a small frame used for a ARQ and signalling, so we want that to work well at 0dB, so it can be used as the back channel for `datac3` as well as faster/higher SNR rates like `datac1`.
2. `datac3` is a little better than `datac0`, probably because the pre and post amble are further separated in time. This is slightly unfortunate as `datac0` is used for the ARQs, so we want that to be super reliable.  We'll just have to make sure that any protocol is robust to lost ARQ packets.
3. The notch channel model has a 10dB notch in the spectrum every 500Hz (2ms delay).  Unlike the MPM/MPP channels the fading is not time varying, but stationary. I included this as previous versions of 700D couldn't sync on a very slow fading channel that kept one part of the spectrum notched out.  Here's a spectrogram of the `datac1` signal with the notch channel model:
   
   Screenshot from 2021-02-26 08-20-05
   
## Next steps ##

I'm taking this slow and careful, as I have been burnt many times by acquisition.  Next steps:
1. Use a unique sequence for the post-amble, different to the pre-amble.  Currently they are the same, as that was easier as a first step.  Re-run the simulation after those changes and debug.
2. Work out how the acquisition system passes time references to the demod, e.g. "the start of the frame is 1056 samples in the past - start demodulating there".  If we detect a post-amble, we may have to tell the demod to start 4 seconds in the past!
3. Work up to the point where I can run the Octave version of the demod with the new acquisition algorithm.  I might start with just the pre-amble.  That will be reduced performance, but much easier to get going and will give us something to play with early.

drowe67:
OK, I now have separate pre and post-amble sequences, and have worked on the frame by frame version of the new acquisition algorithm a little more:

2021-02-27_ofdm_dev_acq_curves_fbf_datac0
2021-02-27_ofdm_dev_acq_curves_fbf_datac1
2021-02-27_ofdm_dev_acq_curves_fbf_datac3

I'm pretty happy with these results, so now I will work on integration with the Octave version of the OFDM modem.  

@srsampson - one major challenge is how to handle the case when the burst is detected via the post-amble.  In this case we need to backtrack 4 seconds to the start of the packet.  I'm not sure how to handle that in the `rxbuf` part of the modem design.

drowe67:
@srsampson So I spent a day trying to figure out that "black magic" timing loop code.  That I wrote a few years ago :grin: I guess we'd always been a bit unsure what was going on.  But I think I understand it now - it's keeping `sample_point` at `timing_est+Ncp`, using either the upper or lower limit of `timing_est` that bounces around due to multipath delays.  It made more sense when I "broke the loop" and plotted the behaivour with a sample clock offset:

Screenshot from 2021-03-06 07-15-36

In this case `timing_est` is gradually slipping backwards, but the algorithm is keeping `sample_point` at `timing_est+Ncp`.  It also does great even with really nasty channels.  Here is the scatter plot for the MPD channel (4ms delay spread) on 700E.  There's no additive noise to make ISI more obvious:

Screenshot from 2021-03-06 07-11-51

drowe67:
@DJ2LS :point_up: this work is the starting point for getting the postamble integrated into the modem.  When a postamble is detected we need a way to "rewind" the past samples that have run through the modem to get back to the start of the burst.  This is all tied up with the timing system that decides where each QPSK symbol is.

octave/ofdm_lib.m
```diff
@@ -722,20 +795,31 @@ This is an alternative algorithm to est_freq_offset() above that is less noisey
...
-      delta_t = ft_est - ceil(ftwindow_width/2);
-      sample_point = max(timing_est+4, sample_point);
-      sample_point = min(timing_est+Ncp-4, sample_point);
+      % Track the ideal sampling point, which is Ncp for a multipath signal whose delay varies between 0 and Ncp.  The
+      % timing est will be bouncing back and forth due to multipath so we may need to use the upper or lower limit of
+      % the timing est to track the ideal sample_point. A good way to explore this algorithm is to disable the feedback
+      % loop for nin adjustment below, and look at the plots from ofdm_rx with +ve and -ve sample clock offsets 
+      % (sox can be used to resample).  The "4" constants are small guard bands so we don't stumble outside of the CP 
+      % due to noise.
+
```
drowe67, line 819: @srsampson :point_up: timing loop logic I am discussing :point_up:

drowe67:
@srsampson would appreciate your thoughts on https://github.com/drowe67/codec2/pull/171/commits/35e86898fe5c32bc8137637cb3941f3081af73a1, e.g. variable names and implementation. I've used an arbitrary `nrxbufhist=1000` just to "ctest" the idea - it's passing OK.  In practice (for a 4 second packet), this might be 4*8000 samples long. 

These results suggest the modem is happy, and just operating at the end of the enlarged `rxbuf`, and -  importantly - still updating/shifting the entire buffer so the earlier samples are still available.  Note that `rxbufst` points to the nominal position of the current modem frame we want to demodulate.

The idea is that if we detect a post-amble during acquisition, we can then set `rxbufst` to our estimate of the start of the packet, and run the demod from there.  This implies it will not need any new samples (`nin` == 0) for many calls of `ofdm_demod()`, as the demod will be chewing through the existing samples in `rxbuf`.

srsampson:
> @srsampson would appreciate your thoughts on [35e8689](https://github.com/drowe67/codec2/commit/35e86898fe5c32bc8137637cb3941f3081af73a1), e.g. variable names and implementation. I've used an arbitrary `nrxbufhist=1000` just to "ctest" the idea - it's passing OK. In practice (for a 4 second packet), this might be 4*8000 samples long.
> 
> These results suggest the modem is happy, and just operating at the end of the enlarged `rxbuf`, and - importantly - still updating/shifting the entire buffer so the earlier samples are still available. Note that `rxbufst` points to the nominal position of the current modem frame we want to demodulate.
> 
> The idea is that if we detect a post-amble during acquisition, we can then set `rxbufst` to our estimate of the start of the packet, and run the demod from there. This implies it will not need any new samples (`nin` == 0) for many calls of `ofdm_demod()`, as the demod will be chewing through the existing samples in `rxbuf`.

Whoo boy! I'll take a look, but I'm afraid you are about a light-year ahead of me now. Hee.

drowe67:
I have a first pass running with post-amble detection, and "rewind" of previously buffered samples.  First generate 1 burst that is one packet long.  Then run the Rx in burst mode, the final argument tells it to expect bursts with one packet:
```
octave:10> ofdm_tx("test_datac0.raw","datac0",1,100,"awgn","bursts",1)
octave:10> ofdm_rx("test_datac0.raw","datac0",0,"packetsperburst",1)
f:   8 nin:  880 st: search  ct_est:   -2 mx: 0.00 coarse_foff: -53.0 timing_valid: 0 post
f:   9 nin:  880 st: search  ct_est:   -2 mx: 0.00 coarse_foff: -53.0 timing_valid: 0 post
f:  10 nin:  880 st: search  ct_est:  882 mx: 0.05 coarse_foff:  27.0 timing_valid: 0 post
f:  11 nin:  880 st: search  ct_est:  837 mx: 0.05 coarse_foff: -23.0 timing_valid: 0 post
f:  12 nin:  880 st: search  ct_est:  372 mx: 0.06 coarse_foff: -35.0 timing_valid: 0 post
f:  13 nin:  880 st: search  ct_est:  605 mx: 0.08 coarse_foff:   0.0 timing_valid: 0 post
f:  14 nin:  880 st: search  ct_est:   94 mx: 0.07 coarse_foff:   3.0 timing_valid: 0 post
f:  15 nin:  880 st: search  ct_est:  277 mx: 0.06 coarse_foff:  -7.0 timing_valid: 0 post
f:  16 nin:  880 st: search  ct_est:  432 mx: 0.49 coarse_foff:   0.0 timing_valid: 1 post
f:  17 nin:    0 st: trial   euw:  10 1 mf:  0 pbw: h eraw:   0 foff: -0.0
f:  18 nin:    0 st: trial   euw:  18 2 mf:  0 pbw: h eraw:   0 foff: -0.0
f:  19 nin:    0 st: trial   euw:   0 3 mf:  3 pbw: h eraw:   0 foff: -0.0
f:  20 nin:    0 st: synced  euw:  14 3 mf:  0 pbw: h eraw:   0 foff: -0.2
f:  21 nin:  880 st: post   
f:  22 nin:  880 st: search  ct_est:    0 mx: 0.00 coarse_foff: -53.0 timing_valid: 0 post
f:  23 nin:  880 st: search  ct_est:   -2 mx: 0.00 coarse_foff: -53.0 timing_valid: 0 post
```
 At `f=10` the demod encounters the preamble frame and the data frames, but I have the preamble detector disabled.  You can see the timing metric `mx` bouncing about but it doesn't get high enough to trigger `timing_valid`.  At `f=16` we find the postamble, we then rewind the internal pointer in the rx buffer and start demodulating frames.  The cool bit: as we are using old samples `nin=0` and so we are not asking for any new samples!  This is the `datac0` mode so we have 4 frames for the packet. At `f=21` we jump to a `post` state to make sure we skip over the postamble and don't get into a loop.  We have then consumed all of the old samples in the rx buffer and start asking for input samples again `nin=880`.
 
This is very much a first pass .... time for some clean up.

The ultimate goal is to improve P(acquisition) in fading channels, still a long way from demonstrating that in the integrated modem.  But this rewind "time travel" part is the toughest coding step :slightly_smiling_face:

drowe67:
@DJ2LS For historical reasons there is some confusing terms that change as we cross from the internal layers to the FreeDV API: 

| FreeDV API | OFDM modem |
| --- | --- | 
| modem frame | frame |
| frame | packet |

In general the FreeDV API user doesn't need to worry about the OFDM internals, they can just think in terms of API level "frames".  

At some point I guess this should be cleared up, but it would be a lot of work and I'd rather write new code :slightly_smiling_face:

drowe67:
@srsampson I've been doing a little refactoring of the Octave code before pushing fwd into the C port of the new burst acquisition.  Our modem design seems to be supporting all these extensions quite well :heavy_check_mark:

drowe67:
@DJ2LS I'm not sure how well the latest commit https://github.com/drowe67/codec2/pull/171/commits/95f5c8197d5f6ced70ad84bd9627306a0ba01826 will work for you as I'm just part way through coding and testing the new preamble detection code in C.  I've swapped to the new preamble, but am still using the previous streaming state machine. This version may not work well for your tests, so you may want to keep an earlier version until the new preamble work matures.

drowe67:
Hmm, mystery fail in one the ctests, doesn't happen on my laptop :thinking: 

@DJ2LS I now have the core burst acquisition functions ported to C and unit tested :smile: Next step is to integrate these routines with the rest of the C data modem.  As I mentioned earlier, I'm building up C support for just the preamble detection initially, will then get the post amble detection working in C as a next step.  The idea is to get you a C version of the new acquisition system/state machine/programming model early, then improve the low SNR performance as a next step.

I've found a few small design issues along the way.  Now I have working tests I will do a little refactoring. 

@srsampson I haven't coded the new burst acquisition stuff for efficiency, `est_timing_and_freq` is computing the sample freq vectors every time.  However the target platform is a laptop so we should be OK.

drowe67:
Hmm still with the mystery ctest fail - some sort of subtle C bug judging by the random output.  Hopefully it will turn up on my machine soon.

@srsampson just found a subtle bug with the burst data modem. It has the familiar frame format of pilots and data symbols:
``` PDDDDPDDDDPDDD```
However I found I was getting rubbish on the last frame, especially for fading channels.  Turns out I needed a final set of pilots at the end:
``` PDDDDPDDDDPDDDP```
The phase estimation on the final frame was getting lost with no final pilot symbol.  Sort of bug we wouldn't find in streaming mode, as there is always another frame, and we use the pilots at the start of frame `n+1` as the pilots at the end of frame `n`.

That's half of my Friday I'll never get back .....

**Comment by drowe67** [comment]
@DJ2LS we now have the new burst acquisition system running the first few tests at the FreeDV API level, for example 3 bursts, each burst 2 frames long:

```
./src/freedv_data_raw_tx --testframes 2 --bursts 3 DATAC1 /dev/zero - | ./src/freedv_data_raw_rx --framesperburst 2 --testframes DATAC1 - /dev/null --vv
<snip>
BER......: 0.0182 Tbits: 12888 Terrs:   234
Coded BER: 0.0000 Tbits:  6144 Terrs:     0
Coded PER: 0.0000 Tpkts:     6 Tpers:     0
```

The raw BER of 0.0182 is a bug (there are no bit errors). Next step is some clean up, tests with channel noise, and tests against your use cases (maybe our TNC use case scripts).

**Comment by DJ2LS** [comment]
@drowe67 I will start testing the new acquisition system with python. Because of some minor problems with python, sample rate conversion and virtual audio devices, I will try to test via real audio devices, the next days and see, if I can improve audio handling for my test programs and TNC as well

**Comment by drowe67** [comment]
Much to my surprise postamble detection is now working C as well. That was easier than I thought :slightly_smiling_face: The PERs were a little higher than expected, so I tuned the Unique Word detection thresholds and am now getting 10% PER at 5dB on datac1.  Getting some seg faults on datac1 and datac3 at 0dB SNR, will fix that soon.

**Comment by drowe67** [comment]
Pausing for a moment to log some test results.  I'm using the Octave Tx as it has nice integration with the channel models, and the C `ofdm_demod` to test the core modem, but results should be similar for `freedv_rx`
```
octave:83> ofdm_ldpc_tx('test_datac0.raw','datac0',1,5,'mpp','bursts',100)
at ../octave/test_datac0.raw | ./src/ofdm_demod --mode datac0 --out /dev/null  --testframes --ldpc --verbose 2 --packetsperburst 1
```

Results:

| Mode | Channel | SNR (dB) | Packets Tx | Packets Rx | Use Case |
|---|---|---|---|---|---|
| datac0 | MPP | 5 | 100 | 92 | reverse link ACK packets when used with datac1 @ 5dB |
| datac0 | MPP | 0 | 100 | 70 | reverse link ACK packets when used with datac3 @ 0dB |
| datac1 | MPP | 5 | 100 | 89 | forward link at 5dB |
| datac1 | MPM | 5 | 100 | 76 | forward link at 5dB |
| datac3 | MPP | 0 | 100 | 74 | forward link at 0dB |

Notes:

1. All tests one packet per burst, MPP channel.
2. `datac0` has a fairly short 440ms duration, which means it can't ride over fades as well as `datac1` and `datac3`.  The failures where DEC decode, the packets were acquired OK. When using `datac0` for ACK information, we may need a mechanism to repeat ACK information across several packets spread in time to provide a reliable reverse channel.
3. `datac3` struggling at 0dB SNR.  While it's duration is long it's a fairly narrow 9 carrier waveform so perhaps gets hit harder than `datac1` by frequency selective fading.  We can't make it wider as the synchronization starts to fall over.  `datac3` was getting 50% throughput at -2dB.  We could consider this a low SNR mode designed to keep some sort of connection up, even at low SNRs.
4. Note `datac1` was a bit worse on the "easier" multipath moderate (MPM) channel.  One theory is the slower fading means the fades can last longer wrt to the FEC codeword length.  Could also be a bug/tuning thing.

**Comment by drowe67** [comment]
I've tested over the air using paths 0.5, 800,000, and 2.600,000 meters long :grinning: I started with two radios across my bench (IC7200 on min power, FT-817 with RF gain wound down), then started Tx-ing at higher power (maybe 20W) using kiwisdrs for the Rx station, recording the file from them and resampling:
```
sox ~/Downloads/sdr-amradioantennas.com_2021-04-09T06_03_59Z_7177.00_lsb.wav -r 8000 -t raw - | ./src/freedv_data_raw_rx datac1 --framesperburst 1 --testframes - /dev/null --vv
```

Couple of test files attached, you can demodulate them with:
```
./src/freedv_data_raw_rx datac0 --framesperburst 1 --testframes test_datac0_001.raw /dev/null --vv
<snip>
BER......: 0.0117 Tbits:  2560 Terrs:    30
Coded BER: 0.0000 Tbits:  1280 Terrs:     0
Coded PER: 0.0000 Tpkts:    10 Tpers:     0
./src/freedv_data_raw_rx datac1 --framesperburst 1 --testframes test_datac1_006.raw /dev/null --vv
<snip>
BER......: 0.0134 Tbits: 73728 Terrs:   986
Coded BER: 0.0000 Tbits: 36864 Terrs:     0
Coded PER: 0.0000 Tpkts:     9 Tpers:     0
```
Screenshot from 2021-04-09 15-54-34
Screenshot from 2021-04-09 15-58-21

[test_files.zip](https://github.com/drowe67/codec2/files/6283532/test_files.zip)

Very pleased that it's not falling apart on real world channels.  Couple of take aways:
1. We need a GUI for freedv_data_rx, e.g. waterfall and scatter diagram, indication of modem sync
1. Impulses can destroy packets, and slam AGC down.  I don't simulate this sort of thing
1. I can see this wisdom in fall back rates, it's annnoying when you can see/hear but not decode. `datac0` punched through a few times when `datac1` didn't.

**Comment by drowe67** [comment]
@DJ2LS @srsampson I might merge this now, as it's a pretty big PR.  Will attend to any bug fixes and improvements in a new PR.

**Comment by drowe67** [comment]
@tmiw :point_up: lot of changes in this PR - just letting you know in case you find any freedv-gui strangeness....

**Comment by tmiw** [comment]
@drowe67, I'm getting new/different unit test failures now:

```
HT-TM05:build_osx mooneer$ ctest -V -R test_OFDM_modem_burst_acq_port\|test_OFDM_modem_700E_AWGN\|test_freedv_api_700D_speech
UpdateCTestConfiguration  from :/Users/mooneer/devel/freedv-gui/codec2/build_osx/DartConfiguration.tcl
Parse Config file:/Users/mooneer/devel/freedv-gui/codec2/build_osx/DartConfiguration.tcl
UpdateCTestConfiguration  from :/Users/mooneer/devel/freedv-gui/codec2/build_osx/DartConfiguration.tcl
Parse Config file:/Users/mooneer/devel/freedv-gui/codec2/build_osx/DartConfiguration.tcl
Constructing a list of tests
test 17
    Start 17: test_OFDM_modem_burst_acq_port

17: Test command: /bin/sh "-c" "cd /Users/mooneer/devel/freedv-gui/codec2/octave;
                            echo "tofdm_acq; quit" | DISPLAY="" octave-cli"
17: Test timeout computed to be: 1500
17: PPPPPPPPPPP
17:  .....U... 
17:  .U....U.. 
17:  ..U....U. 
17:  ...U....U 
17: PPPPPPPPPPP
17:  ....U.... 
17:  U....U... 
17:  .U....U.. 
17:  ..U....U. 
17: PPPPPPPPPPP
17:  ...U....U 
17:  ......... 
17:  ......... 
17:  ......... 
17: PPPPPPPPPPP
17:  ......... 
17:  ......... 
17:  ......... 
17:  ......... 
17: Nc=9 Ts=0.016 Tcp=0.006 Ns: 5 Np: 4
17: Nsymperframe: 45 Nbitsperpacket: 288 Nsamperframe: 880 Ntxtbits: 0 Nuwbits: 32 Nuwframes: 3
17: uncoded bits/s: 654.5
17: SNR3kdB: 10.00 Burst offset: -4.01 SNRdB_setpoint: 5.99
17: Npackets: 1  Nbursts: 1  Peak: 13398.44 RMS: 3644.90 CPAPR: 11.3 clipped:  0.00%
17: 
17: Running Octave version....
17:   0   ct_est:  834 nin:  880 mx: 0.01 foff_est:  -4.0 timing_valid: 0 post
17:   1   ct_est:  650 nin:  880 mx: 0.01 foff_est: -13.0 timing_valid: 0 post
17:   2   ct_est:  437 nin:  880 mx: 0.01 foff_est:  37.0 timing_valid: 0 post
17:   3   ct_est:  272 nin:  880 mx: 0.01 foff_est: -23.0 timing_valid: 0 post
17:   4   ct_est:  529 nin:  880 mx: 0.01 foff_est:  -3.0 timing_valid: 0 post
17:   5   ct_est:  492 nin:  880 mx: 0.01 foff_est:  -4.0 timing_valid: 0  pre
17:   6   ct_est:  195 nin:  880 mx: 0.01 foff_est:  45.0 timing_valid: 0 post
17:   7   ct_est:  179 nin:  880 mx: 0.01 foff_est:  26.0 timing_valid: 0 post
17:   8   ct_est:  438 nin:  880 mx: 0.01 foff_est: -16.0 timing_valid: 0  pre
17:   9   ct_est:  874 nin:  880 mx: 0.04 foff_est:  43.0 timing_valid: 0 post
17:  10   ct_est:  432 nin: 1311 mx: 0.44 foff_est:   0.0 timing_valid: 1  pre
17:  11   ct_est:  840 nin:  880 mx: 0.07 foff_est: -33.0 timing_valid: 0 post
17:  12   ct_est:  431 nin:  880 mx: 0.08 foff_est:   8.0 timing_valid: 0 post
17:  13   ct_est:  431 nin:  880 mx: 0.08 foff_est:   1.0 timing_valid: 0  pre
17:  14   ct_est:  431 nin:  880 mx: 0.09 foff_est:   8.0 timing_valid: 0  pre
17:  15   ct_est:  432 nin:    0 mx: 0.44 foff_est:   0.0 timing_valid: 1 post
17:  16   ct_est:   17 nin:  880 mx: 0.04 foff_est: -36.0 timing_valid: 0 post
17:  17   ct_est:  669 nin:  880 mx: 0.01 foff_est:  41.0 timing_valid: 0  pre
17:  18   ct_est:  644 nin:  880 mx: 0.01 foff_est:  28.0 timing_valid: 0 post
17:  19   ct_est:  226 nin:  880 mx: 0.01 foff_est:  -9.0 timing_valid: 0 post
17:  20   ct_est:  281 nin:  880 mx: 0.01 foff_est: -12.0 timing_valid: 0  pre
17:  21   ct_est:  570 nin:  880 mx: 0.01 foff_est:   8.0 timing_valid: 0  pre
17:  22   ct_est:   62 nin:  880 mx: 0.01 foff_est: -36.0 timing_valid: 0  pre
17:  23   ct_est:  260 nin:  880 mx: 0.02 foff_est:  44.0 timing_valid: 0  pre
17:  24   ct_est:   97 nin:  880 mx: 0.01 foff_est:  -1.0 timing_valid: 0  pre
17:  25   ct_est:  680 nin:  880 mx: 0.01 foff_est: -29.0 timing_valid: 0  pre
17:  26   ct_est:  431 nin:  880 mx: 0.01 foff_est:  25.0 timing_valid: 0 post
17:  27   ct_est:  272 nin:  880 mx: 0.01 foff_est:  41.0 timing_valid: 0  pre
17: 
17: Running C version....
17: path_to_unittest = ../build_linux/unittest
17: sh: ../build_linux/unittest/tofdm_acq: No such file or directory
17: error: load: unable to find file tofdm_acq_out.txt
17: error: called from
17:     tofdm_acq at line 52 column 1
1/3 Test #17: test_OFDM_modem_burst_acq_port ...***Failed  Required regular expression not found. Regex=[PASS
]  1.47 sec
test 29
    Start 29: test_OFDM_modem_700E_AWGN

29: Test command: /bin/sh "-c" "cd /Users/mooneer/devel/freedv-gui/codec2/build_osx/src;
                            ./ofdm_mod --in /dev/zero --testframes 10 --ldpc --mode 700E |
                            ./cohpsk_ch - - -20.5 --Fs 8000 |
                            ./ofdm_demod --mode 700E --ldpc --testframes -v 2 > /dev/null"
29: Test timeout computed to be: 1500
29: cohpsk_ch: Fs: 8000 NodB: -20.50 foff: 0.00 Hz fading: 0 nhfdelay: 0 clip: 32767.00 ssbfilt: 1 complexout: 0
29: using: HRA_56_56
29: LDPC codeword data bits = 56
29: LDPC codeword total bits  = 112
29: LDPC codeword data bits used = 56
29: LDPC codeword total length in modem packet = 112
29: Warning EsNo: 3.000000 hard coded
29:   0 nin:  640 st: search     ct_est:  498 foff_est: -3.0 timing_valid: 0 timing_mx: 0.4977
29:   1 nin:  640 st: search     ct_est:  498 foff_est: -4.0 timing_valid: 1 timing_mx: 0.5861
29:   2 nin:  498 st: trial  euw:  3 0 mf:  0 f:  -4.1 pbw: 1 eraw:   0 ecdd:  14 iter: 100 pcc:  29
29:   3 nin:  640 st: synced euw:  2 0 mf:  0 f:  -4.0 pbw: 1 eraw:   0 ecdd:  11 iter: 100 pcc:  37
29:   4 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.9 pbw: 1 eraw:   0 ecdd:   9 iter: 100 pcc:  34
29:   5 nin:  640 st: synced euw:  4 1 mf:  0 f:  -3.8 pbw: 1 eraw:   0 ecdd:   9 iter: 100 pcc:  27
29:   6 nin:  640 st: synced euw:  2 0 mf:  0 f:  -3.9 pbw: 1 eraw:   0 ecdd:  14 iter: 100 pcc:  33
29:   7 nin:  640 st: synced euw:  0 0 mf:  0 f:  -3.8 pbw: 1 eraw:   0 ecdd:  13 iter: 100 pcc:  26
29:   8 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.9 pbw: 1 eraw:   0 ecdd:   6 iter: 100 pcc:  34
29:   9 nin:  640 st: synced euw:  2 0 mf:  0 f:  -3.8 pbw: 1 eraw:   0 ecdd:   6 iter: 100 pcc:  30
29:  10 nin:  640 st: synced euw:  3 0 mf:  0 f:  -3.7 pbw: 1 eraw:   0 ecdd:  14 iter: 100 pcc:  43
29:  11 nin:  640 st: synced euw:  3 0 mf:  0 f:  -3.6 pbw: 1 eraw:   0 ecdd:  11 iter: 100 pcc:  40
29:  12 nin:  640 st: synced euw:  0 0 mf:  0 f:  -3.5 pbw: 1 eraw:   0 ecdd:  12 iter: 100 pcc:  31
29:  13 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.4 pbw: 1 eraw:   0 ecdd:   7 iter: 100 pcc:  46
29:  14 nin:  640 st: synced euw:  0 0 mf:  0 f:  -3.3 pbw: 1 eraw:   0 ecdd:   6 iter: 100 pcc:  30
29:  15 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29:  16 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.3 pbw: 1 eraw:   0 ecdd:  11 iter: 100 pcc:  31
29:  17 nin:  640 st: synced euw:  3 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:   3 iter: 100 pcc:  52
29:  18 nin:  640 st: synced euw:  0 0 mf:  0 f:  -3.1 pbw: 1 eraw:   0 ecdd:  10 iter: 100 pcc:  33
29:  19 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  49
29:  20 nin:  640 st: synced euw:  2 0 mf:  0 f:  -3.3 pbw: 1 eraw:   0 ecdd:   7 iter: 100 pcc:  31
29:  21 nin:  640 st: synced euw:  2 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:  10 iter: 100 pcc:  26
29:  22 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:   0 iter:  14 pcc:  56
29:  23 nin:  640 st: synced euw:  3 0 mf:  0 f:  -3.1 pbw: 1 eraw:   0 ecdd:   0 iter:  11 pcc:  56
29:  24 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:   6 iter: 100 pcc:  34
29:  25 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.3 pbw: 1 eraw:   0 ecdd:  10 iter: 100 pcc:  20
29:  26 nin:  640 st: synced euw:  0 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:  11 iter: 100 pcc:  49
29:  27 nin:  640 st: synced euw:  0 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  55
29:  28 nin:  640 st: synced euw:  0 0 mf:  0 f:  -3.1 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  52
29:  29 nin:  640 st: synced euw:  0 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:   0 iter:  14 pcc:  56
29:  30 nin:  640 st: synced euw:  0 0 mf:  0 f:  -3.3 pbw: 1 eraw:   0 ecdd:   5 iter: 100 pcc:  41
29:  31 nin:  640 st: synced euw:  1 0 mf:  0 f:  -3.2 pbw: 1 eraw:   0 ecdd:   8 iter: 100 pcc:  35
29:  32 nin:  640 st: synced euw:  4 1 mf:  0 f:  -3.1 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  54
29:  33 nin:  640 st: synced euw:  3 0 mf:  0 f:  -3.0 pbw: 1 eraw:   0 ecdd:  10 iter: 100 pcc:  28
29:  34 nin:  640 st: synced euw:  1 0 mf:  0 f:  -2.9 pbw: 1 eraw:   0 ecdd:  13 iter: 100 pcc:  25
29:  35 nin:  640 st: synced euw:  1 0 mf:  0 f:  -2.8 pbw: 1 eraw:   0 ecdd:   0 iter:   8 pcc:  56
29:  36 nin:  640 st: synced euw:  2 0 mf:  0 f:  -2.7 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  54
29:  37 nin:  640 st: synced euw:  0 0 mf:  0 f:  -2.6 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29:  38 nin:  640 st: synced euw:  1 0 mf:  0 f:  -2.6 pbw: 1 eraw:   0 ecdd:   3 iter: 100 pcc:  42
29:  39 nin:  640 st: synced euw:  3 0 mf:  0 f:  -2.5 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  40 nin:  640 st: synced euw:  1 0 mf:  0 f:  -2.4 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  41 nin:  640 st: synced euw:  3 0 mf:  0 f:  -2.3 pbw: 1 eraw:   0 ecdd:   1 iter: 100 pcc:  53
29:  42 nin:  640 st: synced euw:  1 0 mf:  0 f:  -2.2 pbw: 1 eraw:   0 ecdd:   0 iter:  12 pcc:  56
29:  43 nin:  640 st: synced euw:  2 0 mf:  0 f:  -2.1 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  44 nin:  640 st: synced euw:  3 0 mf:  0 f:  -2.0 pbw: 1 eraw:   0 ecdd:   0 iter:   8 pcc:  56
29:  45 nin:  640 st: synced euw:  0 0 mf:  0 f:  -2.0 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29:  46 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.9 pbw: 1 eraw:   0 ecdd:   0 iter:   7 pcc:  56
29:  47 nin:  640 st: synced euw:  2 0 mf:  0 f:  -1.8 pbw: 1 eraw:   0 ecdd:   0 iter:   8 pcc:  56
29:  48 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.7 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  49 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.8 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29:  50 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.7 pbw: 1 eraw:   0 ecdd:   0 iter:   8 pcc:  56
29:  51 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.8 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  52 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.7 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29:  53 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.8 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  54
29:  54 nin:  640 st: synced euw:  2 0 mf:  0 f:  -1.7 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  55 nin:  640 st: synced euw:  5 1 mf:  0 f:  -1.6 pbw: 1 eraw:   0 ecdd:   0 iter:   7 pcc:  56
29:  56 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.6 pbw: 1 eraw:   0 ecdd:   0 iter:   8 pcc:  56
29:  57 nin:  640 st: synced euw:  3 0 mf:  0 f:  -1.5 pbw: 1 eraw:   0 ecdd:   0 iter:   7 pcc:  56
29:  58 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.4 pbw: 1 eraw:   0 ecdd:   4 iter: 100 pcc:  36
29:  59 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.5 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  60 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.4 pbw: 1 eraw:   0 ecdd:   0 iter:   8 pcc:  56
29:  61 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.4 pbw: 1 eraw:   0 ecdd:   0 iter:   3 pcc:  56
29:  62 nin:  640 st: synced euw:  2 0 mf:  0 f:  -1.4 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  63 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.3 pbw: 1 eraw:   0 ecdd:   0 iter:   8 pcc:  56
29:  64 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.4 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  65 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.3 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  66 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.4 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  55
29:  67 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.5 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  68 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.6 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  69 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.6 pbw: 1 eraw:   0 ecdd:   8 iter: 100 pcc:  23
29:  70 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.5 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  71 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.4 pbw: 1 eraw:   0 ecdd:   6 iter: 100 pcc:  38
29:  72 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.5 pbw: 1 eraw:   0 ecdd:   0 iter:  18 pcc:  56
29:  73 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.6 pbw: 1 eraw:   0 ecdd:   0 iter:   7 pcc:  56
29:  74 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.6 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  75 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.5 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29:  76 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.6 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  77 nin:  640 st: synced euw:  2 0 mf:  0 f:  -1.5 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  78 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.4 pbw: 1 eraw:   0 ecdd:   0 iter:   8 pcc:  56
29:  79 nin:  640 st: synced cohpsk_ch: SNR3k(dB):     1.39  C/No....:    36.16
29: cohpsk_ch: peak.....: 16536.20  RMS.....:  6053.63   CPAPR.....:  8.73 
29: cohpsk_ch: Nsamples.:    80000  clipped.:     0.00%  OutClipped:  0.00%
29: euw:  3 0 mf:  0 f:  -1.3 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  55
29:  80 nin:  640 st: synced euw:  2 0 mf:  0 f:  -1.2 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  81 nin:  640 st: synced euw:  3 0 mf:  0 f:  -1.1 pbw: 1 eraw:   0 ecdd:   0 iter:   7 pcc:  56
29:  82 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.1 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  83 nin:  640 st: synced euw:  4 1 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter:   9 pcc:  56
29:  84 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.1 pbw: 1 eraw:   0 ecdd:   0 iter:  12 pcc:  56
29:  85 nin:  640 st: synced euw:  4 1 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29:  86 nin:  640 st: synced euw:  3 0 mf:  0 f:  -1.1 pbw: 1 eraw:   0 ecdd:   0 iter:   7 pcc:  56
29:  87 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.2 pbw: 1 eraw:   0 ecdd:   0 iter:   3 pcc:  56
29:  88 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.3 pbw: 1 eraw:   0 ecdd:   0 iter:  16 pcc:  56
29:  89 nin:  640 st: synced euw:  2 0 mf:  0 f:  -1.2 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  90 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.1 pbw: 1 eraw:   0 ecdd:   0 iter:   3 pcc:  56
29:  91 nin:  640 st: synced euw:  2 0 mf:  0 f:  -1.2 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  92 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.1 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  93 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter:  11 pcc:  56
29:  94 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29:  95 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter:   8 pcc:  56
29:  96 nin:  640 st: synced euw:  2 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29:  97 nin:  640 st: synced euw:  2 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  55
29:  98 nin:  640 st: synced euw:  0 0 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter:   3 pcc:  56
29:  99 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29: 100 nin:  640 st: synced euw:  2 0 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29: 101 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   1 iter: 100 pcc:  46
29: 102 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter:  13 pcc:  56
29: 103 nin:  640 st: synced euw:  1 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   0 iter:   2 pcc:  56
29: 104 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29: 105 nin:  640 st: synced euw:  1 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   0 iter:   7 pcc:  56
29: 106 nin:  640 st: synced euw:  2 0 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  55
29: 107 nin:  640 st: synced euw:  1 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   0 iter:   3 pcc:  56
29: 108 nin:  640 st: synced euw:  1 0 mf:  0 f:  -1.0 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29: 109 nin:  640 st: synced euw:  3 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29: 110 nin:  640 st: synced euw:  1 0 mf:  0 f:  -0.8 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29: 111 nin:  640 st: synced euw:  3 0 mf:  0 f:  -0.7 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29: 112 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.8 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29: 113 nin:  640 st: synced euw:  1 0 mf:  0 f:  -0.9 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  55
29: 114 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.8 pbw: 1 eraw:   0 ecdd:   2 iter:  15 pcc:  56
29: 115 nin:  640 st: synced euw:  1 0 mf:  0 f:  -0.7 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29: 116 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.6 pbw: 1 eraw:   0 ecdd:   0 iter: 100 pcc:  55
29: 117 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.5 pbw: 1 eraw:   0 ecdd:   0 iter:   5 pcc:  56
29: 118 nin:  640 st: synced euw:  2 0 mf:  0 f:  -0.4 pbw: 1 eraw:   0 ecdd:   0 iter:   3 pcc:  56
29: 119 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.5 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29: 120 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.4 pbw: 1 eraw:   0 ecdd:   0 iter:   6 pcc:  56
29: 121 nin:  640 st: synced euw:  2 0 mf:  0 f:  -0.5 pbw: 1 eraw:   0 ecdd:   0 iter:   4 pcc:  56
29: 122 nin:  640 st: synced euw:  0 0 mf:  0 f:  -0.6 pbw: 1 eraw:   0 ecdd:   0 iter:   7 pcc:  56
29: 123 nin:  640 st: synced euw:  1 0 mf:  0 f:  -0.5 pbw: 1 eraw:   0 ecdd:   0 iter:   3 pcc:  56
29: 124 nin:  640 st: synced euw:  4 1 mf:  0 f:  -0.6 pbw: 1 eraw:   0 ecdd:   0 iter:  16 pcc:  56
29: BER......: 0.0988 Tbits: 13776 Terrs:  1361 Tpackets:   123
29: Coded BER: 0.0379 Tbits:  6888 Terrs:   261
29: Coded PER: 0.2602 Tpkts:   123 Tpers:    32 Tthruput:    91
2/3 Test #29: test_OFDM_modem_700E_AWGN ........***Failed    0.07 sec
test 50
    Start 50: test_freedv_api_700D_speech

50: Test command: /bin/sh "-c" "cd /Users/mooneer/devel/freedv-gui/codec2/build_osx/src;
                            ./freedv_tx 700D ../../raw/ve9qrp_10s.raw - |
                            ./cohpsk_ch - - -20 --Fs 8000 |
                            ./freedv_rx 700D - /dev/null --squelch -2 -vv"
50: Test timeout computed to be: 1500
50: cohpsk_ch: Fs: 8000 NodB: -20.00 foff: 0.00 Hz fading: 0 nhfdelay: 0 clip: 32767.00 ssbfilt: 1 complexout: 0
50:   0 nin: 1280 st: search euw:  0  0 mf:  0 f:   0.0 pbw: 1 snr:  0.0                                        rxst: ----
50:   1 nin:  548 st: search euw:  0  0 mf:  0 f:   0.0 pbw: 1 snr:  0.0                                        rxst: ----
50:   2 nin: 1280 st: trial  euw:  1  0 mf:  0 f:   0.0 pbw: 1 snr:  1.2 eraw:   0 ecdd:   0 iter: 100 pcc:  91 rxst: EBST
50:   3 nin: 1280 st: trial  euw:  0  0 mf:  0 f:   0.0 pbw: 1 snr:  3.6 eraw:   0 ecdd:   0 iter: 100 pcc: 104 rxst: EBST
50:   4 nin: 1280 st: trial  euw:  0  0 mf:  0 f:   0.0 pbw: 1 snr:  3.4 eraw:   0 ecdd:   0 iter: 100 pcc: 102 rxst: EBST
50:   5 nin: 1280 st: trial  euw:  0  0 mf:  0 f:  -0.1 pbw: 0 snr:  3.1 eraw:   0 ecdd:   0 iter: 100 pcc: 107 rxst: EBST
50:   6 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.1 eraw:   0 ecdd:   0 iter: 100 pcc: 109 rxst: EBS-
50:   7 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.9 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:   8 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.1 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:   9 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  4.1 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  10 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  3.9 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  11 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.6 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  12 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.3 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  13 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.2 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  14 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.2 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  15 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.2 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  16 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  3.1 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  17 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  2.9 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  18 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  5.3 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  19 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  5.0 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  20 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.4 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  21 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.0 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  22 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  3.6 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  23 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.1 pbw: 0 snr:  3.6 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  24 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  3.5 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  25 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.4 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  26 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.1 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  27 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  2.6 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  28 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  4.9 eraw:   0 ecdd:   0 iter:   1 pcc: 112 rxst: -BS-
50:  29 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.6 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  30 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  4.1 eraw:   0 ecdd:   0 iter:   4 pcc: 112 rxst: -BS-
50:  31 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.2 eraw:   0 ecdd:   0 iter:   1 pcc: 112 rxst: -BS-
50:  32 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.0 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  33 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  3.7 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  34 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  4.0 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  35 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.8 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  36 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.5 eraw:   0 ecdd:   0 iter:   1 pcc: 112 rxst: -BS-
50:  37 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  3.3 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  38 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.4 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  39 nin: 1280 st: synced euw:  1  0 mf:  0 f:  -0.0 pbw: 0 snr:  5.1 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  40 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.7 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  41 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  6.3 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  42 nin: 1280 st: synced euw:  1  0 mf:  0 f:  -0.0 pbw: 0 snr:  5.8 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  43 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  5.1 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  44 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  4.8 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  45 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  4.4 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  46 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  4.1 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  47 nin: 1280 st: synced euw:  1  0 mf:  0 f:  -0.1 pbw: 0 snr:  3.8 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  48 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  3.4 eraw:   0 ecdd:   0 iter:   1 pcc: 112 rxst: -BS-
50:  49 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  5.7 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  50 nin: 1280 st: synced euw:  2  0 mf:  0 f:   0.0 pbw: 0 snr:  5.6 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  51 nin: 1280 st: synced euw:  1  0 mf:  0 f:   0.0 pbw: 0 snr:  5.5 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  52 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  5.2 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  53 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.4 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  54 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  3.8 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  55 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  5.0 eraw:   0 ecdd:   0 iter:   4 pcc: 112 rxst: -BS-
50:  56 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.7 eraw:   0 ecdd:   0 iter:   4 pcc: 112 rxst: -BS-
50:  57 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.0 pbw: 0 snr:  4.3 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  58 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  4.0 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  59 nin: 1280 st: synced euw:  0  0 mf:  0 f:   0.0 pbw: 0 snr:  3.7 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50:  60 nin: 1280 st: synced euw:  0  0 mf:  0 f:  -0.1 pbw: 0 snr:  3.5 eraw:   0 ecdd:   0 iter:   3 pcc: 112 rxst: -BS-
50:  61 nin: 1280 st: synced euw:  1  0 mf:  0 f:  -0.0 pbw: 0 snr:  3.8 eraw:   0 ecdd:   0 iter:   2 pcc: 112 rxst: -BS-
50: cohpsk_ch: SNR3k(dB):     2.20  C/No....:    36.97
50: cohpsk_ch: peak.....: 17030.04  RMS.....:  7043.93   CPAPR.....:  7.67 
50: cohpsk_ch: Nsamples.:    79360  clipped.:     0.00%  OutClipped:  0.00%
50: frames decoded: 62  output speech samples: 71680
3/3 Test #50: test_freedv_api_700D_speech ......***Failed  Required regular expression not found. Regex=[frames decoded: 62  output speech samples: 72960
]  0.07 sec

0% tests passed, 3 tests failed out of 3

Total Test time (real) =   1.61 sec

The following tests FAILED:
	 17 - test_OFDM_modem_burst_acq_port (Failed)
	 29 - test_OFDM_modem_700E_AWGN (Failed)
	 50 - test_freedv_api_700D_speech (Failed)
Errors while running CTest
HT-TM05:build_osx mooneer$ 
```

drowe67:
OK (1) looks like a path thing, I think we've handled that in similar tests like `tofdm`, (2) I'm not sure about (3) is a different random numbers thing like that test last week that means we'll need to work out how to compare to a >= threshold in the output speech frames.

I'll see if I can make some progress on these on my Sunday ...

PR #172, opened 2021-01-22
A typo in the __EMBEDDED__ guard adds constfloat -> float cast warning

erdgeist:
Since everywhere else in the code the macro __EMBEDDED__ is used, I assume, that's a typo. Right?


drowe67:
Thanks @erdgeist - yes I think you are right.  Note to self - we'll have to check the stm32 build works next time we do some maintenance on that platform.

PR #173, opened 2021-02-07
LPCNet package name is officially in lower-case according to apt.

tmiw:
@drowe67, apparently LPCNet's package name is in lower case in the CPack packages. This is needed for apt-get to properly install the packages (but dpkg -i && apt-get -f still worked with it in mixed-case for some reason /shrug).


drowe67:
@tmiw oops sorry I missed this one somehow.  Pls feel free to ping me if I'm late is responding :slightly_smiling_face:

PR #181, opened 2021-04-30
Single carrier PSK modem

drowe67:
@srsampson I recall you were interested in single carrier modems (QO-100 is a good use case, or UHF).  Turns out the recent data work to the OFDM modem includes the features we need to make single carrier operation work.  Only it's not really OFDM any more as there's no other carriers to be orthogonal too!

Uncoded, burst mode Octave demo:
```
ofdm_tx("test_single.raw","1",1,100,"awgn","bursts",3)
ofdm_rx("test_single.raw","1","packetsperburst",1)
<snip>
BER..: 0.0000 Tbits:   240 Terrs:     0
Packets:   3 Npre: 3 Npost: 0 SNR3k: 16.18
```

I'm working on another project atm, but if you want to push this fwd I'm happy to assist.  Or we can leave it for later :slightly_smiling_face: Obvious addition would be a raised cosine (root Nyquist) filter.  We have them already on some other modems (FDMDV cohpsk).  I just took a guess at the waveform design (it's a low symbol rate), feel free to mess around with it.

Screenshot from 2021-04-30 09-39-22
Screenshot from 2021-04-30 09-39-27

### DIFF
--- a/octave/ofdm_lib.m
+++ b/octave/ofdm_lib.m
@@ -367,7 +367,10 @@ function config = ofdm_init_mode(mode="700D")
     config.state_machine = "data"; 
     config.amp_scale = 300E3; config.clip_gain1 = 2.2; config.clip_gain2 = 0.8;
   elseif strcmp(mode,"1")
-    Ns=5; config.Np=10; Tcp=0; Tframe = 0.1; Ts = Tframe/Ns; Nc = 1;
+    Ns=5; config.Np=10; Tcp=0; Ts=0.010; Nc = 1;
+    config.state_machine = "data"; config.data_mode = "streaming";
+    config.edge_pilots = 0;    
+    config.Ntxtbits = 0; config.Nuwbits = 10; config.bad_uw_errors = 3;
   else
     % try to parse mode string for user defined mode
     vec = sscanf(mode, "Ts=%f Nc=%d Ncp=%f");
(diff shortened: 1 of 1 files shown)


srsampson:
I experimented a bit with costas loop, but that was not going to be good for burst mode. Where I left off was just taking the 4th power of the PSK divided by 4 to get phase. Then for frequency tracking I started looking into an FFT. The FFT might work, but needs more resolution than I was using. I never saw the carrier move with my transmit offset.

It does use a root raised cosine filter though. hs = gen_rn_coeffs(.50, 1.0/8000.0, 1600, 10, 5);

Not sure if that is valid but it seemed to look OK when viewing the spectrum.

https://github.com/srsampson/SingleCarrier

You might look and see if this code is still viable.

drowe67:
Ahh yes I've used that 4th power method, thanks for reminding me. Pros are no pilot symbol overhead, cons are you need the UW for pi/4 ambiguity resolution (no big deal) and given it's based on a non-linearity it might fall over at the low SNRs that modern FEC runs at.

drowe67:
> I saw the 4th power rule used in cohpsk so was trying that out. That cohpsk code was more sophisticated though.
> 
> https://github.com/drowe67/codec2/blob/master/src/cohpsk.c#L889

Oh OK looks like it's being used for freq offset tracking.  I've forgotten about that!  Curious how I didn't use the pilots symbols for freq tracking too in that modem.

Issue #182, opened 2021-05-01
refactor modem_stats system in FreeDV API and modems


drowe67:
The current model has some problems, and should be cleaned up/refactored:
1. Some of the averaging of stats like SNR happens per call, it should be updated at a fixed sample rate set by the current modem.
2. Cleaner sample/update model, they are mixed at the moment.
3.  `modem_stats_get_rx_spectrum()` is a helper function unrelated to the FreeDV API, consider splitting it's states (e.g. for the FFT) out of the `modem_stats` struct.

PR #183, opened 2021-05-01
stop overwriting FFT states when sampling modem_stats with OFDM modes

drowe67:
@tmiw Can you please try this with freedv-gui and see if it resolves the segfault issue with 84e43e2 ?  It works OK for me.

This is a short term fix, I have documented the related design issue in https://github.com/drowe67/codec2/issues/182 so it can be fixed properly at some stage.


tmiw:
@drowe67, looks good to me (macOS and Ubuntu 20.04). 👍

PR #184, opened 2021-05-10
SNR nans from ofdm

drowe67:
Bug reported during freedv-gui testing [here](https://github.com/drowe67/freedv-gui/pull/123#issuecomment-832389359) here of squelch not opening on OFDM modes.  Traced to a `nan` in SNR.  Not repeatable on all machines.

### DIFF
--- a/misc/est_n0.c
+++ b/misc/est_n0.c
@@ -14,7 +14,15 @@
 #include <math.h>
 #include <stdlib.h>
 #include <stdio.h>
+
 #include "defines.h"
+
+#ifndef M_PI
+#define M_PI        3.14159265358979323846f
+#endif
+
+#define cmplx(value) (cosf(value) + sinf(value) * I)
+#define cmplxconj(value) (cosf(value) + sinf(value) * -I)
 
 int main(int argc, char *argv[]) {
     MODEL model;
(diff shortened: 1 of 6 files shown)


tmiw:
Seeing a fair number of compiler warnings similar to below (macOS 11.2.3):

```
/Users/mooneer/devel/freedv-gui/codec2/src/ofdm.c:1134:17: warning: absolute value function 'cabsf' given an argument of type '_Complex double' but has parameter of type '_Complex float' which may cause truncation of value [-Wabsolute-value]
        mag1 += cabsf(known_samples[i]*conj(known_samples[i]));
                ^
/Users/mooneer/devel/freedv-gui/codec2/src/ofdm.c:1134:17: note: use function 'cabs' instead
        mag1 += cabsf(known_samples[i]*conj(known_samples[i]));
                ^~~~~
                cabs
```

They don't seem related to these changes, however.

Anyway, I haven't been able to duplicate the issue with this PR thus far (played back the same input file about 5-6 times now).

drowe67:
Thanks @tmiw - I've had a go at removing some of those warnings in https://github.com/drowe67/codec2/pull/184/commits/2723728ed977ac254508b59e626b276a1779c670.  If any others pop up please let me know.  My `gcc` doesn't support that warning, must be a bit older.  If you want to fix any yourself you can do a quick test with a subset of the ctests:
```
ctest -R test_OFDM
```
Or just push them here....

Good to see we have made some progress on the OP bug :slightly_smiling_face:

srsampson:
I made a slight change from log10() to log10f()

tmiw:
@drowe67, no more warnings in ofdm.c with the latest changes. 👍  There are warnings for other files ([attached](https://github.com/drowe67/codec2/files/6453156/codec2.warnings.txt)) but maybe those can be handled in another PR?

srsampson:
cexp() is a very slow function. Used the cmplx(float) define and also made changes to reduce warnings.

PR #185, opened 2021-05-15
Optimising cohspk modem

drowe67:
Attempting to address https://github.com/drowe67/freedv-gui/issues/126

Test below was initially 9.8s, so 30% faster:

```
time ./src/freedv_rx 700C ~/freedv-gui/wav/ve9qrp_700d.wav /dev/null
frames decoded: 1404  output speech samples: 898560

real	0m6.810s
user	0m6.810s
sys	0m0.000s
```

### DIFF
--- a/octave/esno_est.m
+++ b/octave/esno_est.m
@@ -126,7 +126,7 @@ function esno_est_test_c(channel="awgn")
     printf("\nRunning C version....\n");
     path_to_unittest = "../build_linux/unittest"
     if getenv("PATH_TO_UNITEST")
-      path_to_unittest_exe = getenv("PATH_TO_UNITTEST")
+      path_to_unittest = getenv("PATH_TO_UNITEST")
       printf("setting path from env var to %s\n", path_to_unittest);
     end
     system(sprintf("%s/tesno_est %s %d > tesno_est_out.txt", path_to_unittest, "esno_est.iqfloat", length(rx_syms)));
--- a/src/cohpsk.c
+++ b/src/cohpsk.c
@@ -527,17 +527,20 @@ void tx_filter_and_upconvert_coh(COMP tx_fdm[], int Nc, const COMP tx_symbols[],
 
 void corr_with_pilots(float *corr_out, float *mag_out, struct COHPSK *coh, int t, float f_fine)
 {
-    COMP  acorr, f_fine_rect, f_corr;
+    COMP  acorr, f_fine_rect[NPILOTSFRAME+2], f_corr;
     float mag, corr;
     int   c, p, pc;
+
+    for (p=0; p<NPILOTSFRAME+2; p++) {
+        f_fine_rect[p].real = cosf(f_fine*2.0*M_PI*(sampling_points[p]+1.0)/COHPSK_RS);
+        f_fine_rect[p].imag = sinf(f_fine*2.0*M_PI*(sampling_points[p]+1.0)/COHPSK_RS);
+    }
 
     corr = 0.0; mag = 0.0;
     for (c=0; c<COHPSK_NC*ND; c++) {
         acorr.real = 0.0; acorr.imag = 0.0;
         for (p=0; p<NPILOTSFRAME+2; p++) {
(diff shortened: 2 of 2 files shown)


srsampson:
Built on your optimization to compute inner only once.

srsampson:
Saw a couple atan and log that could be optimized.

tmiw:
One test failure on macOS with this branch:

```
HT-TM05:build_osx mooneer$ ctest -V -R test_OFDM_modem_esno_est_c
UpdateCTestConfiguration  from :/Users/mooneer/devel/freedv-gui/codec2/build_osx/DartConfiguration.tcl
Parse Config file:/Users/mooneer/devel/freedv-gui/codec2/build_osx/DartConfiguration.tcl
UpdateCTestConfiguration  from :/Users/mooneer/devel/freedv-gui/codec2/build_osx/DartConfiguration.tcl
Parse Config file:/Users/mooneer/devel/freedv-gui/codec2/build_osx/DartConfiguration.tcl
Constructing a list of tests
test 16
    Start 16: test_OFDM_modem_esno_est_c

16: Test command: /bin/sh "-c" "cd /Users/mooneer/devel/freedv-gui/codec2/octave;
                            echo 'pkg load signal; esno_est; esno_est_tests_c' |
                            PATH_TO_UNITEST=/Users/mooneer/devel/freedv-gui/codec2/build_osx/unittest/ DISPLAY="" octave-cli"
16: Test timeout computed to be: 1500
16: EsNo:  5.00 EsNo_meas:  4.84 EsNo_est:  4.71
16: 
16: Running C version....
16: path_to_unittest = ../build_linux/unittest
16: path_to_unittest_exe = 
16: setting path from env var to ../build_linux/unittest
16: sh: ../build_linux/unittest/tesno_est: No such file or directory
16: error: load: unable to determine file format of 'tesno_est_out.txt'
16: error: called from
16:     esno_est_test_c at line 133 column 5
16:     esno_est_tests_c at line 143 column 5
1/1 Test #16: test_OFDM_modem_esno_est_c .......***Failed  Required regular expression not found. Regex=[AWGN Pass.*MPP Pass
]  0.22 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.23 sec

The following tests FAILED:
	 16 - test_OFDM_modem_esno_est_c (Failed)
Errors while running CTest
```

tmiw:
100% pass now with the latest commit I pushed. 👍

drowe67:
> Built on your optimization to compute inner only once.

Thanks @srsampson :+1: Did you measure any performance improvement? 

I think most of the CPU is being used in [this](https://github.com/drowe67/codec2/blob/master/src/cohpsk.c#L993) loop, in particular the `rate_Fs_rx_processing()`.  In this modem, we actually do a complete demod at 3 different freq shifts, then do a trial freq estimation on the baseband symbols.  So it's pretty CPU heavy, but that's not really a problem for most x86 platforms.

drowe67:
> 100% pass now with the latest commit I pushed. +1

Nice catch :+1: - thanks @tmiw

srsampson:
> > Built on your optimization to compute inner only once.
> 
> Thanks @srsampson  Did you measure any performance improvement?
> 
> I think most of the CPU is being used in [this](https://github.com/drowe67/codec2/blob/master/src/cohpsk.c#L993) loop, in particular the `rate_Fs_rx_processing()`. In this modem, we actually do a complete demod at 3 different freq shifts, then do a trial freq estimation on the baseband symbols. So it's pretty CPU heavy, but that's not really a problem for most x86 platforms.

I don't have a PI example ready, but I didn't see any difference on my Ubuntu running the command line. I need to round-up my parts and build a PI for freedv.

I remember this loop being quite intensive from before. I wouldn't imagine anyone running cohpsk on a small device. Even the SM1000 would probably collapse. But if it can be done, a PI4 might be good enough.

drowe67:
> I remember this loop being quite intensive from before. I wouldn't imagine anyone running cohpsk on a small device. Even the SM1000 would probably collapse. But if it can be done, a PI4 might be good enough.

Yes the acquisition system would need re-writing for a very small system like the SM1000.  It appears to be running OK (just) on my PI 3B now https://github.com/drowe67/freedv-gui/issues/126

Issue #186, opened 2021-05-17
"In-place" FFT codepath is not used


nevercast:
Hello, I think this line 

https://github.com/drowe67/codec2/blob/master/src/codec2_fft.c#L141

Is meant to read

```
if (cfg->nfft <= 512)
```

Since nfft is multiplied by sizeof during the memcopy, it's safe that the 512 buffer and 512 condition be directly compared without converting into bytes.

nevercast:
I see this is in direct contrast with the original commit https://github.com/drowe67/codec2/commit/413faeb29ceeb7167acf2a8b50e30ddf32d135d2 

So perhaps I'm mistaken in this regard.

drowe67:
@nevercast - thanks.  :thinking: must admit I'm not sure myself.  @db4ple do you have any thoughts on this one?

nevercast:
I was almost 100% confident until I noticed it was less of a typo and more an intentional commit.

drowe67:
@nevercast https://github.com/drowe67/codec2/pull/187

db4ple:
Hi, reading the comment I made I can clearly see that @nevercast is right. If it fits into the 512 element long buffer in allocated on stack, we want to use it. However, in any case, we should move the allocation of `in` inside the `if``.  
So this code needs fixing to make sense at all.

drowe67:
Fixed in #187 thanks @nevercast and @db4ple

PR #187, opened 2021-05-17
codec_fft.c codepath issue

drowe67:
Looking into https://github.com/drowe67/codec2/issues/186, it does indeed appear that the first codepath is never taken, even though it should be for `nfft=256` or `nfft=512`:

```
ctest -V _R test_codec2
<snip>
codepath 2 nfft: 512
<snip>
codepath 2 nfft: 256
```

### DIFF
--- a/src/codec2_fft.c
+++ b/src/codec2_fft.c
@@ -133,20 +133,18 @@ void codec2_fft_inplace(codec2_fft_cfg cfg, codec2_fft_cpx* inout)
 {
 
 #ifdef USE_KISS_FFT
-    kiss_fft_cpx in[512];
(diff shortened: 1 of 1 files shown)


db4ple (commented):
As  @nevercast identified it has to be `if (cfg->nfft < 512)`.
We should move the allocation of `in` inside the `if`.

drowe67:
@nevercast @db4ple does https://github.com/drowe67/codec2/pull/187/commits/c4681f3bdc775a53f763ffb135e0fda60b2ddef2 look correct?  All the ctests passed on my machine.

nevercast:
LGTM 🚀  Thanks for fixing that :)

PR #188, opened 2021-05-19
SNR and Scatter not working on 2020

drowe67:
Hi @tmiw - thanks for reporting this.  As per https://github.com/drowe67/codec2/issues/182 - this general area of the FreeDV API (stats reporting/SNR/squelch) has been problematic in the past which suggests some more refactoring/design review is required.  Still thinking about the best way forward :thinking: 

Changes made:
1. Hooked up 2020 to new OFDM stats reporting logic, which fixed OP issue.
2. Secondary issue: OFDM modes now reset SNR to -5dB when out of sync, it was previously frozen as `modem_stats` are not updated when out of sync.

I tried full duplex 2020 with an ALSA loopback device and it worked OK.

Further work - test the OFDM modes (700D/E/2020) on air with some real QSOs, make sure squelch/SNR is working correctly.  The changes so far have been tested with the `freedv-gui/wav` demo files which don't have any noise.

### DIFF
--- a/README_freedv.md
+++ b/README_freedv.md
@@ -99,18 +99,18 @@ FreeDV 2400A and FreeDV 2400B are modes designed for VHF radio. FreeDV 2400A is 
 
 Demos of FreeDV 2400A and 2400B:
 ```
-$ ./freedv_tx 2400A ../../raw/ve9qrp_10s.raw - | ./freedv_rx 2400A - - | play -t raw -r 8000 -s -2 -
-$ ./freedv_tx 2400B ../../raw/ve9qrp_10s.raw - | ./freedv_rx 2400B - - | play -t raw -r 8000 -s -2 -
+$ ./freedv_tx 2400A ../../raw/ve9qrp_10s.raw - | ./freedv_rx 2400A - - | play -t .s16 -r 8000 -
+$ ./freedv_tx 2400B ../../raw/ve9qrp_10s.raw - | ./freedv_rx 2400B - - | play -t .s16 -r 8000 -
 ```
 Note for FreeDV 2400A/2400B the modem signal sample rate is 48kHz.  To
 listen to the modem tones from FreeDV 2400B, or play them into a FM HT
 mic input:
 ```
-$ ./freedv_tx 2400B ../../raw/ve9qrp_10s.raw - | play -t raw -r 48000 -s -2 -
+$ ./freedv_tx 2400B ../../raw/ve9qrp_10s.raw - | play -t .s16 -r 48000 -
 ```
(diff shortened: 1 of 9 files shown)


tmiw:
@drowe67, seems to work okay here. 

Question: what is/should be the expected SNR behavior when out of sync? Wouldn't lack of sync imply that the signal is below the minimum needed for that particular mode (and thus -5 isn't necessarily inaccurate)?

drowe67:
> Question: what is/should be the expected SNR behavior when out of sync? Wouldn't lack of sync imply that the signal is below the minimum needed for that particular mode (and thus -5 isn't necessarily inaccurate)?

Good questions ...  In the absence of a signal I guess we could say the SNR is undefined (or super low), so -something dB is a good choice.  I set it at -5dB as 700D can decode OK down to -2dB on an AWGN channel.  I have designed the waveforms to get sync under the SNR where speech decodng kicks in, so SNR should be meaningful (and useful for the squelch logic) whenever we have sync.

drowe67:
Thanks @tmiw - I have just pushed some changes that resolves the issues you reported above for 2400B and 800XA on my tests (using the `freedv-gui/wav` samples).  That might explain why those modes didn't work with your multi-rx feature.

A better design for the sync/SNR/modem_stats features is gradually taking shape:
1. When the demod for each mode is called via the `freedv_comprx()`, we update stats like SNR and sync.
1. When `freedv_get_modem_extended_stats()` is called we sample the SNR and sync, and other info like scatter diagram points, clock offset etc.

Previously, the sampling/estimation interface has not been separate, for example some of the SNR averaging was done when `freedv_get_modem_extended_stats()` was called (the sampling point).  The conventions are also been different for different modes.  Gradually working through that now, but still some work to go.

tmiw:
@drowe67, as mentioned in https://github.com/drowe67/freedv-gui/pull/129 2400B seems to be stuck at around +5dB with sync remaining almost always on. This is with normal band noise on 20 meters, too.

drowe67:
> @drowe67, as mentioned in [drowe67/freedv-gui#129](https://github.com/drowe67/freedv-gui/pull/129) 2400B seems to be stuck at around +5dB with sync remaining almost always on. This is with normal band noise on 20 meters, too.

Doh! OK with band noise input I'm seeing:
1. 800XA sync almost always on, SNR around 0.
2. 2400B SNR at 4-5dB, but sync flickering (like other modes).
1. 2400B analog pass through is not working.  This mode does have odd sample rates, but IIRC I had analog pass through working a few months ago.  800XA analog pass through works well.
2. When I switch to decode with a stored file ....  800XA sync quite slow, but 2400B comes up straight away.

I'll have another think :thinking: These are rarely used modes but it would be good to sort them out.

PR #192, opened 2021-06-15
README: microphone example

l29ah:


README.md
```diff
@@ -60,6 +60,11 @@ Also included:
    ```
    ./src/c2enc 1300 ../raw/hts1a.raw - | ./src/c2dec 1300 - - | aplay -f S16_LE
    ```
+   ```
+1. Or you can use your microphone and headphones to encode and listen to the result on the fly:
```
tmiw, line 64: Indentation should match the other paragraphs above it.

BTW, @drowe67, should each of the "1." here be incremented to make the raw .md file easier to read?

drowe67: @tmiw the indent looks OK to me, and it renders OK.  Sorry if I missed something?  There was an extra set of quotes.

I could go either way with the numbering, I usually just stick in all 1s but I note this file has a mixture of styles :thinking:

tmiw: Just checked the actual file and it looks correct. I think the diff view in the PR had weird spacing or something. /shrug

PR #193, opened 2021-06-18
Automated Voice Testing

drowe67:
Set of scripts to automate testing of FreeDV voice modes against compressed SSB.

Goals:
1. Compare SSB to 700C/D/E at same peak power, and over the same channel.
1. Use automation to run multiple tests over time in order to sample different channels, e.g. AWGN, NVIS, interstate, international.
1. Discover any corner cases that cause FreeDV to fall over, e.g. sync problems, fast fading or large delay spread, impulse noise, or other real world artifacts.

The key point is the script ensures the analog and FreeDV peak Tx power is the same.  It also uses compression on the SSB, but in a controlled fashion using software (not rig dependent).  It plays the samples straight after each other, on the assumption the channel is "stationary" over a short time period.  I've been using a 10 second sample.

Method:
1. Takes a speech sample file as input.
1. Compresses it using a Hilbert clipper to create the "analog" reference.  This can then be fed into an analog SSB Tx to simulate/implement SSB radio compression in a controlled way.
1. Take the same speech input file and encode it to a FreeDV waveform.
1. Concatenate the compressed analog and FreeDV waveform into one file.  They will be normalised to the same peak level.
1. Transmit that file using a SSB radio under hamlib control.  Make sure SSB radio mic compression is off (e.g. use data mode) and "just nudge ALC" during Tx as per normal FreeDV set up.
1. Simultaneously sample (record) the received signal on a remote KiwiSDR.
1. Decode and compare.

Tx only:
```
cd codec2/unittest
./ota_voice_test.sh ~/Desktop/vk5dgr_test.wav -t -d -g 4
```
Replace `vk5dgr_test.wav` with a (say 10 second) recording of your own voice/callsign.  Run the above script while manually monitoring the signal on a receiver.

Tx and Rx:
```
./ota_voice_test.sh ~/Desktop/vk5dgr_test.wav sdr.ironstonerange.com -d -g 4
```
`-g` is the gain used to adjust the compression used for SSB.


drowe67:
@tmiw :point_up:

drowe67:
@tmiw some plots from a test on ironstone on 700D. Will email you the wave file as GitHub doesn't like wave files.  I also realised FreeDV clipping/compression wasn't on, I think that's made a signficant difference.

spec
time_snr

There's still a bug in the SSB versus FreeDV levels, FreeDV looks a bit low on audacity (`tx.wav` file):
Screenshot from 2021-06-19 12-06-47
I'm not sure if this is significant, I'll drill down a bit more.  0.4 versus 0.8 is 2dB, but the ALC might be normalising them.  I notice on my IC7200 the SSB is at 2 bars of ALC, the FreeDV at 1.

tmiw:
@drowe67, FWIW, 2 bars on the 7200 gets me up to full power, not 1. The number of bars also seems dependent on your SWR; I've noticed a greater number when my magnetic loop is slightly out of tune, for instance (due to temperature or other factors).

unittest/ota_voice_test.sh
```diff
@@ -181,7 +183,11 @@ usb_lsb_upper=$(echo ${usb_lsb} | awk '{print toupper($0)}')
 run_rigctl "\\set_mode PKT${usb_lsb_upper} 0" $model
 run_rigctl "\\set_freq ${freq_Hz}" $model
 run_rigctl "\\set_ptt 1" $model
-aplay --device="plughw:CARD=CODEC,DEV=0" -f S16_LE tx.raw 2>/dev/null
+if [ `uname` == "Darwin" ]; then
+    play -t raw -b 16 -c 1 -r 8000 -e signed-integer --endian little tx.raw 
+else
+    aplay --device="plughw:CARD=CODEC,DEV=0" -f S16_LE tx.raw 2>/dev/null
+fi
```
drowe67, line 190: You might be able to use `play -t -.s16 -r 8000 -c 1 .....`

Yes the `--device="plughw:CARD=CODEC,DEV=0"` selects the IC7200 USB device by name so I don't have to set it as the default sound card.  They may be an equivalent `sox` method on macOS.

drowe67:
> @drowe67, FWIW, 2 bars on the 7200 gets me up to full power, not 1. The number of bars also seems dependent on your SWR; I've noticed a greater number when my magnetic loop is slightly out of tune, for instance (due to temperature or other factors).

OK interesting!  I can't seem to get more than 75W peak out of mine, not sure why.  I have a 13.8V 20A power supply.

unittest/ota_voice_test.sh
tmiw: That command didn't work, unfortunately, but I did find out how to override the output audio device. Unfortunately there seem to be limits, as evidenced here:

```
HT-TM05:unittest mooneer$ sox -V6 -n -t coreaudio junkname
sox:      SoX v
time:     Dec  1 2020 00:54:51
issue:    macosx
uname:    Darwin Mooneers-16-MBP.home.lifeafterking.org 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:33 PDT 2021; root:xnu-7195.121.3~9/RELEASE_X86_64 x86_64
compiler: gcc Apple LLVM 12.0.0 (clang-1200.0.32.27)
arch:     1288 48 88 L 
sox INFO nulfile: sample rate not specified; using 48000

Input File     : '' (null)
Channels       : 1
Sample Rate    : 48000
Precision      : 32-bit

sox INFO coreaudio: Found Audio Device "Background Music"

sox INFO coreaudio: Found Audio Device "Background Music"

sox INFO coreaudio: Found Audio Device "MacBook Pro Micr"

sox INFO coreaudio: Found Audio Device "MacBook Pro Spea"

sox INFO coreaudio: Found Audio Device "VB-Cable"

sox INFO coreaudio: Found Audio Device "VB-Cable"

sox INFO coreaudio: Found Audio Device "VB-Cable"

sox INFO coreaudio: Found Audio Device "ZoomAudi"

sox FAIL formats: can't open output file `junkname': can not open audio device
```

(I should be seeing "VB-Cable", "VB-Cable A" and "VB-Cable B". The latter is the one my SDR is normally set to for TX.)

tmiw:
> > @drowe67, FWIW, 2 bars on the 7200 gets me up to full power, not 1. The number of bars also seems dependent on your SWR; I've noticed a greater number when my magnetic loop is slightly out of tune, for instance (due to temperature or other factors).
> 
> OK interesting! I can't seem to get more than 75W peak out of mine, not sure why. I have a 13.8V 20A power supply.

I should clarify. That's full power as in it won't increase any further, not necessarily full as in 100W. The actual power output depends on other factors as well (e.g. SWR).

unittest/ota_voice_test.sh
drowe67: Oh dear, close but not quite there :thinking:

tmiw: Not sure there's much that can be done about that, unfortunately. That seems like a sox bug that they'll need to fix. Though perhaps I can run a MacPorts selfupdate and see if there's a new version available?

drowe67: I'm actually using an old Linux laptop as a server to run the tests.  That way when the cron stuff is set up I can leave it connected to the radio and running while my main development laptop get moved about through my working day.

Running `ssh -X` I can display the plots, and even replay wave files on my local machine using:
```
scp bear.lan:codec2/unittest/rx_freedv.wav /dev/stdout | aplay -f S16_LE
```

drowe67:
@tmiw I had a go at adjusting the relative SSB/FreeDV levels in https://github.com/drowe67/codec2/pull/193/commits/17bc504e41185fc429b3d0c2d2085081416596a1, however I'm a bit concerned that it might give different results across different input speech samples. you can check your relative levels with the `-x` options:
```
./ota_voice_test.sh ~/Desktop/vk5dgr_test.wav -x -d
eog tx.jpg
```

tx

tmiw:
@drowe67, looks okay to me:

tx

I did need to adjust one more path in the script to get that, though. I'll push the change momentarily.

drowe67:
@tmiw - sorry I'll try to make these scripts path independent in future ...

PR #194, opened 2021-06-18
FreeDV 700E and codec 2 equaliser

drowe67:
While working on https://github.com/drowe67/codec2/pull/193 I discovered the automatic mic equaliser was not being switched on for FreeDV 700E - Ouch!

This would be lowering the speech quality for freedv-gui users on 700E compared to 700C/D, as the Tools-Filter-Auto EQ check box would have been doing nothing for 700E.  I've also made sure the EQ is on by default when using the command line `freedv_tx` tool.

Screenshot from 2021-06-19 09-38-45

You can tell if the Auto EQ is working by:
1. Set up freedv-gui with a loopback sound device/full duplex.
1. Start freedv-gui and Tx
1. Tx a canned speech file Tools - Start Play File Mic In, check loop.
1. Monitor the stats window/Var field while toggling Tools-Filter-Auto EQ.  I get about 18 with Auto EQ off, about 10 with it on.  Toggling Auto EQ also resets the variance (spectral distortion) estimator.

### DIFF
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1250,7 +1250,7 @@ void freedv_set_phase_est_bandwidth_mode(struct freedv *f, int val) {
 }
 
 void freedv_set_eq(struct freedv *f, int val) {
-    if (FDV_MODE_ACTIVE( FREEDV_MODE_700C, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode)) {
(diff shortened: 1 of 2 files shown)


drowe67:
@tmiw @hobbes1069 :point_up: just found this one ... I'll merge into codec2/master shortly.  I'm just merging bug fixes in codec2 at the moment in order to get to a "release".

tmiw:
Not strictly needed but I also removed the checks from freedv-gui: https://github.com/drowe67/freedv-gui/pull/140.

PR #196, opened 2021-06-20
Enable Codec 2 Auto EQ for 800XA

drowe67:
Forgot one FreeDV mode in https://github.com/drowe67/codec2/pull/194 ! 800XA also uses codec 2 700C ....

### DIFF
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1236,7 +1236,6 @@ void freedv_set_tx_bpf(struct freedv *f, int val) {
 }
 
 /* DPSK option for OFDM modem, useful for high SNR, fast fading */
-
 void freedv_set_dpsk(struct freedv *f, int val) {
     if (FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_2020, f->mode)) {
         ofdm_set_dpsk(f->ofdm, val);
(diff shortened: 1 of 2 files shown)


src/freedv_api.c
```diff
@@ -1250,7 +1250,8 @@ void freedv_set_phase_est_bandwidth_mode(struct freedv *f, int val) {
 }
 
 void freedv_set_eq(struct freedv *f, int val) {
-    if (FDV_MODE_ACTIVE( FREEDV_MODE_700C, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_700E, f->mode)) {
+    if (FDV_MODE_ACTIVE( FREEDV_MODE_700C, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode) ||
```
tmiw, line 1253: Maybe checks like these can be consolidated in the individual Codec2 modes? For instance, `if (codec2_is_mode_700c(f->mode)) { ... }`.

drowe67: :+1:  Yes I was thinking along similar lines, e.g. if we add a field to note the codec 2 mode to the freedv states:
```
if (f->codec2_mode == CODEC2_MODE_700C)
```

drowe67:
@tmiw - I poked around a bit and found a simpler solution, the `codec2_700c_eq()` function is only active for Codec 2 700C, so it's safe to call from other modes.  I think this is a reasonable design choice - "do nothing silently" if the requested API function is not relevant to the current mode.  Simplifies the upper layers.

Pls tell me what you think of this solution, and also:

https://github.com/drowe67/freedv-gui/pull/141

tmiw:
@drowe67, looks good to me. 👍

PR #198, opened 2021-07-08
Avoid NANs in cohpsk modem with 0 inputs

drowe67:
@tmiw this seems to have fixed https://github.com/drowe67/freedv-gui/issues/143 for me, can you please confirm?


tmiw:
@drowe67, yep, all good here now. 👍

PR #206, opened 2021-08-07
Add 700E support to the SM1000.

tmiw:
Adds 700E support to the SM1000. Based on the dr-stm32-update branch due to requiring its fixes to actually run the firmware.

@drowe67, I noticed that the 700E signal seemed noisier on my other radio's waterfall than 1600 and 700D. Not sure why that's the case since I thought they were all supposed to be around the same output level?

CPU Load measurements (PE3):

| Mode | State | Period | On time | CPU Load est | Comment |
| 700D | Rx | 159.5 ms | 84.0 ms | 53% | Acquisition is probably worst case | 
| 700D | Tx | 158.0 ms | 98.5 ms | 62% | 
| 700D | Tx BPF + Clip | 161.0 ms | 128.5 ms | 80% | 
| 700E | Rx | 80.0 ms | 44.0 ms | 55% | 
| 700E | Tx | 80.0 ms | 50.0 ms | 63% | 
| 700E | Tx BPF + Clip | 80.1 ms | 65.0 ms | 81% |

### DIFF
--- a/stm32/unittest/scripts/tst_api_demod_setup
+++ b/stm32/unittest/scripts/tst_api_demod_setup
@@ -92,7 +92,7 @@ case "${TEST_OPT}" in
         # Copy N frames of a raw audio file to stm_in.raw. 
         dd bs=1280 count=100 if=../../../../raw/hts1.raw of=spch_in.raw \
 		> setup.log 2>&1
-	freedv_tx 700E spch_in.raw stm_in.raw --testframes --txbpf 0 \
+	freedv_tx 700E spch_in.raw stm_in.raw --testframes --txbpf 1 \
 		>> setup.log 2>&1
 	#
         # Reference
@@ -107,7 +107,7 @@ case "${TEST_OPT}" in
         # Copy N frames of a raw audio file to stm_in.raw. 
         dd bs=1280 count=96 if=../../../../raw/hts1.raw of=spch_in.raw \
 		> setup.log 2>&1
-	freedv_tx 700E spch_in.raw mod_bits.raw --testframes --txbpf 0 \
+	freedv_tx 700E spch_in.raw mod_bits.raw --testframes --txbpf 1 \
 		  >> setup.log 2>&1
 	cohpsk_ch mod_bits.raw stm_in.raw -20 --Fs 8000 -f -5  2>&1 | tee setup.log
 	
@@ -136,7 +136,7 @@ case "${TEST_OPT}" in
         # Copy N frames of a raw audio file to stm_in.raw. 
         dd bs=1280 count=48 if=../../../../raw/hts1.raw of=spch_in.raw \
 		> setup.log 2>&1
-	freedv_tx 700E spch_in.raw mod_bits.raw --txbpf 0 \
+	freedv_tx 700E spch_in.raw mod_bits.raw --txbpf 1 \
 		>> setup.log 2>&1
 	#
         # Reference - give it a hard time with some noise to exercise the LDPC codec and get us to max CPU
@@ -159,7 +159,7 @@ case "${TEST_OPT}" in
 	;;
     
     *)
-	printf "ERROR: invalid test option. Valid options are:\n  700D_plain_test\n  700D_AWGN_test\n  700D_AWGN_codec\n  1600_plain_codec\n"
+	printf "ERROR: invalid test option. Valid options are:\n  700[DE]_plain_test\n  700[DE]_AWGN_test\n  700[DE]_AWGN_codec\n  1600_plain_codec\n"
 	exit 1
       	;;
 
--- a/stm32/unittest/scripts/tst_api_demod_setup
+++ b/stm32/unittest/scripts/tst_api_demod_setup
@@ -109,22 +109,7 @@ case "${TEST_OPT}" in
 		> setup.log 2>&1
 	freedv_tx 700E spch_in.raw mod_bits.raw --testframes --txbpf 1 \
 		  >> setup.log 2>&1
-	cohpsk_ch mod_bits.raw stm_in.raw -20 --Fs 8000 -f -5  2>&1 | tee setup.log
-	
-        # Reference: - When the OFDM modem initially syncs, it often
-	#              has residual freq offset that causes abnormally
-	#              high LDPC decoder bit errors forthe first few
-	#              seconds.  This leads to a high coded BER being
-	#              reported for short duration tests.  This
-	#              settles down after a few seconds, and we get
-	#              the expected coded BER when averaging over
-	#              longer periods (e.g. 60s). However this
-	#              particular test is necessarily short due to the
-	#              slow speed of the semihosting system.  It is
-	#              therefore sufficient to check that the
-	#              performance is similar to the x86 C verison,
-	#              rather than expecting a low coded BER for a
-	#              short run.
+	cohpsk_ch mod_bits.raw stm_in.raw -22 --Fs 8000 -f -5  2>&1 | tee setup.log
 	
         freedv_rx 700E stm_in.raw ref_demod.raw -v --testframes  2>&1 --discard | tee ref_gen.log
 	;;
--- a/stm32/unittest/scripts/tst_api_demod_setup
+++ b/stm32/unittest/scripts/tst_api_demod_setup
@@ -92,7 +92,7 @@ case "${TEST_OPT}" in
         # Copy N frames of a raw audio file to stm_in.raw. 
         dd bs=1280 count=100 if=../../../../raw/hts1.raw of=spch_in.raw \
     	> setup.log 2>&1
-    freedv_tx 700E spch_in.raw stm_in.raw --testframes --txbpf 0 \
+    freedv_tx 700E spch_in.raw stm_in.raw --testframes --txbpf 1 \
     	>> setup.log 2>&1
 
         # Reference
--- a/stm32/unittest/scripts/tst_api_demod_setup
+++ b/stm32/unittest/scripts/tst_api_demod_setup
@@ -88,62 +88,62 @@ case "${TEST_OPT}" in
     700E_plain_test )
     	# Config is <mode>, <teswtframes> 
     	echo "81000010" > stm_cfg.txt
-
+	#
         # Copy N frames of a raw audio file to stm_in.raw. 
         dd bs=1280 count=100 if=../../../../raw/hts1.raw of=spch_in.raw \
-    	> setup.log 2>&1
-    freedv_tx 700E spch_in.raw stm_in.raw --testframes --txbpf 1 \
-    	>> setup.log 2>&1
-
+		> setup.log 2>&1
+	freedv_tx 700E spch_in.raw stm_in.raw --testframes --txbpf 0 \
+		>> setup.log 2>&1
+	#
         # Reference
         freedv_rx 700E stm_in.raw ref_demod.raw -v --testframes \
-    	> ref_gen.log 2>&1
-    ;;
+		> ref_gen.log 2>&1
+	;;
 
     700E_AWGN_test )
     	# Config is <mode>, <teswtframes> 
     	echo "81000010" > stm_cfg.txt
-
+	#
         # Copy N frames of a raw audio file to stm_in.raw. 
         dd bs=1280 count=96 if=../../../../raw/hts1.raw of=spch_in.raw \
-    	    > setup.log 2>&1
-        freedv_tx 700E spch_in.raw mod_bits.raw --testframes --txbpf 1 \
-        	  >> setup.log 2>&1
-        cohpsk_ch mod_bits.raw stm_in.raw -20 --Fs 8000 -f -5  2>&1 | tee setup.log
-
+		> setup.log 2>&1
+	freedv_tx 700E spch_in.raw mod_bits.raw --testframes --txbpf 0 \
+		  >> setup.log 2>&1
+	cohpsk_ch mod_bits.raw stm_in.raw -20 --Fs 8000 -f -5  2>&1 | tee setup.log
+	
         # Reference: - When the OFDM modem initially syncs, it often
-        #              has residual freq offset that causes abnormally
-        #              high LDPC decoder bit errors forthe first few
-        #              seconds.  This leads to a high coded BER being
-        #              reported for short duration tests.  This
-        #              settles down after a few seconds, and we get
-        #              the expected coded BER when averaging over
-        #              longer periods (e.g. 60s). However this
-        #              particular test is necessarily short due to the
-        #              slow speed of the semihosting system.  It is
-        #              therefore sufficient to check that the
-        #              performance is similar to the x86 C verison,
-        #              rather than expecting a low coded BER for a
-        #              short run.
-
+	#              has residual freq offset that causes abnormally
+	#              high LDPC decoder bit errors forthe first few
+	#              seconds.  This leads to a high coded BER being
+	#              reported for short duration tests.  This
+	#              settles down after a few seconds, and we get
+	#              the expected coded BER when averaging over
+	#              longer periods (e.g. 60s). However this
+	#              particular test is necessarily short due to the
+	#              slow speed of the semihosting system.  It is
+	#              therefore sufficient to check that the
+	#              performance is similar to the x86 C verison,
+	#              rather than expecting a low coded BER for a
+	#              short run.
+	
         freedv_rx 700E stm_in.raw ref_demod.raw -v --testframes  2>&1 --discard | tee ref_gen.log
-    ;;
+	;;
 
     700E_AWGN_codec )
     	# Config is <mode>, <teswtframes> 
     	echo "80000020" > stm_cfg.txt
-
+	#
         # Copy N frames of a raw audio file to stm_in.raw. 
         dd bs=1280 count=48 if=../../../../raw/hts1.raw of=spch_in.raw \
-    	    > setup.log 2>&1
-        freedv_tx 700E spch_in.raw mod_bits.raw --txbpf 1 \
-        	>> setup.log 2>&1
-
+		> setup.log 2>&1
+	freedv_tx 700E spch_in.raw mod_bits.raw --txbpf 0 \
+		>> setup.log 2>&1
+	#
         # Reference - give it a hard time with some noise to exercise the LDPC codec and get us to max CPU
-        cohpsk_ch mod_bits.raw stm_in.raw -20 --Fs 8000 -f -5  2>&1 | tee setup.log
-            freedv_rx 700E stm_in.raw ref_demod.raw -v \
-        	> ref_gen.log 2>&1
-    ;;
+	cohpsk_ch mod_bits.raw stm_in.raw -20 --Fs 8000 -f -5  2>&1 | tee setup.log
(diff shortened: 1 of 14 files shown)


drowe67:
Wow @tmiw - that's impressive! Well done :grinning: I guess it also shows the refactoring we've done over recent years is paying off, adding new modes across all platforms is getting easier ...

Couple of thoughts:
1. I still feel we need unit test(s) to support this.  It's one thing to have a functional test work in real time (or even OTA), quite another to show the performance is OK.  So easy for subtle bugs to slip in.  The current unit tests cover the basic OFDM modem, voice codec, and FreeDV API in 700D mode.  It would be good to test the FreeDV API with 700E signals over AWGN/fading channels, with data frames and voice.  
1. We should also measure the CPU load (I imagine it's less than 700E).
1. Re 700E (from the stm32 I presume?) being noisy - back in December I went to some effort to make this signal highly compressed, to minimise the PAPR.  The compressor works by first (Hilbert) clipping, then band pass filtering to remove some of the wideband noise the clipper introduces.  However that assumes the presence of a tx band pass filter, which is currently disabled for the stm32 due to CPU load.  It might be time to consider ways to run that filter (or a reduced CPU version) on the stm32.  Alternatively we could back off the compression, or just accept the noise as a limitation of the stm32 platform (the radios SSB filter will ultimately limit its extent).
1. We might also find that 700D is now a bit noisier on the stm32 - as I bumped up the compression for that mode back in December too.

tmiw:
@drowe67, think arm_fir_f32() could be used here? I took a quick look but there's some sort of filter tuning step that's needed too that I'm not sure about.

drowe67:
Yes possibly - hard to say if it would help (ie run faster).  A unit test to measure execution speed of the current filter might be useful, see how close that is to what we need.  In Tx, the modem CPU load is generally lighter, as we aren't running the LDPC decoder.  TBH I can't recall where the limitation was - it may have been memory (usually stack) at the time we initially did the 700D port.

tmiw:
@drowe67, I did some testing and enabling the BPF on the STM32 seemed to work okay for 700E. Out of an abundance of caution, though, I didn't enable it for 700D.

tmiw:
@drowe67, looks like bear went down while I was on it so I can't test the UT commits I made right now.

drowe67:
Sorry about that - I had to move it to make room for my day-job (WFH atm).  It's set up in another location now.

tmiw:
@drowe67, thanks! I'm seeing some intermittent failures in the 700E tests I added so I'm trying to figure out what's going on.

tmiw:
Hmm, it seems like the failures are pretty consistent. Something I'm noticing is that SNR seems to bounce between ~38dB and -5dB on tst_api_demod_700E_plain_test, so the target decode check ends up failing. The 700E demod tests also seem to take significantly longer to execute than the 700D versions for some reason. Any ideas would be appreciated, but I'll also keep looking.

tmiw:
Another note: OTA RX of 700E *is* working properly on the SM1000 itself in that the audio sounds correct. The only thing is that the Sync indicator flickers while in 700E when transmitting from the other radio (unlike with 700D and 1600). This may also be related to the UT issue.

tmiw:
@drowe67, turns out that nuwbits wasn't getting properly sync'd between struct freedv and the OFDM configuration (see https://github.com/drowe67/codec2/pull/206/commits/3a50957fc32eb2100e85fc95c6da7859444389ec). Fixing this fixed the failing unit tests as well as the flickering sync LED on the SM1000. 👍 

Anyway, all current tests are passing now:

```
Test project /home/mooneer/codec2/stm32/build_stm32
      Start  1: check_ram_limit
 1/28 Test  #1: check_ram_limit ..................   Passed    0.05 sec
      Start  2: tst_ldpc_enc
 2/28 Test  #2: tst_ldpc_enc .....................   Passed   14.92 sec
      Start  3: tst_ldpc_dec_ideal
 3/28 Test  #3: tst_ldpc_dec_ideal ...............   Passed   16.14 sec
      Start  4: tst_ldpc_dec_noise
 4/28 Test  #4: tst_ldpc_dec_noise ...............   Passed   20.57 sec
      Start  5: tst_ofdm_mod_plain
 5/28 Test  #5: tst_ofdm_mod_plain ...............   Passed   19.07 sec
      Start  6: tst_ofdm_mod_ldpc
 6/28 Test  #6: tst_ofdm_mod_ldpc ................   Passed   19.12 sec
      Start  7: tst_ofdm_demod_quick
 7/28 Test  #7: tst_ofdm_demod_quick .............   Passed   21.77 sec
      Start  8: tst_ofdm_demod_ideal
 8/28 Test  #8: tst_ofdm_demod_ideal .............   Passed   21.78 sec
      Start  9: tst_ofdm_demod_AWGN
 9/28 Test  #9: tst_ofdm_demod_AWGN ..............   Passed   28.86 sec
      Start 10: tst_ofdm_demod_fade
10/28 Test #10: tst_ofdm_demod_fade ..............   Passed   89.83 sec
      Start 11: tst_ofdm_demod_ldpc
11/28 Test #11: tst_ofdm_demod_ldpc ..............   Passed   17.70 sec
      Start 12: tst_ofdm_demod_ldpc_AWGN
12/28 Test #12: tst_ofdm_demod_ldpc_AWGN .........   Passed   39.65 sec
      Start 13: tst_ofdm_demod_ldpc_fade
13/28 Test #13: tst_ofdm_demod_ldpc_fade .........   Passed   63.11 sec
      Start 14: tst_codec2_enc_1300
14/28 Test #14: tst_codec2_enc_1300 ..............   Passed   35.16 sec
      Start 15: tst_codec2_enc_700C
15/28 Test #15: tst_codec2_enc_700C ..............   Passed   36.06 sec
      Start 16: tst_codec2_dec_1300
16/28 Test #16: tst_codec2_dec_1300 ..............   Passed   22.28 sec
      Start 17: tst_codec2_dec_700C
17/28 Test #17: tst_codec2_dec_700C ..............   Passed   21.48 sec
      Start 18: tst_api_mod_700D_TEST
18/28 Test #18: tst_api_mod_700D_TEST ............   Passed   30.74 sec
      Start 19: tst_api_mod_700D_CODEC
19/28 Test #19: tst_api_mod_700D_CODEC ...........   Passed   30.77 sec
      Start 20: tst_api_mod_700E_TEST
20/28 Test #20: tst_api_mod_700E_TEST ............   Passed   36.71 sec
      Start 21: tst_api_mod_700E_CODEC
21/28 Test #21: tst_api_mod_700E_CODEC ...........   Passed   36.70 sec
      Start 22: tst_api_demod_700D_plain_test
22/28 Test #22: tst_api_demod_700D_plain_test ....   Passed   59.46 sec
      Start 23: tst_api_demod_700D_AWGN_test
23/28 Test #23: tst_api_demod_700D_AWGN_test .....   Passed   60.14 sec
      Start 24: tst_api_demod_700D_AWGN_codec
24/28 Test #24: tst_api_demod_700D_AWGN_codec ....   Passed   52.22 sec
      Start 25: tst_api_demod_700E_plain_test
25/28 Test #25: tst_api_demod_700E_plain_test ....   Passed   92.30 sec
      Start 26: tst_api_demod_700E_AWGN_test
26/28 Test #26: tst_api_demod_700E_AWGN_test .....   Passed   94.89 sec
      Start 27: tst_api_demod_700E_AWGN_codec
27/28 Test #27: tst_api_demod_700E_AWGN_codec ....   Passed   79.24 sec
      Start 28: tst_api_demod_1600_plain_codec
28/28 Test #28: tst_api_demod_1600_plain_codec ...   Passed  165.60 sec

100% tests passed, 0 tests failed out of 28

Total Test time (real) = 1226.39 sec
```

The 700E tests do take longer to run than the 700D ones but the audio quality does sound fine on OTA testing, so it might not be that important.

tmiw:
OK, after fixing the Python errors I had noticed in the API demod tests, I get one failure in tst_api_demod_700E_AWGN_test due to the BER being out of bounds:

```
Check reference decode
REF   uncoded BER: 0.211900 coded BER: 0.211300
BAD
Check target decode
TARGET uncoded BER: 0.211900 coded BER: 0.211300
BAD

Test FAILED!
```

(uncoded > 0.1, coded > 0.01)

However, since both ref and target are equal to each other, perhaps the thresholds for that test should be raised?

drowe67:
Great progress @tmiw !

Suggest reducing the SNR (lowering the noise level), until you get the coded BER less that 1%, so it's tested at a realistic operating point.  The 700E LDPC code will fall over at a raw BERs of around 8% (it's a smaller code than 700D so not as effective).

stm32/unittest/scripts/tst_api_demod_setup
```diff
@@ -85,6 +85,66 @@ case "${TEST_OPT}" in
...
+	#              settles down after a few seconds, and we get
+	#              the expected coded BER when averaging over
+	#              longer periods (e.g. 60s). However this
+	#              particular test is necessarily short due to the
+	#              slow speed of the semihosting system.  It is
+	#              therefore sufficient to check that the
+	#              performance is similar to the x86 C verison,
+	#              rather than expecting a low coded BER for a
+	#              short run.
+
```
drowe67, line 128: :point_up: this text refers to 700D, 700E can sync much faster - so we should see good coded BER results here

stm32/unittest/scripts/tst_api_demod_setup
```diff
@@ -85,6 +85,66 @@ case "${TEST_OPT}" in
...
+    	echo "80000020" > stm_cfg.txt
+	#
+        # Copy N frames of a raw audio file to stm_in.raw. 
+        dd bs=1280 count=48 if=../../../../raw/hts1.raw of=spch_in.raw \
+		> setup.log 2>&1
+	freedv_tx 700E spch_in.raw mod_bits.raw --txbpf 1 \
+		>> setup.log 2>&1
+	#
+        # Reference - give it a hard time with some noise to exercise the LDPC codec and get us to max CPU
+	cohpsk_ch mod_bits.raw stm_in.raw -20 --Fs 8000 -f -5  2>&1 | tee setup.log
```
drowe67, line 143: Take a look at `test_OFDM_modem_700E_AWGN` in the x86 ctests, I think we use a Noise argument of -22 for 700E, ie a few dB higher to account for the smaller code, and higher overhead to due  increased pilot symbol rate.

drowe67:
@tmiw Re https://github.com/drowe67/codec2/pull/206/commits/954beae5e5d2c68c52c6fc52bbed41c0236b87b6 is 700E in the unit tests and SM1000 running with the clipper and tx BPF enabled?

drowe67:
A described in https://github.com/drowe67/codec2/pull/42, we can measure the CPU load by probing pin PE3 with an oscilloscope. This is toggled in sm1000_main.c by lines containing `GPIOE->ODR = (1 << 3)`

It would useful to make a table here with the on period/on time/CPU load estimate during Tx/Rx for 700D/700E, with and without the TX BPF.  I'll pop a template at the top of this PR.

tmiw:
> @tmiw Re [954beae](https://github.com/drowe67/codec2/commit/954beae5e5d2c68c52c6fc52bbed41c0236b87b6) is 700E in the unit tests and SM1000 running with the clipper and tx BPF enabled?

The BPF is on, but I don't think the clipper is (per the defaults in ofdm_mode.c).



> A described in #42, we can measure the CPU load by probing pin PE3 with an oscilloscope. This is toggled in sm1000_main.c by lines containing `GPIOE->ODR = (1 << 3)`
> 
> It would useful to make a table here with the on period/on time/CPU load estimate during Tx/Rx for 700D/700E, with and without the TX BPF. I'll pop a template at the top of this PR. Feel free to modify the template.

I'll have to figure out how to connect everything together here but if you get to it before me, go for it.

stm32/unittest/scripts/tst_api_demod_setup
tmiw: Yep, looks like -22 works. 👍

stm32/unittest/scripts/tst_api_demod_setup
tmiw: Yep:

```
26: Breakpoint 1, Reset_Handler () at /home/mooneer/codec2/stm32/unittest/src/startup_stm32f4xx.s:119
26: 119	  bl  .
26: shutdown command invoked
26: shutdown command invoked
26: Remote connection closed
26: Check reference decode
26: REF   uncoded BER: 0.075200 coded BER: 0.018300
26: OK
26: Check target decode
26: TARGET uncoded BER: 0.074000 coded BER: 0.019100
26: OK
26: 
26: Test PASSED
26: 
26: Test tst_api_demod_700E_AWGN_test PASSED
1/1 Test #26: tst_api_demod_700E_AWGN_test .....   Passed   92.90 sec
```

tmiw:
FWIW, I just enabled clipping along with TXBPF and the test times don't seem to be that much longer:

```
mooneer@bear:~/codec2/stm32/build_stm32$ ctest -R tst_api_mod_700E
Test project /home/mooneer/codec2/stm32/build_stm32
    Start 20: tst_api_mod_700E_TEST
1/2 Test #20: tst_api_mod_700E_TEST ............   Passed   36.77 sec
    Start 21: tst_api_mod_700E_CODEC
2/2 Test #21: tst_api_mod_700E_CODEC ...........   Passed   36.72 sec

100% tests passed, 0 tests failed out of 2

Total Test time (real) =  73.50 sec
```

drowe67:
> It looks like it's nuwbits that's off based on the spreadsheet and the assertion:

Actually there may be a bug in that expression. The idea is that we have (ns-1)*nc data symbols per frame, which we use for the LDPC codeword, UW bits, and txt bits.  They all have to add up neatly to fit into the frame. So for example you can compute the # txt bits of # UW bits as the left over after everything else is subtracted.

However I note that expression above doesn't have a `*nc` (number of carriers) term, or a term for the LDPC codeword symbols ...  I'm not even sure how it's even working for 700D.  So this is probably a bug that your fine work has exposed :thinking: 

Your solution is fine for now.  I'll poke about and work out why the sums aren't correct :slightly_smiling_face:

tmiw:
OK, I have my SM1000 taken apart and I got a few readings, but they don't seem right based on what I understand about how that pin's supposed to be triggered:

*700E RX*:

IMG_20210809_222242

*700E TX BPF + Clipping*:

IMG_20210809_222422

*700D RX*:

IMG_20210809_222617

*700D TX*:

IMG_20210809_222723

Is it supposed to be elsewhere on revision D? I was probing P8 on the back per the schematic that I have.

drowe67:
@tmiw those plots don't like quite right - I'd expect to see a square-ish wave about 3v3 in amplitude. However it's been a few years since I've done this myself.  The net you need to probe is connected to pin 2 of the chip so, you could check that with a multimeter.  Was the scope DC coupled, good ground connection?  I can see some spikes in roughly the right places.  I think you're close ....

If you get stuck I can try it here later this week, but busy with the day job right now ...

tmiw:
OK, turns out that the mounting holes are *not* grounds, which explains why I was having issues the other night.

New plots (directly from the scope, too, for better quality):

_700D RX_:

700D RX

_700D TX_:

700D TX

_700D TX BPF + Clip_:

700D TX BPF

_700E RX_:

700E RX

_700E TX_:

700E TX

_700E TX BPF + Clip_:

700E TX BPF

I updated the table in OP as well.

drowe67:
Great work @tmiw :+1:  :grinning: 

With the Rx results, was the CPU load higher in acquisition (just listening to noise or an invalid signal), or when synced up and decoding audio?

Interesting that Tx CPU load higher, that's kind of surprising.  Usually Rx is heavier, at least from the modem point of view.  I can't recall if the codec CPU load is heavier in Tx or Rx.

Good news that we can run the Tx BPF in both modes!

srsampson:
Something fishy about that 700D RX waveform data (for example). It says 52%/47% duty cycle, but the 1914us/223.2us tVmax/tVmin.

But anyway, it looks like 1.9ms to 1.6ms max which is way less than 160ms.

tmiw:
> Something fishy about that 700D RX waveform data (for example). It says 52%/47% duty cycle, but the 1914us/223.2us tVmax/tVmin.
> 
> But anyway, it looks like 1.9ms to 1.6ms max which is way less than 160ms.

My understanding was that +Wid/period = CPU load (as PE3 = 1 during freedv_rx/freedv_tx), but I could be mistaken.

drowe67:
Yep it looks like PE3 is high (3v3) when we are performing core FreeDV API processing.  So 100% of our time spent in FreeDV processing would mean PE3 stuck at 3v3.  There are also some "background" processes such as interrupt service routines and other main loop code - so in practice we can't use all of our CPU for FreeDV processing.  The time spent in the ISRs can be measured by looking at PE0 (ADC ISR) and PE1 (DAC ISR).

I'm inclined to think the small variations in measured period (159.5, 158, 81) are measurement errors.  It could be due to when the PE3 GPIO is set/reset - it's not a hard real time process so there could be some cycle/cycle jitter.

If the CPU is truly overloaded on Tx, the transmit modem signal would be messed up, and a receiving station would not be able to get sync.

If the FreeDV part of Tx is taking too long, then we would run out of samples for the DAC. We have some counters to detect DAC FIFO underflow/overflow (e.g. `dac_underflow` in `stm32f4_dac.c`) but are not using them at the moment. There is similar logic to detect ADC FIFO overflows, also not connected.  This is probably OK - as I said CPU overload would be very obvioius when the SM1000 is used, and the scope plots above show plenty of margin.

tmiw:
> If the CPU is truly overloaded on Tx (or the frame period is out by 1ms), the transmit modem signal would be messed up, and a receiving station would not be able to get sync. The freedv-gui Timing plot would show very big timing offsets.
> 
> If the FreeDV part of Tx is taking too long, then we would run out of samples for the DAC. We have some counters to detect DAC FIFO underflow/overflow (e.g. `dac_underflow` in `stm32f4_dac.c`) but are not using them at the moment. There is similar logic to detect ADC FIFO overflows, also not connected. This is probably OK - as I said CPU overload would be very obvioius when the SM1000 is used, and the scope plots above show plenty of margin.

Indeed, I just retried RX and TX on both 700D and E and both sounded okay to me. 👍

drowe67:
@tmiw - shall we distribute the v5 binary to some beta testers?  I usually post development version to the codec 2 mailing list.  I'm happy for you to make the post, or I can if you like.

tmiw:
> @tmiw - shall we distribute the v5 binary to some beta testers? I usually post development version to the codec 2 mailing list. I'm happy for you to make the post, or I can if you like.

Go for it!

tmiw:
So I found one issue. It seems that the Back button won't go from Analog to 700E. I'm not sure if it's an issue that existed prior to this PR or not, though.

drowe67:
Good work on the menu fix @tmiw :+1: Lets give this one a few days to gather feedback from beta testers, then we can merge I think.  You've certainly become an expert at many areas of codec2 and freedv-gui development :slightly_smiling_face:

drowe67:
@tmiw I haven't had any bug reports from beta testers so am inclined to merge this, what do you think?

I can post the latest binary with your recent menu bug fixes.

tmiw:
@drowe67, go for it. 👍

PR #209, opened 2021-08-16
Add reliable_text library to codec2 to enable PSK Reporter support.

tmiw:
Porting work done in freedv-gui on a reliable text library to codec2 for testing and refinement. Currently targeting full decode +2dB above the minimum SNR for voice per README_freedv.md:

* 1600: 6dB
* 700D: 0dB
* 700E: 3dB
* 2020: 4dB

The current work doesn't meet that goal thus far except with 1600:

```
test 69
    Start 69: test_freedv_reliable_text_awgn_1600

69: Test command: /bin/sh "-c" "cd /Users/mooneer/devel/freedv-gui/codec2/build_osx/src;
                        ./freedv_tx 1600 ../../raw/ve9qrp_10s.raw - --reliabletext AB1CDEF | ./cohpsk_ch - - -27 --Fs 8000 -f -5 > 1600_reliable_raw;
                        ./freedv_rx 1600 1600_reliable_raw /dev/null --txtrx 1600_reliable.txt --reliabletext;
                        cat 1600_reliable.txt"
69: Test timeout computed to be: 1500
69: cohpsk_ch: Fs: 8000 NodB: -27.00 foff: -5.00 Hz fading: 0 nhfdelay: 0 clip: 32767.00 ssbfilt: 1 complexout: 0
69: cohpsk_ch: SNR3k(dB):     5.74  C/No....:    40.51
69: cohpsk_ch: peak.....: 14959.54  RMS.....:  4726.79   CPAPR.....: 10.01 
69: cohpsk_ch: Nsamples.:    80000  clipped.:     0.00%  OutClipped:  0.00%
69: frames decoded: 502  output speech samples: 76160
69: AB1CDEF
1/3 Test #69: test_freedv_reliable_text_awgn_1600 ...   Passed    0.09 sec
test 70
    Start 70: test_freedv_reliable_text_awgn_700D

70: Test command: /bin/sh "-c" "cd /Users/mooneer/devel/freedv-gui/codec2/build_osx/src;
                     ./freedv_tx 700D ../../raw/ve9qrp_10s.raw - --reliabletext AB1CDEF | ./cohpsk_ch - - -18 --Fs 8000 -f -5 > 700D_reliable_raw;
                     ./freedv_rx 700D 700D_reliable_raw /dev/null --txtrx 700D_reliable.txt --reliabletext;
                     cat 700D_reliable.txt"
70: Test timeout computed to be: 1500
70: cohpsk_ch: Fs: 8000 NodB: -18.00 foff: -5.00 Hz fading: 0 nhfdelay: 0 clip: 32767.00 ssbfilt: 1 complexout: 0
70: cohpsk_ch: SNR3k(dB):     0.22  C/No....:    34.99
70: cohpsk_ch: peak.....: 16745.60  RMS.....:  7055.41   CPAPR.....:  7.51 
70: cohpsk_ch: Nsamples.:    79360  clipped.:     0.00%  OutClipped:  0.00%
70: /bin/sh: line 2: 88988 Illegal instruction: 4  ./freedv_rx 700D 700D_reliable_raw /dev/null --txtrx 700D_reliable.txt --reliabletext
2/3 Test #70: test_freedv_reliable_text_awgn_700D ...***Failed  Required regular expression not found. Regex=[AB1CDEF
]  0.10 sec
test 71
    Start 71: test_freedv_reliable_text_awgn_700E

71: Test command: /bin/sh "-c" "cd /Users/mooneer/devel/freedv-gui/codec2/build_osx/src;
                      ./freedv_tx 700E ../../raw/ve9qrp_10s.raw - --reliabletext AB1CDEF | ./cohpsk_ch - - -22 --Fs 8000 -f -5 > 700E_reliable_raw;
                      ./freedv_rx 700E 700E_reliable_raw /dev/null --txtrx 700E_reliable.txt --reliabletext;
                      cat 700E_reliable.txt"
71: Test timeout computed to be: 1500
71: cohpsk_ch: Fs: 8000 NodB: -22.00 foff: -5.00 Hz fading: 0 nhfdelay: 0 clip: 32767.00 ssbfilt: 1 complexout: 0
71: cohpsk_ch: SNR3k(dB):     3.25  C/No....:    38.02
71: cohpsk_ch: peak.....: 17200.10  RMS.....:  6309.85   CPAPR.....:  8.71 
71: cohpsk_ch: Nsamples.:    80000  clipped.:     0.00%  OutClipped:  0.00%
71: frames decoded: 125  output speech samples: 80000
3/3 Test #71: test_freedv_reliable_text_awgn_700E ...***Failed  Required regular expression not found. Regex=[AB1CDEF
]  0.11 sec

The following tests passed:
	test_freedv_reliable_text_awgn_1600

33% tests passed, 2 tests failed out of 3

Total Test time (real) =   0.31 sec

The following tests FAILED:
	 70 - test_freedv_reliable_text_awgn_700D (Failed)
	 71 - test_freedv_reliable_text_awgn_700E (Failed)
Errors while running CTest
Output from these tests are in: /Users/mooneer/devel/freedv-gui/codec2/build_osx/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```


tmiw:
@drowe67, looks like the minimum SWR required with AWGN is as follows:

* 700D: 4dB (-22 to cohpsk_ch)
* 700E: 8dB (-27 to cohpsk_ch)

One possibility is that it's something to do with varicode. I don't recall how far down it normally works with PSK31, so maybe we're asking too much of it? But still, I'd imagine Golay would compensate somewhat.

drowe67:
Good work @tmiw, looks like a nice test framework :+1: 

Looking at the test output I'm not clear on:
1. How many test PSK reporter packets were sent?
1. What was the PER of the packets?

Sorry if I missed something (I haven't dug into the code yet) ...

Varicode is pretty sensitive to bit errors, but will eventually recover from a bit error.  However that's probably not good enough for a data protocol where no errors are allowed in the decoded string.

So my understanding is you have:
1. Unprotected "raw" bits from the demodulator, ie 2 bits/frame
1. These are protected by a Golay code
1. which after Golay decoding are passed to the Varicode decoder.

Some ideas:

1. I'm wondering what bit error rate we have on those raw bits from the demodulator at your operating point?  This could be measured by running the freedv_tx/freedv_tx test in test frame mode (e.g. many ctests do this).  
1. Is that BER something a Golay code can handle?  IIRC Golay is good for a few % BER. You could test this experimentally too, e.g. measure the coded BER through the Golay protected txt channel.
1. If the raw BER is so high the Golay FEC is falling over at the operating point you could try your original idea of using a LDPC code for the PSK reporter packets. Or it might just be a bug somewhere, e.g. the Golay decoding not working right.
1. On a clean (high SNR channel), try inserting a few it errors in the Golay codeword that makes up the PSK reporter packet, make sure its correcting them.
1. Try raising the SNR until the tests do pass, determine how close you are to your goal.

tmiw:
@drowe67, to answer your questions:

> 1. How many test PSK reporter packets were sent?

One 7 character callsign was able to be sent and decoded in the 10 second transmission.

> 2. What was the PER of the packets?

I added some fprintfs in reliable_text.c to track what was going on with the Golay encoded bytes being sent and received via the freedv_api text callbacks and from what I can tell, the error rate was something on the order of 25% (e.g. ~2 bits *per byte* erroring, not just 2 per block) for the unit test 700D.

As a sanity check, I tried to bring 1600 down even further and it seems I was still able to decode the callsign as low as ~3dB SNR (cohpsk_ch -24):

```
MacBook-Pro:src mooneer$ ./tx 1600 ../../raw/ve9qrp_10s.raw - --reliabletext AB1CDEF | ./cohpsk_ch - - -24 --Fs 8000 -f -5 > 1600_reliable_raw
cohpsk_ch: Fs: 8000 NodB: -24.00 foff: -5.00 Hz fading: 0 nhfdelay: 0 clip: 32767.00 ssbfilt: 1 complexout: 0
sent char: 7
sent char: 8
sent char: 16
sent char: 48
sent char: 10
sent char: 10
sent char: 33
sent char: 10
sent char: 5
sent char: 43
sent char: 3
sent char: 50
sent char: 11
sent char: 44
sent char: 2
sent char: 54
sent char: 12
sent char: 32
sent char: 9
sent char: 37
sent char: 20
sent char: 0
sent char: 0
sent char: 0
cohpsk_ch: SNR3k(dB):     2.74  C/No....:    37.51
cohpsk_ch: peak.....: 14959.54  RMS.....:  4726.79   CPAPR.....: 10.01 
cohpsk_ch: Nsamples.:    80000  clipped.:     0.00%  OutClipped:  0.00%
MacBook-Pro:src mooneer$ ./freedv_rx 1600 1600_reliable_raw /dev/null --txtrx 1600_reliable.txt --reliabletext
received char: 7
received char: 8
received char: 16
received char: 48
received char: 10
received char: 10
received char: 33
received char: 10
received char: 5
received char: 43
received char: 3
received char: 50
received char: 11
received char: 44
received char: 2
received char: 54
received char: 12
received char: 32
received char: 9
received char: 37
received char: 20
received char: 0
received char: 0
frames decoded: 502  output speech samples: 74560
MacBook-Pro:src mooneer$ cat 1600_reliable.txt 
AB1CDEFMacBook-Pro:src mooneer$
```

That's about 1dB under what 1600 itself is supposed to work to, so I'm not sure if there's still headroom. It *does* kinda make me think there's some sort of issue in how 700D/E handle the text field, though, but we'll need to investigate further to rule that out.

BTW reliable_text.c works on the bytes after the underlying mode does the varicode encode/decode (as in, taking the Golay encoded bytes and translating them to/from the 10-bit varicode that ultimately goes out over the air).

tmiw:
@drowe67, based on [how QPSK is supposed to work](https://en.wikipedia.org/wiki/Phase-shift_keying#Quadrature_phase-shift_keying_(QPSK)_, shouldn't there be a 1 + i somewhere below in ofdm.c:

```
static const complex float qpsk[] = {
    1.0f + 0.0f * I,
    0.0f + 1.0f * I,
    0.0f - 1.0f * I,
    -1.0f + 0.0f * I
};
```

srsampson:
The table is in Gray code form (only one bit changes in each entry). In Java, when I'm really verbose, it would be something like:

```
    /*
     * QPSK Quadrant bit-pair values - Gray Coded
     */
    private final static Complex[] CONSTELLATION = new Complex[]{
        ComplexMath.cexp(new Complex(0.0, Math.toRadians(0.0))),   //  1 or  1+0i
        ComplexMath.cexp(new Complex(0.0, Math.toRadians(90.0))),  //  j or  0+1i or +90 BPSK
        ComplexMath.cexp(new Complex(0.0, Math.toRadians(270.0))), // -j or  0-1i
        ComplexMath.cexp(new Complex(0.0, Math.toRadians(180.0)))  // -1 or -1+0i or -90 BPSK
    };
```
The C99 Complex format just uses the "imaginary * I" syntax.

srsampson:
Hmm, you're right. That second example doesn't look gray coded. But the symbol is rotated 45 degrees I believe.

Poor docs anyway! Who wrote that crap :-) ha

srsampson:
The second comment, I believe was to show the dibit pattern each quadrant will have.

While the constellation would be more like:
```
           0 + 1i
               |
-1 + 0i ---+---  1 + 0i
               |
            0 - 1i
```
and for straightcoding we select the constellation values by:
```
00 = 1 + 0i
01 = 0 + 1i
10 = -1 + 0i
11 = 0 - 1i
```
But for gray code, we swap the last two, in order to change by only one bit:
```
00 = 1 + 0i
01 = 0 + 1i
11 = 0 - 1i
10 = -1 + 0i
```
Which is what is in the Octave code.  Sorry for the edits.

tmiw:
Yeah, I actually played with that and all I was doing was making it worse :) I also played with creating a new varicode format that effectively just encodes the raw value + some separator bits (in case it was something with the existing varicode format that wasn't helping) but I don't know if that really did anything.

Anyway, here's a sample run with 700E and cohpsk_ch -19 (resulting in an SNR of ~5dB). There are just enough bits changing that I'm not sure even LDPC would help much:

| Tx(Dec) | Tx(Bin)  | Rx(Dec) | Rx(Bin)  | XOR(Rx, Tx) | BitsChanged |
|---------|----------|---------|----------|-------------|-------------|
|       7 | 00000111 |       7 | 00000111 | 00000000    |           0 |
|       8 | 00001000 |       8 | 00001000 | 00000000    |           0 |
|      16 | 00010000 |       8 | 00001000 | 00011000    |           2 |
|      48 | 00110000 |      53 | 00110101 | 00000101    |           2 |
|      10 | 00001010 |      10 | 00001010 | 00000000    |           0 |
|      10 | 00001010 |      52 | 00110100 | 00111110    |           5 |
|      33 | 00100001 |      37 | 00100101 | 00000100    |           1 |
|      10 | 00001010 |      32 | 00100000 | 00101010    |           3 |
|       5 | 00000101 |       5 | 00000101 | 00000000    |           0 |
|      43 | 00101011 |      43 | 00101011 | 00000000    |           0 |
|       3 | 00000011 |       3 | 00000011 | 00000000    |           0 |
|      50 | 00110010 |      50 | 00110010 | 00000000    |           0 |
|      11 | 00001011 |      57 | 00111001 | 00110010    |           3 |
|      44 | 00101100 |      58 | 00111010 | 00010110    |           3 |
|       2 | 00000010 |       2 | 00000010 | 00000000    |           0 |
|      54 | 00110110 |      55 | 00110111 | 00000001    |           1 |
|      12 | 00001100 |      32 | 00100000 | 00101100    |           3 |
|      32 | 00100000 |       9 | 00001001 | 00101001    |           3 |
|       9 | 00001001 |      37 | 00100101 | 00101100    |           3 |
|      37 | 00100101 |      20 | 00010100 | 00110001    |           3 |
|      20 | 00010100 |       8 | 00001000 | 00011100    |           3 |
|       0 | 00000000 |       0 | 00000000 | 00000000    |           0 |
|       0 | 00000000 |       0 | 00000000 | 00000000    |           0 |
|       0 | 00000000 |       0 | 00000000 | 00000000    |           0 |

Also, at -20, the test can't get the text from the 10s VE9QRP sample using 700E but can from a longer one. That seems to be about the lowest possible SNR where we can get text at some point during the transmission.

drowe67:
> One 7 character callsign was able to be sent and decoded in the 10 second transmission.

It's best to measure PER over many PSK reporter packets, say 20-100, same with uncoded and coded BER.  Looking at individual bytes & bit errors can be misleading due to statistical variations.  Also 700D/E tend to have a higher BER just after sync as the estimators settle down.

You can determine what the uncoded (raw) BER should be by running freedv_tx/rx in test frame mode (many examples in the ctests).  Over many bits, we should get the same uncoded (raw) BER for the txt bits at a given operating point.

> > 1. What was the PER of the packets?

PER is the number of error free packets received/number received.  So you need a number of complete PSK reporter packets to measure this.

tmiw:
Hopefully the below at least partially answers the question. I used the following command (+ a file basically containing ve9qrp.raw repeated a fair number of times):

```
./freedv_tx 700E ve9qrp_multi.raw - --clip 1 --txbpf 1 --reliabletext AB1CDEF 2>/dev/null | ./cohpsk_ch - - [cohpsk_ch parameter] --Fs 8000 -f -5 > 700E_reliable.raw &&
./freedv_rx 700E 700E_reliable.raw /dev/null --txtrx 700E_reliable.txt --reliabletext 2>/dev/null && 
grep "AB1CDEF" 700E_reliable.txt | wc -l
```


| cohpsk_ch parameter | cohpsk_ch SNR | Number of valid callsigns received | Error rate vs. first
|---------------------|---------------|------------------------------------|--------------|
| -22                 | 8.01          | 5067                               | N/A |
| -21                 | 6.98          | 5066                               |  ~0% |
| -20                 | 6.02          | 4933                               |  2.6% |
| -19                 | 5.01          | 0                                  | 100% |

Maybe I'll look into using LDPC anyway just to see if there's much of a difference. Even 1dB improvement would be good.

tmiw:
With LDPC(112,56):

| cohpsk_ch parameter | cohpsk_ch SNR | Number of valid callsigns received | Error rate vs. first
|---------------------|---------------|------------------------------------|--------------|
| -22                 | 8.01          | 160                               | N/A |
| -21                 | 6.98          | 108                               |  33.0% |
| -20                 | 6.02          | 46                               |  71.3% |
| -19                 | 5.01          | 14                                  | 91.3% |
| -18                 | 3.98          | 0                                  | 100% |

About ~1dB better as I suspected. Note that I fixed a bug in how the byte queuing works, too, so I'll need to rerun with Golay to get comparable numbers.

tmiw:
In the LDPC version, I had it only return the callsign if BER < 0.1 (per example from ldpc_dec.c). However, bumping that up to 0.2 gave me the following:

| cohpsk_ch parameter | cohpsk_ch SNR | Number of valid callsigns received | Error rate vs. first
|---------------------|---------------|------------------------------------|--------------|
| -22                 | 8.01          | 5414                               | N/A |
| -21                 | 6.98          | 5402                               |  ~0% |
| -20                 | 6.02          | 5480\*                               |  -1.2% |
| -19                 | 5.01          | 5216                                  | 3.7% |
| -18                 | 3.98          | 4163                                  | 23.1% |
| -17                 | 3.02          | 609                                  | 88.8% |
| -16                 | 2.01         | 0                                  | 100% |

I'm not sure if 0.2 is a good idea, though, but it does get us closer to the 2dB over minimum sync goal. Perhaps we can come up with a cutoff that's somewhere in between, too.

tmiw:
I did some more investigating and it turns out that the 5000+ numbers aren't actually correct after all. Considering that it should take ~11 seconds or so to transmit one block, the true number is more like in the 200s range for the "ideal" signal (without AWGN), assuming a ~37 or so minute transmission as I've been using. After fixing it for real, it turns out that this isn't as great as initially thought.

OTOH, however, perhaps the goal should be whether a callsign will go over during a typical transmission. I'd say those are maybe a couple of minutes *tops* per "over", so would a decode rate of 1 callsign/minute be adequate you guys think? New measurements below:

_700E_:

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.98 |             14.98 |                                214 |                  0.00% |   5.709204091 |
|                 -29 |         15.02 |             14.02 |                                214 |                  0.00% |   5.709204091 |
|                 -28 |         14.01 |             13.01 |                                214 |                  0.00% |   5.709204091 |
|                 -27 |         12.98 |             11.98 |                                214 |                  0.00% |   5.709204091 |
|                 -26 |         12.02 |             11.02 |                                215 |                 -0.47% |   5.735882614 |
|                 -25 |         11.01 |             10.01 |                                229 |                 -7.01% |   6.109381948 |
|                 -24 |          9.98 |              8.98 |                                218 |                 -1.87% |   5.815918186 |
|                 -23 |          9.02 |              8.02 |                                210 |                  1.87% |   5.602489996 |
|                 -22 |          8.01 |              7.01 |                                170 |                 20.56% |   4.535349044 |
|                 -21 |          6.98 |              5.98 |                                121 |                 43.46% |   3.228101378 |
|                 -20 |          6.02 |              5.02 |                                 64 |                 70.09% |   1.707425522 |
|                 -19 |          5.01 |              4.01 |                                 27 |                 87.38% |  0.7203201423 |
|                 -18 |          3.98 |              2.98 |                                  3 |                 98.60% | 0.08003557137 |
|                 -17 |          3.02 |              2.02 |                                  1 |                 99.53% | 0.02667852379 |
|                 -16 |          2.01 |              1.01 |                                  0 |                100.00% |             0 |

_700D_:

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.75 |             17.75 |                                214 |                  0.00% |   5.709204091 |
|                 -29 |          14.8 |              16.8 |                                214 |                  0.00% |   5.709204091 |
|                 -28 |         13.78 |             15.78 |                                214 |                  0.00% |   5.709204091 |
|                 -27 |         12.75 |             14.75 |                                214 |                  0.00% |   5.709204091 |
|                 -26 |          11.8 |              13.8 |                                214 |                  0.00% |   5.709204091 |
|                 -25 |         10.78 |             12.78 |                                213 |                  0.47% |   5.682525567 |
|                 -24 |          9.75 |             11.75 |                                229 |                 -7.01% |   6.109381948 |
|                 -23 |           8.8 |              10.8 |                                229 |                 -7.01% |   6.109381948 |
|                 -22 |          7.78 |              9.78 |                                237 |                -10.75% |   6.322810138 |
|                 -21 |          6.75 |              8.75 |                                217 |                 -1.40% |   5.789239662 |
|                 -20 |           5.8 |               7.8 |                                188 |                 12.15% |   5.015562472 |
|                 -19 |          4.78 |              6.78 |                                144 |                 32.71% |   3.841707426 |
|                 -18 |          3.75 |              5.75 |                                 94 |                 56.07% |   2.507781236 |
|                 -17 |           2.8 |               4.8 |                                 43 |                 79.91% |   1.147176523 |
|                 -16 |          1.78 |              3.78 |                                 14 |                 93.46% |   0.373499333 |
|                 -15 |          0.75 |              2.75 |                                  3 |                 98.60% | 0.08003557137 |
|                 -14 |          -0.2 |               1.8 |                                  0 |                100.00% |             0 |

I'm still not sure what's up with the rows that have more decoded callsigns than ideal, though. That will require more investigation for sure.

drowe67:
Nice work @tmiw :+1: - the PERs you have obtained look rather high at those SNRs, so I think we are getting closer to finding the issue.

So here's the OFDM modem running in test frame mode, as I mentioned above.  We can use this as a simple sanity check to compare:
```
./freedv_tx 700D ../../raw/ve9qrp.raw - --testframes --clip 1 --txbpf 1 | ./cohpsk_ch - - -18 --Fs 8000 | ./freedv_rx 700D - /dev/null --testframes
cohpsk_ch: SNR3k(dB):     3.81
frames decoded: 702  output speech samples: 2560
BER......: 0.0090  Tbits:   163800  Terrs:     1481
```
So the OFDM modem is delivering bits at just less that 1% (raw) BER at 3.81dB SNR.  A Golay code should correct that BER no problem, as the chances of >3 errors in a 23 bit codeword is pretty small. So coded bit error rate of the PSK reporter bits (after Golay decoding) should be about 0, and the PSK reporter PER should be about 0.

My probability is a bit weak but I think this is modelled by the binomial distribution:
```
octave:5> binopdf(3, 23, 0.01) 
ans =  0.0014485
octave:6> binopdf(4, 23, 0.01) 
ans =  0.000073157
```
So you chances of 4 bit errors in a Golay codeword is pretty small, given a BER of 0.01.

Couple of ideas:
1. How are you getting frame sync on the PSK Reporter frames in the txt bit stream?  If you're using a unique word for frame sync this post shows how you can compute the right size https://www.rowetel.com/?p=7467
1. Compare the raw BER of *just* the txt bits to the above, make sure it's about the same.
1. Inject some bit errors in the PSK Reporter bit stream (say 1 in 100) and see  what happens.  This can be a unit test that doesn't involve the modems at all, just have a PSK Reporter Tx -> xor in bit errors -> PSK Reporter Rx.  Make sure it can get  frame sync with random timing offsets, doesn't get false sync too often etc, and can correct errors up to a few % BER.
1. Once we get it working at low SNRs, an interleaver might be a good idea.  This will spread any error bursts over the entire PSK reporter frame, giving us a better chance of correcting them.  This will help with fading channels.  The PSK reporter frames have the advantage of being quite long (unlike voice), so we can "ride over" fades.

src/reliable_text.c
```diff
@@ -0,0 +1,346 @@
...
+    unsigned char output[LDPC_TOTAL_SIZE_BITS];
+    int parityCheckCount = 0;
+    
+    for (int bitIndex = 0; bitIndex < LDPC_TOTAL_SIZE_BITS; bitIndex++)
+    {
+        // Map to value expected by sd_to_llr().
+        int bit = (src[bitIndex / 7] & (1 << (bitIndex % 7))) != 0;
+        incomingData[bitIndex] = 1.0 - 2.0 * bit;
+    }
+
```
drowe67, line 161: This is hard decision decoding, so you will lose much of the power of the LDPC code......

drowe67:
> OTOH, however, perhaps the goal should be whether a callsign will go over during a typical transmission. I'd say those are maybe a couple of minutes _tops_ per "over", so would a decode rate of 1 callsign/minute be adequate you guys think?

Yes I agree, we have a similar requirement for the High Altitude Balloon telemetry (like project Horus) - you just need to get the packet once every now and again, as its the same info being repeated all the time (e.g. the balloons GPS coords).  That would satisfy the "PSK Reporter" use case - ie get enough info top pop a report up on a web page.  Quite different to the voice stream, where you need every packet to get through.

However there is also another requirement for the humans watching - they would like to see the station ID pop up in the approx 10 second timeframe.  I hit this a few years ago when I tried a similar project.  End users like that "real time" cue.

tmiw:
@drowe67, I actually wonder if the use of varicode for the framing is interfering with use of LDPC and/or Golay. Currently the code simply stores [frame width] plus a bit of the bytes that come in and iterates across each possible location (selecting the first block that produces a BER that is below the threshold). It's possible that a flipped bit or two at the varicode level results in a more significant bit change at this level, meaning that LDPC/Golay have to work harder to correct the incoming data (if it's even possible).

Maybe something I can try is revisiting the custom varicode encoder/decoder idea. Perhaps it can be as simple as "pass the received bit to the higher level code" and let reliable_text look for the UW (which would definitely be required if we went this route). BTW, I'm thinking 16 bit UW of all ones with a threshold of 4 zeroes based on the blog post you linked as bit 8 isn't used in the current implementation.

tmiw:
It looks like just piping the bits directly to reliable_text improved things significantly. In particular, it seems we've attained 2x the transfer speed at high enough SNR (1 callsign every ~5-6 seconds instead of 11-12 seconds like before) and we can get 1 decode/minute down to a bit lower SNR than before (3/5dB vs. 1/3dB with 700D/E).

_700D_:

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.75 |             17.75 |                                439 |                  0.00% |   11.71187194 |
|                 -29 |          14.8 |              16.8 |                                439 |                  0.00% |   11.71187194 |
|                 -28 |         13.78 |             15.78 |                                439 |                  0.00% |   11.71187194 |
|                 -27 |         12.75 |             14.75 |                                439 |                  0.00% |   11.71187194 |
|                 -26 |          11.8 |              13.8 |                                439 |                  0.00% |   11.71187194 |
|                 -25 |         10.78 |             12.78 |                                439 |                  0.00% |   11.71187194 |
|                 -24 |          9.75 |             11.75 |                                438 |                  0.23% |   11.68519342 |
|                 -23 |           8.8 |              10.8 |                                437 |                  0.46% |    11.6585149 |
|                 -22 |          7.78 |              9.78 |                                430 |                  2.05% |   11.47176523 |
|                 -21 |          6.75 |              8.75 |                                414 |                  5.69% |   11.04490885 |
|                 -20 |           5.8 |               7.8 |                                381 |                 13.21% |   10.16451756 |
|                 -19 |          4.78 |              6.78 |                                323 |                 26.42% |   8.617163184 |
|                 -18 |          3.75 |              5.75 |                                254 |                 42.14% |   6.776345042 |
|                 -17 |           2.8 |               4.8 |                                174 |                 60.36% |   4.642063139 |
|                 -16 |          1.78 |              3.78 |                                 97 |                 77.90% |   2.587816807 |
|                 -15 |          0.75 |              2.75 |                                 45 |                 89.75% |    1.20053357 |
|                 -14 |          -0.2 |               1.8 |                                 21 |                 95.22% |  0.5602489996 |
|                 -13 |         -1.21 |              0.79 |                                  0 |                100.00% |             0 |

_700E_:

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.98 |             14.98 |                                439 |                  0.00% |   11.71187194 |
|                 -29 |         15.02 |             14.02 |                                439 |                  0.00% |   11.71187194 |
|                 -28 |         14.01 |             13.01 |                                439 |                  0.00% |   11.71187194 |
|                 -27 |         12.98 |             11.98 |                                439 |                  0.00% |   11.71187194 |
|                 -26 |         12.02 |             11.02 |                                437 |                  0.46% |    11.6585149 |
|                 -25 |         11.01 |             10.01 |                                427 |                  2.73% |   11.39172966 |
|                 -24 |          9.98 |              8.98 |                                413 |                  5.92% |   11.01823032 |
|                 -23 |          9.02 |              8.02 |                                388 |                 11.62% |   10.35126723 |
|                 -22 |          8.01 |              7.01 |                                356 |                 18.91% |   9.497554469 |
|                 -21 |          6.98 |              5.98 |                                284 |                 35.31% |   7.576700756 |
|                 -20 |          6.02 |              5.02 |                                191 |                 56.49% |   5.095598044 |
|                 -19 |          5.01 |              4.01 |                                105 |                 76.08% |   2.801244998 |
|                 -18 |          3.98 |              2.98 |                                 48 |                 89.07% |   1.280569142 |
|                 -17 |          3.02 |              2.02 |                                 16 |                 96.36% |  0.4268563806 |
|                 -16 |          2.01 |              1.01 |                                  2 |                 99.54% | 0.05335704758 |
|                 -15 |          0.98 |             -0.02 |                                  0 |                100.00% |             0 |

One thing we could maybe do is change the callback so that it returns whether we got a valid decode. freedv-gui and other clients could then use that to determine whether to report to e.g. PSK Reporter while unconditionally displaying what was received (even if incorrect). This may satisfy at least some of the demand for quick feedback, though I'm not sure if "always display something even if wrong" is better than "only display if correct".

Anyway, maybe one of the next steps could be to figure out how to get the raw OFDM, etc. symbols over to reliable_text as well to actually use soft decision LDPC. Also, interleaving still needs to be implemented here, which may help a bit even in the absence of fading.

drowe67:
> Perhaps it can be as simple as "pass the received bit to the higher level code" and let reliable_text look for the UW (which would definitely be required if we went this route). 

Yes I think this is a good approach.  You might be able to use a shorter UW (say 8-12 bits). As per Req 1 above it's OK if you get a few UW misses or false detects.  Anyway, you can experiment with that.  The current Varicode design is OK, as it will be FEC (and I presume) CRC protected.

The other complication is that your Varicode payload will vary in length, so you'll need to set an upper limit, or provide a numberOfBytes (numberOfBits/numberOfGolay codewords) field early in your payload packet data.

A possible packet design is:
```
|UW|Data|Parity|
```
Data is then FEC decoded and further broken down to:
```
|CRC|Varicode bit stream ........|
```
You toss out packets that don't pass the CRC check.  You could optionally add a length field, and use that to work out the length of the packet.  Or delineate packets by the location of the next UW.

drowe67:
My first guess at the theoretical performance (anyone who knows probability please feel free to correct!):

The probability of a single good Golay decode (0,1,2, or 3 errors) at a 6% raw BER (0dB SNR using 700D) is:
```
octave:46> sum(binopdf(0:3, 23, 0.06))
ans =  0.95405
```
Say the PSK reporter packet has 6 Golay codewords (5*12=60 payload data bits), the probability of 5 consecutive good codewords is:
```
octave:47> sum(binopdf(0:3, 23, 0.06))^5
ans =  0.79044
```
Or the probability of a bad packet (at least one dud Golay codeword) is:
```
octave:48> 1-sum(binopdf(0:3, 23, 0.06))^5
ans =  0.20956
```
At 10% raw BER (where 700D falls over for voice) it's 65% PER - still usable for PSK reporting.  This is the probability given we have frame sync.

tmiw:
@drowe67, I added soft decision support tonight (for 700D/E at least, 2020 should be possible but isn't hooked in yet) and it looks like we've met the initial goal I had. In fact, even at 0dB above the mode's minimum, we're still decoding quickly enough that the user should see a valid callsign within ~10-12 seconds.

_700D_:

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.75 |             17.75 |                                439 |                  0.00% |   11.71187194 |
|                 -29 |          14.8 |              16.8 |                                439 |                  0.00% |   11.71187194 |
|                 -28 |         13.78 |             15.78 |                                439 |                  0.00% |   11.71187194 |
|                 -27 |         12.75 |             14.75 |                                439 |                  0.00% |   11.71187194 |
|                 -26 |          11.8 |              13.8 |                                439 |                  0.00% |   11.71187194 |
|                 -25 |         10.78 |             12.78 |                                439 |                  0.00% |   11.71187194 |
|                 -24 |          9.75 |             11.75 |                                439 |                  0.00% |   11.71187194 |
|                 -23 |           8.8 |              10.8 |                                439 |                  0.00% |   11.71187194 |
|                 -22 |          7.78 |              9.78 |                                439 |                  0.00% |   11.71187194 |
|                 -21 |          6.75 |              8.75 |                                439 |                  0.00% |   11.71187194 |
|                 -20 |           5.8 |               7.8 |                                439 |                  0.00% |   11.71187194 |
|                 -19 |          4.78 |              6.78 |                                439 |                  0.00% |   11.71187194 |
|                 -18 |          3.75 |              5.75 |                                439 |                  0.00% |   11.71187194 |
|                 -17 |           2.8 |               4.8 |                                439 |                  0.00% |   11.71187194 |
|                 -16 |          1.78 |              3.78 |                                439 |                  0.00% |   11.71187194 |
|                 -15 |          0.75 |              2.75 |                                439 |                  0.00% |   11.71187194 |
|                 -14 |          -0.2 |               1.8 |                                438 |                  0.23% |   11.68519342 |
|                 -13 |         -1.21 |              0.79 |                                398 |                  9.34% |   10.61805247 |
|                 -12 |         -2.24 |             -0.24 |                                252 |                 42.60% |   6.722987995 |
|                 -11 |          -3.2 |              -1.2 |                                 93 |                 78.82% |   2.481102712 |
|                 -10 |         -4.21 |             -2.21 |                                 27 |                 93.85% |  0.7203201423 |
|                  -9 |         -5.24 |             -3.24 |                                  6 |                 98.63% |  0.1600711427 |
|                  -8 |          -6.2 |              -4.2 |                                  0 |                100.00% |             0 |

_700E_:

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.98 |             14.98 |                                439 |                  0.00% |   11.71187194 |
|                 -29 |         15.02 |             14.02 |                                439 |                  0.00% |   11.71187194 |
|                 -28 |         14.01 |             13.01 |                                439 |                  0.00% |   11.71187194 |
|                 -27 |         12.98 |             11.98 |                                439 |                  0.00% |   11.71187194 |
|                 -26 |         12.02 |             11.02 |                                439 |                  0.00% |   11.71187194 |
|                 -25 |         11.01 |             10.01 |                                439 |                  0.00% |   11.71187194 |
|                 -24 |          9.98 |              8.98 |                                439 |                  0.00% |   11.71187194 |
|                 -23 |          9.02 |              8.02 |                                439 |                  0.00% |   11.71187194 |
|                 -22 |          8.01 |              7.01 |                                439 |                  0.00% |   11.71187194 |
|                 -21 |          6.98 |              5.98 |                                439 |                  0.00% |   11.71187194 |
|                 -20 |          6.02 |              5.02 |                                439 |                  0.00% |   11.71187194 |
|                 -19 |          5.01 |              4.01 |                                439 |                  0.00% |   11.71187194 |
|                 -18 |          3.98 |              2.98 |                                439 |                  0.00% |   11.71187194 |
|                 -17 |          3.02 |              2.02 |                                437 |                  0.46% |    11.6585149 |
|                 -16 |          2.01 |              1.01 |                                410 |                  6.61% |   10.93819475 |
|                 -15 |          0.98 |             -0.02 |                                297 |                 32.35% |   7.923521565 |
|                 -14 |          0.02 |             -0.98 |                                106 |                 75.85% |   2.827923522 |
|                 -13 |         -0.99 |             -1.99 |                                 28 |                 93.62% |  0.7469986661 |
|                 -12 |         -2.02 |             -3.02 |                                  8 |                 98.18% |  0.2134281903 |
|                 -11 |         -2.98 |             -3.98 |                                  0 |                100.00% |             0 |

I also adjusted the new unit tests given the above results and they pass on my machine (and should hopefully pass in the build system too). Tasks remaining:

* Add interleaving (need to investigate how to do this when we're also doing soft decision).
* Add some better error handling (e.g. detect attempts to set strings longer than our current maximum and truncate accordingly).
* Link the soft decision stuff into 2020 as well and add unit testing for that mode + reliable_text.

src/reliable_text.c
```diff
@@ -0,0 +1,444 @@
...
+        if (byte & (1 << bitToCheck))
+        {
+            ibits[ibitsBitIndex] = 1;
+        }
+    }
+    
+    encode(&impl->ldpc, ibits, pbits);  
+    
+    // Split LDPC encoded bits into individual bits, with the first RELIABLE_TEXT_UW_LENGTH_BITS being UW.
+    memset(impl->tx_text, 1, RELIABLE_TEXT_UW_LENGTH_BITS);
```
drowe67, line 419: OK so an all 1's UW?  We usually avoid that as but I think it's OK in this case.

src/reliable_text.c
```diff
@@ -0,0 +1,444 @@
...
+        }
+        xor_uw <<= 1;
+    }
+    obj->uw_bits_valid = obj->bit_index >= RELIABLE_TEXT_UW_LENGTH_BITS && num_zeroes <= RELIABLE_TEXT_MAX_ZEROES_IN_UW;
+}
+
+static void reliable_text_freedv_callback_rx(void *state, char chr)
+{
+    //fprintf(stderr, "received char: %d\n", (chr & 0x3F));
+
```
drowe67, line 247: Perhaps consider a state/next_state switch/case state machine implementation in this function for the acquisition (looking for UW) and synced states.  Tends to work out neater.  Many examples in other parts of codec2.

src/reliable_text.c
drowe67: The reason we avoid it is occasionally we get edge cases (e.g sync on noise) that lead to the same QPSK symbols being decoded.  The reason I don't think it's a concern here is that the OFDM modem already has solid sync so the data your algorithms get fed won't be too crazy.

drowe67:
Great work @tmiw :champagne:  You are certainly picking up all the technologies around codec2 and FreeDV quickly!  You've developed your own custom packet data protocol :slightly_smiling_face:

Also looks like your measured PER results (a 50% PER knee at 0dB SNR for 700D) are pretty close to the theoretical model sketched out above :+1: (albeit for a different code).  Note the sharp knee in PER versus SNR - that's pretty typical for powerful FEC codes like LDPC.  The good news is the PER Reporter use case still works with very high PERs (unlike say generic packet data or IP links)

It might be useful to generate some plots using your tables above (which I imagine are machine generated markup).  PER against SNR is a good one, with a line drawn for where voice falls over.

To give the interleaver a work out try adding a ctest on Multipath Poor `cohpsk --mpp`, that's the standard fading channel model.  I wouldn't be surprised if it works better that voice, as you are interleaving over a really long time.  Typically, multipath channels require about 4dB more SNR than AWGN for the same BER/PER.

Be exciting to see FreeDV popping up PSK reporter soon....

src/reliable_text.c
tmiw: I ended up redoing how UW gets sync'd anyway due to issues I found while trying to implement interleaving. Will push it up shortly.

tmiw:
@drowe67, hmm, I just linked 2020 in and it seems that it's behaving significantly worse than 700D/E even though I'm not applying any AWGN or fading:

```
HT-TM05:src mooneer$ ./freedv_tx 2020 ../../raw/ve9qrp_multi.raw 2020_reliable.raw --reliabletext AB1CDEF 2>/dev/null
HT-TM05:src mooneer$ ./freedv_rx 2020 2020_reliable.raw /dev/null --txtrx 2020_reliable.txt --reliabletext
frames decoded: 6247  output speech samples: 17991360
HT-TM05:src mooneer$ grep "AB1CDEF" 2020_reliable.txt | wc -l
     195
HT-TM05:src mooneer$ 
```

Maybe this provides some clues for #217?

tmiw:
Also, I generated some new data for fading and non-fading after adding interleaving:

_700D, no fading_:

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.75 |             17.75 |                                439 |                  0.00% |   11.71187194 |
|                 -29 |          14.8 |              16.8 |                                439 |                  0.00% |   11.71187194 |
|                 -28 |         13.78 |             15.78 |                                439 |                  0.00% |   11.71187194 |
|                 -27 |         12.75 |             14.75 |                                439 |                  0.00% |   11.71187194 |
|                 -26 |          11.8 |              13.8 |                                439 |                  0.00% |   11.71187194 |
|                 -25 |         10.78 |             12.78 |                                439 |                  0.00% |   11.71187194 |
|                 -24 |          9.75 |             11.75 |                                439 |                  0.00% |   11.71187194 |
|                 -23 |           8.8 |              10.8 |                                439 |                  0.00% |   11.71187194 |
|                 -22 |          7.78 |              9.78 |                                439 |                  0.00% |   11.71187194 |
|                 -21 |          6.75 |              8.75 |                                439 |                  0.00% |   11.71187194 |
|                 -20 |           5.8 |               7.8 |                                439 |                  0.00% |   11.71187194 |
|                 -19 |          4.78 |              6.78 |                                439 |                  0.00% |   11.71187194 |
|                 -18 |          3.75 |              5.75 |                                439 |                  0.00% |   11.71187194 |
|                 -17 |           2.8 |               4.8 |                                439 |                  0.00% |   11.71187194 |
|                 -16 |          1.78 |              3.78 |                                439 |                  0.00% |   11.71187194 |
|                 -15 |          0.75 |              2.75 |                                437 |                  0.46% |    11.6585149 |
|                 -14 |          -0.2 |               1.8 |                                432 |                  1.59% |   11.52512228 |
|                 -13 |         -1.21 |              0.79 |                                403 |                  8.20% |   10.75144509 |
|                 -12 |         -2.24 |             -0.24 |                                312 |                 28.93% |   8.323699422 |
|                 -11 |          -3.2 |              -1.2 |                                163 |                 62.87% |   4.348599378 |
|                 -10 |         -4.21 |             -2.21 |                                 51 |                 88.38% |   1.360604713 |
|                  -9 |         -5.24 |             -3.24 |                                 12 |                 97.27% |  0.3201422855 |
|                  -8 |          -6.2 |              -4.2 |                                  2 |                 99.54% | 0.05335704758 |

_700D, fading:_

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.75 |             17.75 |                                435 |                  0.91% |   11.60515785 |
|                 -29 |          14.8 |              16.8 |                                435 |                  0.91% |   11.60515785 |
|                 -28 |         13.78 |             15.78 |                                435 |                  0.91% |   11.60515785 |
|                 -27 |         12.75 |             14.75 |                                433 |                  1.37% |    11.5518008 |
|                 -26 |          11.8 |              13.8 |                                434 |                  1.14% |   11.57847932 |
|                 -25 |         10.78 |             12.78 |                                433 |                  1.37% |    11.5518008 |
|                 -24 |          9.75 |             11.75 |                                432 |                  1.59% |   11.52512228 |
|                 -23 |           8.8 |              10.8 |                                430 |                  2.05% |   11.47176523 |
|                 -22 |          7.78 |              9.78 |                                414 |                  5.69% |   11.04490885 |
|                 -21 |          6.75 |              8.75 |                                409 |                  6.83% |   10.91151623 |
|                 -20 |           5.8 |               7.8 |                                388 |                 11.62% |   10.35126723 |
|                 -19 |          4.78 |              6.78 |                                376 |                 14.35% |   10.03112494 |
|                 -18 |          3.75 |              5.75 |                                345 |                 21.41% |   9.204090707 |
|                 -17 |           2.8 |               4.8 |                                293 |                 33.26% |    7.81680747 |
|                 -16 |          1.78 |              3.78 |                                241 |                 45.10% |   6.429524233 |
|                 -15 |          0.75 |              2.75 |                                196 |                 55.35% |   5.228990663 |
|                 -14 |          -0.2 |               1.8 |                                139 |                 68.34% |   3.708314807 |
|                 -13 |         -1.21 |              0.79 |                                 80 |                 81.78% |   2.134281903 |
|                 -12 |         -2.24 |             -0.24 |                                 37 |                 91.57% |  0.9871053802 |
|                 -11 |          -3.2 |              -1.2 |                                 13 |                 97.04% |  0.3468208092 |
|                 -10 |         -4.21 |             -2.21 |                                  7 |                 98.41% |  0.1867496665 |
|                  -9 |         -5.24 |             -3.24 |                                  3 |                 99.32% | 0.08003557137 |
|                  -8 |          -6.2 |              -4.2 |                                  1 |                 99.77% | 0.02667852379 |

_700E, no fading_:

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.98 |             14.98 |                                439 |                  0.00% |   11.71187194 |
|                 -29 |         15.02 |             14.02 |                                439 |                  0.00% |   11.71187194 |
|                 -28 |         14.01 |             13.01 |                                439 |                  0.00% |   11.71187194 |
|                 -27 |         12.98 |             11.98 |                                439 |                  0.00% |   11.71187194 |
|                 -26 |         12.02 |             11.02 |                                439 |                  0.00% |   11.71187194 |
|                 -25 |         11.01 |             10.01 |                                439 |                  0.00% |   11.71187194 |
|                 -24 |          9.98 |              8.98 |                                439 |                  0.00% |   11.71187194 |
|                 -23 |          9.02 |              8.02 |                                439 |                  0.00% |   11.71187194 |
|                 -22 |          8.01 |              7.01 |                                439 |                  0.00% |   11.71187194 |
|                 -21 |          6.98 |              5.98 |                                439 |                  0.00% |   11.71187194 |
|                 -20 |          6.02 |              5.02 |                                439 |                  0.00% |   11.71187194 |
|                 -19 |          5.01 |              4.01 |                                439 |                  0.00% |   11.71187194 |
|                 -18 |          3.98 |              2.98 |                                439 |                  0.00% |   11.71187194 |
|                 -17 |          3.02 |              2.02 |                                437 |                  0.46% |    11.6585149 |
|                 -16 |          2.01 |              1.01 |                                426 |                  2.96% |   11.36505113 |
|                 -15 |          0.98 |             -0.02 |                                361 |                 17.77% |   9.630947088 |
|                 -14 |          0.02 |             -0.98 |                                174 |                 60.36% |   4.642063139 |
|                 -13 |         -0.99 |             -1.99 |                                 56 |                 87.24% |   1.493997332 |
|                 -12 |         -2.02 |             -3.02 |                                 14 |                 96.81% |   0.373499333 |
|                 -11 |         -2.98 |             -3.98 |                                  1 |                 99.77% | 0.02667852379 |

_700E, fading:_

| cohpsk_ch parameter | cohpsk_ch SNR | dB Above Mode Min | Number of valid callsigns received | Error rate vs. "ideal" | Callsigns/min |
|---------------------|---------------|-------------------|------------------------------------|------------------------|---------------|
|                 -30 |         15.98 |             14.98 |                                439 |                  0.00% |   11.71187194 |
|                 -29 |         15.02 |             14.02 |                                438 |                  0.23% |   11.68519342 |
|                 -28 |         14.01 |             13.01 |                                435 |                  0.91% |   11.60515785 |
|                 -27 |         12.98 |             11.98 |                                436 |                  0.68% |   11.63183637 |
|                 -26 |         12.02 |             11.02 |                                428 |                  2.51% |   11.41840818 |
|                 -25 |         11.01 |             10.01 |                                428 |                  2.51% |   11.41840818 |
|                 -24 |          9.98 |              8.98 |                                424 |                  3.42% |   11.31169409 |
|                 -23 |          9.02 |              8.02 |                                420 |                  4.33% |   11.20497999 |
|                 -22 |          8.01 |              7.01 |                                409 |                  6.83% |   10.91151623 |
|                 -21 |          6.98 |              5.98 |                                389 |                 11.39% |   10.37794575 |
|                 -20 |          6.02 |              5.02 |                                364 |                 17.08% |   9.710982659 |
|                 -19 |          5.01 |              4.01 |                                323 |                 26.42% |   8.617163184 |
|                 -18 |          3.98 |              2.98 |                                253 |                 42.37% |   6.749666518 |
|                 -17 |          3.02 |              2.02 |                                199 |                 54.67% |   5.309026234 |
|                 -16 |          2.01 |              1.01 |                                133 |                 69.70% |   3.548243664 |
|                 -15 |          0.98 |             -0.02 |                                 82 |                 81.32% |   2.187638951 |
|                 -14 |          0.02 |             -0.98 |                                 37 |                 91.57% |  0.9871053802 |
|                 -13 |         -0.99 |             -1.99 |                                 27 |                 93.85% |  0.7203201423 |
|                 -12 |         -2.02 |             -3.02 |                                  7 |                 98.41% |  0.1867496665 |
|                 -11 |         -2.98 |             -3.98 |                                  2 |                 99.54% | 0.05335704758 |

I didn't run the fading tests prior to adding fading but it does look like there was some improvement at lower SNR in the non-fading case. 👍

drowe67:
> I didn't run the fading tests prior to adding fading but it does look like there was some improvement at lower SNR in the non-fading case. +1

So is "non fading" is AWGN and "fading" a MPP multipath channel?

It could just be statistical variation - for an AWGN channel we would expect the interleaver to have no effect.  Fading results look good :+1:   I note the knee in the PER versus SNR curve is more gentle than AWGN.  This is useful for the PSK Reporter use case - we still get a few packets through at very low SNRs.

src/reliable_text.c
```diff
@@ -0,0 +1,433 @@
...
+        
+        // Deinterleave symbols
+        gp_deinterleave_comp ((COMP*)deinterleavedSyms, (COMP*)&obj->inbound_pending_syms[RELIABLE_TEXT_UW_LENGTH_BITS/2], Npayloadsymsperpacket);
+        gp_deinterleave_float(deinterleavedAmps, &obj->inbound_pending_amps[RELIABLE_TEXT_UW_LENGTH_BITS/2], Npayloadsymsperpacket);
+        
+        float EsNo = 3.0; // note: constant from freedv_700.c
+        
+        symbols_to_llrs(llr, (COMP*)deinterleavedSyms, deinterleavedAmps,
+                        EsNo, obj->fdv->ofdm->mean_amp, Npayloadsymsperpacket);
+    }
```
drowe67, line 184: OK so `mean_amp` is estimated [here](https://github.com/drowe67/codec2/blob/master/src/ofdm.c#L1735), it's a time averaged estimate of the pilot symbol amplitude.  I think this is OK, as the mean amplitude for a HF channel won't vary too quickly (e.g. over the PSK Packet length).

drowe67:
@tmiw re 2020 with no noise, lets compare a reference simulation;
```
./freedv_tx 2020 ../../raw/ve9qrp.raw - --testframes | ./freedv_rx 2020 - /dev/null --testframes
BER......: 0.0000  Tbits:   137640  Terrs:        0
Coded BER: 0.0000  Tbits:    96720  Terrs:        0
Coded PER:  -nan  Tpkts:        0  Tpers:        0
```
So we would expect to have no bit errors with no channel noise. (it also appears the Coded PER metric is broken for 2020).

drowe67:
ve9qrp is a very big file (34 Mbytes).  Perhaps we could make this a makefile target that gets built from ve9qrp?  Hmm, maybe I could do that for the fading files too.

tmiw:
> So is "non fading" is AWGN and "fading" a MPP multipath channel?

Correct.

> ve9qrp is a very big file (34 Mbytes). Perhaps we could make this a makefile target that gets built from ve9qrp? Hmm, maybe I could do that for the fading data files too. Need to work out how to make cmake build targets that aren't C files.

Yeah, I noticed that after I pushed that commit up. I was probably going to update the unit tests to just use ve9qrp and not ve9qrp_multi, but that depends on whether the former is long enough to get enough packets on the AWGN tests.

> @tmiw - did you generate fading files that were long enough for your tests? A warning pops out of `cohpsk_ch` is we run out of fading file samples. After that it's all AWGN simulation ....

Yep, I used ofdm_fade.sh as an example of what to provide to octave to generate the files.

> 
> A better solution would be generating fading samples on the fly, but that's another project....

tmiw:
So yeah, I think this is good for review. Current status of unit tests:

```
HT-TM05:build_osx mooneer$ ctest -R test_freedv_reliable_text
      Start 66: test_freedv_reliable_text_truncate_string
 1/13 Test #66: test_freedv_reliable_text_truncate_string ...   Passed    1.24 sec
      Start 67: test_freedv_reliable_text_ideal_1600
 2/13 Test #67: test_freedv_reliable_text_ideal_1600 ........   Passed    0.80 sec
      Start 68: test_freedv_reliable_text_ideal_700D
 3/13 Test #68: test_freedv_reliable_text_ideal_700D ........   Passed    0.90 sec
      Start 69: test_freedv_reliable_text_ideal_700E
 4/13 Test #69: test_freedv_reliable_text_ideal_700E ........   Passed    0.96 sec
      Start 70: test_freedv_reliable_text_awgn_1600
 5/13 Test #70: test_freedv_reliable_text_awgn_1600 .........   Passed    1.02 sec
      Start 71: test_freedv_reliable_text_awgn_700D
 6/13 Test #71: test_freedv_reliable_text_awgn_700D .........   Passed    1.55 sec
      Start 72: test_freedv_reliable_text_awgn_700E
 7/13 Test #72: test_freedv_reliable_text_awgn_700E .........   Passed    1.48 sec
      Start 73: test_freedv_reliable_text_fade_1600
 8/13 Test #73: test_freedv_reliable_text_fade_1600 .........   Passed    1.05 sec
      Start 74: test_freedv_reliable_text_fade_700D
 9/13 Test #74: test_freedv_reliable_text_fade_700D .........   Passed    1.16 sec
      Start 75: test_freedv_reliable_text_fade_700E
10/13 Test #75: test_freedv_reliable_text_fade_700E .........   Passed    1.14 sec
      Start 76: test_freedv_reliable_text_ideal_2020
11/13 Test #76: test_freedv_reliable_text_ideal_2020 ........   Passed    9.86 sec
      Start 77: test_freedv_reliable_text_awgn_2020
12/13 Test #77: test_freedv_reliable_text_awgn_2020 .........   Passed   10.98 sec
      Start 78: test_freedv_reliable_text_fade_2020
13/13 Test #78: test_freedv_reliable_text_fade_2020 .........   Passed   10.21 sec

100% tests passed, 0 tests failed out of 13

Total Test time (real) =  42.36 sec
HT-TM05:build_osx mooneer$ 
```

tmiw:
Interesting, test_freedv_reliable_text_truncate_string passes on my machine but not when run as a GitHub action. Since it's based off test_freedv_reliable_text_ideal_1600 I would have expected both to fail if there was actually an issue. Any ideas?

drowe67:
> Interesting, test_freedv_reliable_text_truncate_string passes on my machine but not when run as a GitHub action. Since it's based off test_freedv_reliable_text_ideal_1600 I would have expected both to fail if there was actually an issue. Any ideas?

I've experienced this in the past with:
1. Small floating point differences between machines (or different C library builds) leading to different test results. In this case the test design/requirements sometimes need tweaking.
1. uninitialised memory issues that pop up on different machines.

It fails on my Ubuntu 20 machine too:
```
ctest -V -R test_freedv_reliable_text_truncate_string
<snip>
6: Test command: /bin/sh "-c" "cd /home/david/codec2/build_linux/src;
                          ./freedv_tx 1600 ../../raw/ve9qrp.raw - --reliabletext AB1CDEFGH > 1600_reliable.raw 2>/dev/null;
                          ./freedv_rx 1600 1600_reliable.raw /dev/null --txtrx 1600_reliable.txt --reliabletext 2>/dev/null;
                          grep 'AB1CDEFG' 1600_reliable.txt | wc -l"
66: Test timeout computed to be: 1500
66: 0
1/1 Test #66: test_freedv_reliable_text_truncate_string ...***Failed  Required regular expression not found. Regex=[20
]  1.18 sec

0% tests passed, 1 tests failed out of 1
```

If you get stuck on this one/can't reproduce let me know and I can dig in some time.

tmiw:
> It fails on my Ubuntu 20 machine too:

Yeah, looks like it was a stack overflow issue; I was able to duplicate it in my Ubuntu VM too. Should be fixed now. 👍

PR #227, opened 2021-09-07
reliable_text code cleanup

tmiw:
I was looking into an issue decoding my callsign (e.g. K6AQF or K6AQR instead of K6AQ) earlier. I'm not sure these changes help but they're probably good practice regardless.


drowe67:
@tmiw - looks good, pls press merge when you're happy with this and it's freedv-gui companion PR :slightly_smiling_face:

PR #230, opened 2021-09-21
Save bytes in newamp1.c

srsampson:
Change ideal[] to array of char and cast to float when used.

Saves 60 bytes, although it would be in constant flash anyway.

### DIFF
--- a/src/newamp1.c
+++ b/src/newamp1.c
@@ -438,14 +438,14 @@ void determine_autoc(C2CONST *c2const, float Rk[], int order, MODEL *model, int 
 
 /* update and optionally run "front eq" equaliser on before VQ */
 void newamp1_eq(float rate_K_vec_no_mean[], float eq[], int K, int eq_en) {
-    static float ideal[] = {8,10,12,14,14,14,14,14,14,14,14,14,14,14,14,14,14,14,14,-20};
-    float gain = 0.02;
+    static const char ideal[] = {8,10,12,14,14,14,14,14,14,14,14,14,14,14,14,14,14,14,14,-20};
+    float gain = 0.02f;
     float update;
     
     for(int k=0; k<K; k++) {
-        update = rate_K_vec_no_mean[k] - ideal[k];
-        eq[k] = (1.0-gain)*eq[k] + gain*update;
-        if (eq[k] < 0.0) eq[k] = 0.0;
+        update = rate_K_vec_no_mean[k] - (float) ideal[k];
+        eq[k] = (1.0f-gain)*eq[k] + gain*update;
+        if (eq[k] < 0.0f) eq[k] = 0.0f;
         if (eq_en)
             rate_K_vec_no_mean[k] -= eq[k];
     }


srsampson:
No failures, ready to merge.

drowe67:
Hi @srsampson.  Thanks, couple of questions:
1. Have you confirmed the equaliser performs exactly as before?  `test_700c_eq.sh` checks if the EQ lowers the VQ variance (ie quantiser error) which is a start - but in this case it would be nice to ensure the actual VQ values are very close to those obtained from previous float version.  It's an important part of the on air experience, and in code that affects voice quality bugs can be very hard to detect. 
1. Do you have the ability to repeat the stm32 tests (i.e. a Discovery board)? If not I can set up my machine here and run them (or give you ssh access).

srsampson:
Well, I didn't think it was that complex, i.e.:
```
#include <stdio.h>

int main(char **argv, int argc) {
	char c = 14;
	float f = 14;
	float math1 = 1E-6f + (float) c;
	float math2 = 1E-6f + f;

        printf("c = %d c cast = %f  f = %f\n", c, (float)c, f);
	printf("math1 = %f  math2 = %f\n", math1, math2);
}

```
Produces:
```
c = 14 c cast = 14.000000  f = 14.000000
math1 = 14.000001  math2 = 14.000001
```

But I need to compile a new SM1000 version anyway.

drowe67:
Re (1) I dumped the eq coeffs from master acaed6bb9 and this branch: https://github.com/drowe67/codec2/pull/230/commits/92163feb0f0c8838b0e3d7295fff9edfadaf74ad

master:
```
101: 0 10.901608
101: 1 5.595476
101: 2 1.031233
101: 3 0.000000
101: 4 0.000000
101: 5 0.000000
101: 6 0.000000
101: 7 0.000000
101: 8 0.000000
101: 9 0.000000
101: 10 0.000000
101: 11 0.000000
101: 12 0.000000
101: 13 0.000000
101: 14 0.000000
101: 15 0.000000
101: 16 0.000000
101: 17 0.000000
101: 18 0.000000
101: 19 6.812810
101: ../raw/kristoff.raw var: 12.63 std:  3.55
101: var: 12.63 var_eq: 10.47
```

ss-savebytes:
```
92: 0 10.901614
92: 1 5.595479
92: 2 1.031235
92: 3 0.000000
92: 4 0.000000
92: 5 0.000000
92: 6 0.000000
92: 7 0.000000
92: 8 0.000000
92: 9 0.000000
92: 10 0.000000
92: 11 0.000000
92: 12 0.000000
92: 13 0.000000
92: 14 0.000000
92: 15 0.000000
92: 16 0.000000
92: 17 0.000000
92: 18 0.000000
92: 19 6.812815
92: ../raw/kristoff.raw var: 12.63 std:  3.55
92: var: 12.63 var_eq: 10.47
```

The var_eq from both is the same which is a good sign, and the eq values at the end of the test run look very close so I think we're good :+1:  Not exactly the same which is curious.

Nothing to do with this PR, but also curious that many of them are zero.  This could mean that this sample only needs EQ on the band edges.  Might be worth digging into some time.

drowe67:
Hi @srsampson - these tests seem to be failing on the stm32:
```
ctest --timeout 300
<snip>
19 - tst_api_mod_700D_CODEC (Failed)
21 - tst_api_mod_700E_CODEC (Failed)
23 - tst_api_demod_700D_AWGN_test (Timeout)
```
The first two pass on master, however `23 - tst_api_demod_700D_AWGN_test` is also timing out on master. I'm not sure why :thinking: might be something else that's been broken recently, or maybe I messed up in my configuration.  I ran the tests 2 times on each branch with the same results.  

The `--timeout` option prevents 23 hanging for 1500 seconds (the ctest default).  Looking at the [recent 700E PR](https://github.com/drowe67/codec2/pull/206) this test should take 60 seconds.

@tmiw any thoughts?  I've left bear on.  Steve, just email me a public ssh key if you'd like an account on this machine.

srsampson:
You are fast, I'm hooking up my Discovery board and will give it some runs in the morning. The float differences are really surprising. I'll also look at the test failures.

srsampson:
I changed the 'char' to 'int8_t' and the 700D and 700E tests now pass. Will look at the other one.

The char was returning 236 (or so) instead of -20. So much for my simple tests, argh.

srsampson:
tst_api_demod_700D_AWGN_test timeout for me also.

srsampson:
Fail list:
	 10 - tst_ofdm_demod_fade (Failed)
	 13 - tst_ofdm_demod_ldpc_fade (Failed)
	 22 - tst_api_demod_700D_plain_test (Timeout)
	 23 - tst_api_demod_700D_AWGN_test (Timeout)

tmiw:
Looks like tst_api_demod_700D_AWGN_test might have crashed processing frame 5?

```
23: frame: 3  demod sync: 1  nin: 1280 demod snr: 0.05 dB  bit errors: 12
23: frame: 4, 2560 bytes read
23:   2 nin: 1280 st: trial  euw:  0  0 mf:  0 f:  -4.3 pbw: 1 snr:  0.0 eraw:   11 ecdd:    3 iter:  10 pcc:  92 rxst: EBST
23:   wvecpilot  0.01 msecs
23:   corr  1.04 msecs
23:   freedv_rx 27.20 msecs
23:   0 short speech values returned
23: frame: 4  demod sync: 1  nin: 1280 demod snr: 0.01 dB  bit errors: 23
23: frame: 5, 2560 bytes read
^C
```

srsampson:
First two are:
10: ERROR in tst_ofdm_demod_setup! Exiting...
13: ERROR in tst_ofdm_demod_setup! Exiting...

srsampson:
> Looks like tst_api_demod_700D_AWGN_test might have crashed processing frame 5?
> 
Same error spot as mine.

srsampson:
Got tests 10 and 13 to pass by using octave to create the flot fading files in /raw

tmiw:
Is it possible it's the reliable_text stuff? I'm not sure how, though, since that should only be getting enabled if you call reliable_text_use_with_freedv().

drowe67:
> I changed the 'char' to 'int8_t' and the 700D and 700E tests now pass. Will look at the other one.

> The char was returning 236 (or so) instead of -20. So much for my simple tests, argh.

Yep - this is why we need to test very carefully, especially anything that processes voice signals where bugs can be subtle and hard to track down.  The modem work has better objective measures that make bugs more obvious.

I am concerned that the x86 based ctests didn't pick up this issue.  It would be good to work out why and make the EQ ctest more robust :thinking:

donmr:
I think you are seeing compiler differenced.

"The C and C++ standards allows the character type char to be signed or 
unsigned, depending on the platform and compiler"

Don

drowe67:
`ctest -V -R tst_api_demod_700D_AWGN_test`

1. freezes at frame 5 on `ff8c33e3` (last commit in #206)
1. runs to completion and passes at 7b8c259a:
```
ommit 7b8c259a89f36b5cdf354ad7885a7aad6d793def (HEAD)
Author: drowe67 <david@rowetel.com>
Date:   Tue Aug 10 13:28:18 2021 +0930

    fix sanity check assert to make sure #bits in frame adds up
```
So it looks like something that snuck it at the end of that PR.  Can't believe we missed that!

srsampson:
I've been putting a bunch of ifndef __EMBEDDED__ in the filter code, which should have unmapped a ton of RAM use (malloc etc). Which makes me think it isn't memory exhaustion.

srsampson:
Actually, I think I made things worse! What!??

master           = 1423916 sm1000v5.elf
ss-savebytes = 1427152 sm1000v5.elf

tmiw:
Forcing BPF to 0 (effectively reverting https://github.com/drowe67/codec2/pull/206/commits/4686212c15a5ee753ac37700f31ed9263ac08d37) fixes the test. However, trying to bypass other code that uses tx_bpf instead doesn't seem to work. Hmm...

srsampson:
> Forcing BPF to 0 (effectively reverting [4686212](https://github.com/drowe67/codec2/commit/4686212c15a5ee753ac37700f31ed9263ac08d37)) fixes the test. However, trying to bypass other code that uses tx_bpf instead doesn't seem to work. Hmm...

Good catch. The size of the filter structure seems so minimal, it must be just too compute intensive.

srsampson:
ctest --timeout 300 -R tst_api_demod_700D_plain_test

Still fails at Frame 3.

tmiw:
Actually, I have 100% of them passing on bear now with the latest changes. That said, I'm also not sure disabling BPF for != 700E is a great plan. Maybe we should figure out why enabling the BPF for 700D causes issues?

srsampson:
> Actually, I have 100% of them passing on bear now with the latest changes. That said, I'm also not sure disabling BPF for != 700E is a great plan. Maybe we should figure out why enabling the BPF for 700D causes issues?

Agree, my last change was just to implement what you found, and I hope we can find a way to keep filter on.

srsampson:
I removed the last change, which should allow BPF switching unmolested. Then I went into the unit test itself and started reducing the 10,000 byte memory whacking by 1000 until the test passed. They both passed when I got to 4,000 whacks.

So I'm showing all tests passing now.

srsampson:
Uploaded firmware to SM1000 and can run through all modes rx/tx and no hangs. Still need to run the audio into FreeDV to check the two-way. P.S. Best Regards to VK5DGR and K8BMA on excellent hardware design.

drowe67:
> I removed the last change, which should allow BPF switching unmolested. Then I went into the unit test itself and started reducing the 10,000 byte memory whacking by 1000 until the test passed. They both passed when I got to 4,000 whacks.

Good work @srsampson :+1: 

That's interesting, I wonder what's consuming the spare RAM?  Hmm, perhaps it is the heap we are allocating for the BPF colliding with the stack.  I forget how the RAM is laid out on this machine, but often one grows up while the other grows down.

Yes it would be a good idea to functionally test all modes in real time, although I suspect that's been covered in the testing done for https://github.com/drowe67/codec2/pull/206

srsampson:
From ARM Developer
```
...a better practice is to write portable code by declaring char variables appropriately.

Unsigned char must be used for accessing memory as a block of bytes or for small unsigned integers.

Signed char must be used for small signed integers and 

simple char must be used only for ASCII characters and strings...
```

drowe67:
Thanks @srsampson - sounds like good guidelines.  Any idea why the x86 ctest passed when `char` was used?  Or did `char` just happen to work for the x86 compiler?  I'd like to dig into why that test failed so we can make it more robust for next time.  I can look into this if you like, but it won't be for a while as I'm busy with another project.

We should also run freedv-gui and check the "var" metric on the LHS stats panel is reduced when the Auto-EQ is enabled.  This test is described here https://github.com/drowe67/codec2/pull/194.

srsampson:
> Thanks @srsampson - sounds like good guidelines. Any idea why the x86 ctest passed when `char` was used? Or did `char` just happen to work for the x86 compiler? I'd like to dig into why that test failed so we can make it more robust for next time. I can look into this if you like, but it won't be for a while as I'm busy with another project.
> 
> We should also run freedv-gui and check the "var" metric on the LHS stats panel is reduced when the Auto-EQ is enabled. This test is described here #194.

I looked around in forum comments, and it seems that the ARM compiler was specifically designed that way. The normal gcc treats it as a signed character if you leave "signed" off. My little test program worked fine for -20.

The reason I chose "char" was it didn't look like "int8_t" was used anywhere around the function, so I wanted to keep the code style the same. My fist instinct was to use "int8_t". 

You can also use "-fsigned-char" but I think that would be very confusing to programmers. 

Maybe need a new ctest to exercise EQ on or off?

drowe67:
> I looked around in forum comments, and it seems that the ARM compiler was specifically designed that way. The normal gcc treats it as a signed character if you leave "signed" off. My little test program worked fine for -20.

OK so that explains it - the x86 and ARM compiler were treating char differently (as Don suggested above https://github.com/drowe67/codec2/pull/230#issuecomment-926902392). So the x86 code (with char) was working fine, and the ctest verified it :heavy_check_mark: Even better - the stm32 ctest picked up the fact that the ARM compiler treated char differently.  So I think this suggests the ctests have a good enough coverage.  

My concern was that we could break the equaliser and the tests would not pick it up.

Thanks for digging into this one @srsampson.

> Maybe need a new ctest to exercise EQ on or off?

`test_700c_eq.sh` already does that I think.

So all that's left now is to try the EQ on freedv-gui,as per https://github.com/drowe67/codec2/pull/194

srsampson:
One of the comments in the ARM Developer guide was that using an 8-bit integer is much less efficient, and you should use 32-bit integers. This makes sense.

But in this case, I was trying to save storage bytes, and the computations are done later in floating point.

I played around a bit with loopback tonight, but I never got anywhere. I think I'll find my old headset sound card and just do it the normal way.

srsampson:
Here's a table form I stuck on my monitor :
```
ARM 'char' Programming Rules

char...............ASCII Characters and Strings Only
signed char........Range = -128 to +127 (2's Compliment)
unsigned char......Range = 0 to 255 (Data bytes)
```
These rules would need to be followed when developing  
software which is compiled on both ARM and x86.

donmr:
If you care about the size then the stdint types are better.  Chars can be 
bigger on a platform although that is uncommon.  int8_t and uint8_t are 
always 8 bits and always signed or unsigned respectively.

Don

srsampson:
Unexpected result, on FreeDV version 1.6.1 ```var``` reads 0.0 always, in modes 700D and 700E whether or not Auto-EQ is on or off. I'm driving it with a SM1000.  Guess I'll recompile FreeDV.

srsampson:
Sorry for my last, I'm getting different values with AutoEQ on and off, but they seem to settle to about the same numbers over time, say "testing 1 2 3" etc. There is a difference initially. Also getting different numbers with Speex on and off.

drowe67:
@srsampson Hmm, that doesn't sound right :thinking:  With freedv-gui as the Tx you should get a lower Var value with AutoEQ on, as per #194.

Using the SM1000 as a Tx, we could perform a similar test by printf-ing the variance to the serial port, if that doesn't affect real time operation.  It would be prudent to confirm the EQ is working on the SM1000, given the [subtle problems](https://github.com/drowe67/codec2/pull/230#issuecomment-926373543) we encountered above.  

As I mentioned above it's very hard to spot bugs in the audio processing code so we have to be super careful before merging.

srsampson:
My old Logitech souncard headset may be skewing the values...

Here we go using a trusted tool:

Values read using ```c2enc``` EQ On with ```ss-savebytes``` code:
```
$ ./c2enc 700C ../../raw/kristoff.raw test.c2 --var --eq
../../raw/kristoff.raw var: 10.47 std:  3.24
```
EQ Off
```
$ ./c2enc 700C ../../raw/kristoff.raw test.c2 --var
../../raw/kristoff.raw var: 12.63 std:  3.55
```
EQ On With ```master``` code:
```
$ ./c2enc 700C ../../raw/kristoff.raw test.c2 --var --eq
../../raw/kristoff.raw var: 10.47 std:  3.24
```
EQ Off:
```
$ ./c2enc 700C ../../raw/kristoff.raw test.c2 --var
../../raw/kristoff.raw var: 12.63 std:  3.55
```
Values are the same for various RAW input files.

drowe67:
Hi @srsampson.  Thanks for looking into it.  I'm not sure the update rate should matter too much,  I'm guessing the two versions should converge to about the same values.

I've just tried a few more tests.  Using codec2 8e54a628 and freedv-gui 24776be (#194) I'm starting to get more sensible results.  I let the sample loop a few times until the Var more or less converges, but its still bouncing around +/-0.2 or so when I log the result:

| Sample | freedv-gui EQ off | freedv-gui EQ On |  c2enc EQ off | c2enc EQ On | 
| kristoff.raw | 13.3 | 11.0 | 12.63 | 10.47 |
| ve9qrp_10.wav | 15.2 | 10.1 | 13.13 | 9. 24 |

When comparing `freedv-gui` and `c2enc` versions the absolute values are a bit different but the deltas look similar, e.g. a 2dB gain for kristoff, 4-5dB for ve9qrp_10s. I notice with with `.raw` files I get a random junk sound as freedv-gui rewinds, I'm not sure if that's upsetting things. 

Need to head out now, will think about this some more, and maybe try a few more git revisions (like back to freedv-gui master).

Man .. speech stuff is so much harder than modems!

drowe67:
Thanks @srsampson - curious that we get different results.  The signal path for this test is just reading from file, then sending to codec 2 encoder, sound cards aren't involved.  I reset to defaults and have the other EQs off.  

Ahh ok, I must have created the wave file version on ve9qrp_10s.

PR #232, opened 2021-10-01
_Complex support for swig parser (gnuradio)

ra1nb0w:


tmiw:
As in #233, perhaps we should use COMP instead of _Complex?

srsampson:
Doesn't GNU compilers work on Windows?

tmiw:
@srsampson, sure, but it would probably be nice not to be dependent on GNU compilers as a matter of good practice. Additionally, having it work well on Visual Studio would be good for encouraging developer adoption--assuming it's not a total pain to do so, of course.

Speaking of which, I keep meaning to play around with CMake's VS project generation functionality.

drowe67:
@srsampson @tmiw the issue of MSVC compatibility was discussed https://github.com/drowe67/codec2/pull/145 at some length.

For some reason the ctests haven't run for this PR or https://github.com/drowe67/codec2/pull/233 ?

srsampson:
I think the best course to follow, is to fork the repository, make all the changes you need to get the gnuradio monolith to compile and run using a proprietary closed source IDE, and then present your case.

Little changes to headers, week by week, will just cause extra work and won't move the ball into the end-zone.

tmiw:
FWIW, VS2019 does support using CMake files. I'm not 100% sure how it works exactly, so that'd need more investigation.

Anyway, I also tried just running "cmake" on a Windows VM and the default VS project file generation didn't successfully complete. I'll make a discussion post about it as it could be possible to generate project files for several different IDEs that way with hopefully low effort.

For now, I'm fine with the change in this PR if you guys are.

ra1nb0w:
The problem is swig and its C/C++ parser. Then, the fix depends on which version do you use: release or trunk. The former doesn't have https://github.com/swig/swig/issues/252 and https://github.com/swig/swig/pull/1770 and this is the reason for https://github.com/drowe67/codec2/pull/233 to work.

drowe67:
@ra1nb0w the need for `_Complex` has been removed in https://github.com/drowe67/codec2/pull/235

ra1nb0w:
thank you very much!

PR #233, opened 2021-10-01
Include <complex.h> in freedv_api.h since it now uses _Complex.

ryanvolz:
This was needed to fix compiling GNU Radio against `codec2` 1.0.1 (as with #232, but for GNU Radio 3.9 instead of 3.8!).

There still remains an issue with compiling GNU Radio with MSVC on Windows against this header, since MSVC doesn't support `complex` or `_Complex` with `#include <complex.h>` (grrrrrr), but I'm not sure what to do about that yet.

### DIFF
--- a/src/freedv_api.h
+++ b/src/freedv_api.h
@@ -33,6 +33,7 @@
 #ifndef __FREEDV_API__
 #define __FREEDV_API__
 
+#include <complex.h>
 #include <sys/types.h>
 // This declares a single-precision (float) complex number
 #include "comp.h"


tmiw:
Would it be better to use COMP instead of _Complex? Supposedly both types are stored the same way in memory but I'm not 100% sure if that can always be relied on. Thoughts on that would be appreciated.

ryanvolz:
> Would it be better to use COMP instead of _Complex? Supposedly both types are stored the same way in memory but I'm not 100% sure if that can always be relied on. Thoughts on that would be appreciated.

Yeah, from the MSVC build issue perspective, using a custom-defined type like that would work around the problem of Microsoft's poor compliance with C99. Otherwise, just adding the header makes the sane compilers happy :). If it's a more natural solution to switch to `COMP`, then go for it!

tmiw:
I'm fine with just doing the header change for now. Moving to COMP might be a more long-term measure. Agree @drowe67?

drowe67:
> I'm fine with just doing the header change for now. Moving to COMP might be a more long-term measure. Agree @drowe67?

I'd prefer to remove the need for `_Complex` in `freedv_api.h`, to avoid the need for two complex number types in the API.  IIRC we assume `COMP == _Complex` already in many parts of the OFDM modem code, as for historical reasons we move back and forth between the types quite a bit.

Would settling on just `COMP` also remove the need for #232 ?

srsampson:
> I'd prefer to remove the need for `_Complex` in `freedv_api.h`

I think this is the correct way also. The API should all be COMP (standardized), and the functions can easily get to/from _Complex from that.

drowe67:
@ryanvolz the need for `_Complex` has been removed in https://github.com/drowe67/codec2/pull/235

PR #235, opened 2021-10-02
moved rx_sym callback to freedv_api_internal.h

drowe67:
Attempt to remove the need for `_Complex` support in `freedv_api.h`, and hopefully remove the need for #232 and #233

We probably don't need this API function and it's typedef exposed to general API users, as it's pretty much a custom hook for the reliable text system.

@tmiw is this a reasonable fix?  I couldn't see this being used anywhere else but internally in `libcodec2`.  Sorry if I've missed something ...


tmiw:
Seems okay to me 👍

PR #239, opened 2021-10-10
Embedded ARM fixes and optimizations

tmiw:
I've been borrowing a Flex 6300 to get the FreeDV waveform working again. It's not 100% yet but I do have the following Codec2 changes so far:

1. Integrate the build fixes originally forked off at https://github.com/mcdermj/codec2.git (minus the cross compile stuff as that's already implemented in master).
2. Improve performance on ARM for the BPF by using CMSIS intrinsics (and thus NEON). I *think* this works but I'll need to try on another ARM based system to be sure (e.g. M1 Macs or maybe SM1000 if we can spare the stack/heap).
3. Get FDV_ARM_MATH working again to enable CMSIS use elsewhere in Codec2.

(CC @mcdermj FYI.)

### DIFF
--- a/src/mbest.c
+++ b/src/mbest.c
@@ -122,21 +122,20 @@ void mbest_search(
 )
 {
    float   e;
-   int     j;
 
-   for(j=0; j<m; j++) {
+   for(int j=0; j<m; j++) {
         float   diff;
         int i;
 
-	e = 0.0;
-	for(i=0; i<k; i++) {
-	    diff = cb[j*k+i]-vec[i];
-	    e += diff*w[i]*diff*w[i];
-	}
+        e = 0.0;
+        for(int i = 0; i < k && e < mbest->list[mbest->entries - 1].error; i++) {
+            diff = cb[j*k+i]-vec[i];
+            e += diff*w[i]*diff*w[i];
+        }
 
-	index[0] = j;
+        index[0] = j;
         if (e < mbest->list[mbest->entries - 1].error)
-	    mbest_insert(mbest, index, e);
+            mbest_insert(mbest, index, e);
    }
 }
 


drowe67:
Another thought is how we maintain and automatically test code across various platforms, for example anything that is platform dependent such as `#ifndef STM32F40_41xxx` will need testing on multiple platforms.  We've seen some divergence on the stm32 and x86 results at times, indeed it sometimes helps us pick up bugs.

Not sure the best way to handle this :thinking: If possible, the same C code on all platforms is best.  Lots of conditional defines also makes the code hard to follow and maintain.  However I appreciate there are some times when conditionals makes sense.  Working out a way to automate testing on `stm32` and any other platforms would also be great.

drowe67:
Hmm, I wonder why `ofdm_sync_search_core` is running, I think that's on the rx side.

Can you run the command line programs e.g. `freedv_tx` from the console?  That's a pretty good way to measure CPU.

Sounds like a tough environment to work with, and I understand the Flex CPU used for 3rd party waveforms is pretty low in terms of MIPs.  

BTW I wouldn't expect the 700D tx side to take much more CPU that 1600 (which I understand runs on Flex OK).  On my x86 laptop:
```
david@boomer:~/codec2/build_linux/src dr-reliable-text $ time ./freedv_tx 1600 ../../raw/ve9qrp.raw /dev/null
real	0m0.592s
david@boomer:~/codec2/build_linux/src dr-reliable-text $ time ./freedv_tx 700D ../../raw/ve9qrp.raw /dev/null
real	0m0.889s
```
Although the frame sizes are bigger for 700D, so more stack required.  Another problem with these really small processors is they have such a small amount of cache they choke on any serious DSP, as the arrays they are working on don't fit in the cache.

srsampson:
> @drowe67, since @srsampson closed his PR I'm going to recreate this PR without his changes to avoid messing up the merge later.

Yes, appologize, I don't have the experience with FreeDV and SM1000, I was targeting codec2 for the changes, but it looks like that part of the code is dependent on external programs.

tmiw:
No worries. It doesn't seem like that branch will impact the Flex anyway except for maybe the pow() removal, but will confirm when I have a chance.

amcdermond:
Just some comments on the platform here since @drowe67 mentioned it.

With the Flex Waveform API on flex6k, you're working with a TI DM8168 running Linux 2.6.  So, that'a an ARM Cortex-A8 for the main processor.  Waveforms do not have access to the C674x DSP processor, and it'd pretty difficult for us to facilitate that with the current architecture.  Remember that the ARM core also is running the main radio firmware, so if you start sucking up a lot of CPU to do waveform processing, you'll start starving the radio of CPU cycles to do important things like deliver VITA-49 packets to where they need to go.

For reference, I started with the SM1000 ARM optimizations purely from the aspect that they were there and a starting place.  I also tried to do some hand coding of some of the hot spots that profiling showed up to try to use GCC intrinsics to inject some NEON optimizations in there, but my code wasn't that great and it was slower than what the compiler produced.

Note that there is an errata on the DM8168 that causes the processor to lock up when using a VLD NEON instruction in protected memory. (see https://www.ti.com/lit/er/sprz329f/sprz329f.pdf?ts=1634587356375 page 64).  I don't think this should apply to any waveforms (it does to the Flex software team because the shared memory segment in which we pass DSP back and forth to the C674x DSP is protected memory), but if you start seeing weird lockups, this could be why.

I apologize for the late response on this.  I've been on vacation for the last ten days visiting New England.

Also, to dispel any confusion, @mcdermj is my personal GitHub account.  This is my FlexRadio account.

--
Annaliese McDermond
Senior DevOps Engineer
FlexRadio Systems

drowe67:
Suggest we set up some requirements and benchmark where we are today before diving into the optimisation.  Lets decide how far we need to go and in what direction before starting on the journey.
1. @amcdermond - can you pls tell us what % of the CPU is available for 3rd party waveforms, e.g. 25%, 50%?  A rough guess is fine.  Assume it's average CPU, and that the FreeDV code can be made to run nicely in the background.
1. @tmiw when convenient can you pls post benchmarks for freedv_tx & freedv_rx processing say ve9qrp (120 seconds) on the CPU so we have a baseline?

@tmiw rather than closing this PR another option is to revert the unwanted commits.

tmiw:
@drowe67, I created a new PR without those commits as reverting the merges in this one means that the changes in @srsampson's PR up to this point won't merge in when that PR finally reopens (as the revert commits would also end up in master otherwise). Seemed a lot easier to git cherry-pick my commits from this one into a new branch and continue the PR from that branch instead.

Anyway, I'll work on getting freedv_tx and _rx to compile. The current CMakeLists.txt in https://github.com/mcdermj/smartsdr-codec2/pull/17 only seems to generate libcodec2.a right now, so that'll need to be updated first.

PR #240, opened 2021-10-13
Make `rx_filter_coh()` static

xtkoba:
Downstream issue: https://bugs.gentoo.org/817437

### DIFF
--- a/src/cohpsk.c
+++ b/src/cohpsk.c
@@ -809,7 +809,7 @@ typedef float float4 __attribute__ ((vector_size (16)));
 
 \*---------------------------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


drowe67:
I also noticed he ctests haven't run automatically.  I wonder if that's because it's an external repo :thinking:

tmiw:
FWIW, "inline" actually did improve performance a bit for 700C on my system. It seems weird that it would be causing problems on some systems, though, so it's probably better to undo that for the time being.

drowe67:
> FWIW, "inline" actually did improve performance a bit for 700C on my system. It seems weird that it would be causing problems on some systems, though, so it's probably better to undo that for the time being.

Well if it's useful for us, and @xtkoba can't explain this PR as I requested above, I'm not inclined to merge it.  Perhaps if we understand the reasoning we can work out a way to satisfy every ones needs.  If there is no response to my question in another week, I'll close this PR.  We can always re open it later.

drowe67:
Closed as no response from OP to our questions.

PR #241, opened 2021-10-17
Add reference to reliable_text in the official FreeDV specs.

tmiw:
As the subject states. Probably a good idea to do a pass through too and make sure this is fully up to date in general.


tmiw:
@drowe67, does the FreeDV spec need any other updates as well?

tmiw:
Just pushed some basic API usage instructions. Do those sound reasonable?

PR #242, opened 2021-10-18
Embedded ARM fixes and optimizations

tmiw:
*Note: recreation of #239 to avoid merge issues later. See previous comments in that PR.*

I've been borrowing a Flex 6300 to get the FreeDV waveform working again. It's not 100% yet but I do have the following Codec2 changes so far:

1. Integrate the build fixes originally forked off at https://github.com/mcdermj/codec2.git (minus the cross compile stuff as that's already implemented in master).
2. ~Improve performance on ARM for the BPF by using CMSIS intrinsics (and thus NEON). I *think* this works but I'll need to try on another ARM based system to be sure (e.g. M1 Macs or maybe SM1000 if we can spare the stack/heap).~
2. Get FDV_ARM_MATH working again to enable CMSIS use elsewhere in Codec2.

(CC @mcdermj FYI.)


tmiw:
@drowe67, @amcdermond, as requested in #239, here are the results from running freedv_tx and freedv_rx with the current PR code:

```
root@c6a816x-evm:/FreeDV# time ./freedv_rx2 700D ve9qrp_processed.raw ve9qrp_decoded.raw
frames decoded: 702  output speech samples: 898560

real	0m15.456s
user	0m11.220s
sys	0m0.310s
root@c6a816x-evm:/FreeDV# time ./freedv_tx2 700D ve9qrp.raw ve9qrp_processed.raw

real	0m17.636s
user	0m10.320s
sys	0m0.630s
root@c6a816x-evm:/FreeDV#
```

gprof results:

[freedv_tx](https://github.com/drowe67/codec2/files/7378310/freedv_tx_current.txt)
[freedv_rx](https://github.com/drowe67/codec2/files/7378311/freedv_rx_current.txt)

drowe67:
Good work @tmiw  .... I feel we are very close :slightly_smiling_face: 

> > 1. Is the Flex API being driven at the correct levels?
> 
> It was actually driving lower than expected originally (resulting in ~30-40W out when the power level is set to full). I [added a 6dB increase](https://github.com/tmiw/smartsdr-codec2/blob/embedded-optimizations/src/freedv-processor.c#L399) to bump up the power level to something more reasonable. Removing that change didn't seem to make a difference, however.

Best to refer to the Flex spec for 3rd party waveform levels, e.g. in FreeDV API we use samples in 16 bit short range ( +/- 2^15).  Just wanted to check the possibility of us over driving and causing distortion.

> > 1. Maybe try `master`, and check the baseline CPU load and Tx signal quality. I'm wondering if a bug has crept in with the tx BPF optimisation.  I'm guessing it's hard to run the `ctests` on this platform and verify any changes with tests?
> 
> We probably wouldn't be able to go all the way back to master, but would going back to [81eff39](https://github.com/drowe67/codec2/commit/81eff3922c6bd7c8101a619ff2e692306a3e3c19) be reasonable? That should at least allow the code to build without any of the optimization changes.

How about the commit before?  It would be nice if can avoid merging per-device optimisations. They make `ofdm.c` more complex, harder to read, maintain, and most importantly automatically test.  So lets see if they are actually needed first.

@amcdermond - thanks, so looks like we are good from an average CPU point of view.  Cool.  I didn't understand the R/L term in your comment - can you (or @tmiw) pls point me at the Flex API documentation for 3rd party waveforms?  The FreeDV API has complex or real valued sample interfaces at 8 kHz, depending on the API call.

tmiw:
I had to go to 2ab504b to get a buildable codec2 (mainly due to needing to have a correct reference to arm_math.h).

```
root@c6a816x-evm:/FreeDV# time ./freedv_tx2 700D ve9qrp.raw ve9qrp_processed.raw

real	0m23.215s
user	0m20.830s
sys	0m0.390s
root@c6a816x-evm:/FreeDV# time ./freedv_rx2 700D ve9qrp_processed.raw ve9qrp_decoded.raw
frames decoded: 702  output speech samples: 898560

real	0m11.563s
user	0m8.760s
sys	0m1.360s
root@c6a816x-evm:/FreeDV# time ./freedv_rx2 700D random.raw random_decoded.raw
frames decoded: 702  output speech samples: 898560

real	0m4.916s
user	0m4.400s
sys	0m0.170s
root@c6a816x-evm:/FreeDV#
```

gprof outputs:

[freedv_rx_random_2ab504b.txt](https://github.com/drowe67/codec2/files/7384987/freedv_rx_random_2ab504b.txt)
[freedv_tx_2ab504b.txt](https://github.com/drowe67/codec2/files/7384988/freedv_tx_2ab504b.txt)
[freedv_rx_2ab504b.txt](https://github.com/drowe67/codec2/files/7384989/freedv_rx_2ab504b.txt)

If desired, I can remove __EMBEDDED__ and __REAL__ from the compiler options for freedv_tx and freedv_rx and try again with the requested commit.

drowe67:
@tmiw - thanks, looks like your Tx optimisation is certainly doing something, 50% improvements are indeed significant :+1:   However I feel both versions should still be fast enough for real time :thinking:, so the problem is more likely how FreeDV is talking to Flex:
1. Did https://github.com/drowe67/codec2/commit/2ab504b9559b7eb25f4cee019ba4dce87fa249f4 have the same issues with the Tx signal?  
1. Another option is to disable the BPF.
1. Can you pls email me  `ve9qrp_processed.raw` please, ie the modulated file generated on the Flex CPU?  I'll take a look on Octave.
1. Is it possible to run the resampling code @amcdermond mentioned, then dump the 24 kHz sampled signal back out to a file?

@amcdermond - thanks for that explanation, I'll poke around in the existing documentation and code and get back to you with any questions.  This discussion has also prompted me to discuss sample formats and levels in the FreeDV documentation :slightly_smiling_face: .  Sure, feel free to [email me](https://www.rowetel.com/?page_id=439) any drafts.

drowe67:
Thanks @tmiw for the samples emailed off line, I'll summarize here so we have a record to share:
1. `ve9qrp_processed.raw` (FreeDV 700D modulated signal at 8 kHz, int16) looks good to me in the time and frequency domain, no sign of that low pass DC, and decodes just fine.  This suggests the FreeDV encoding is working OK on the Flex CPU.
1. The 24 kHz mono sine wave (float32) also looks good, nice and clean just before the point it is packed into a VITA packet.  However the signal sample off air has some issues.

@amcdermond can you pls develop a simple demo using the API that shows a clean sine wave being transmitted using this API?  We could then reconcile that with what we have.

src/dump.c
```diff
@@ -32,7 +32,7 @@
 #include <string.h>
 #include <math.h>
 
-#ifdef __EMBEDDED__
+#ifdef STM32F40_41xxx
 #include "gdb_stdio.h"
```
drowe67, line 36: It would be nice to avoid an arch-dep conditionals.

Is the issue here that we need to differentiate between sm1000/stm32 and other embedded/small platforms?

drowe67:
@tmiw have the stm32 ctests passed?  Ill fire up `bear` just in case you need that machine.

src/dump.c
tmiw: Correct. gdb_stdio.h from what I can tell is only in the build environment used for SM1000 builds.

tmiw:
@drowe67, all tests pass on STM32 except for one:

```
The following tests FAILED:
	 23 - tst_api_demod_700D_AWGN_test (Timeout)
Errors while running CTest
```

I think it's an issue we've seen before, though.

tmiw:
FYI, e15e591 is needed to forego -fPIC when building Codec2 on microcontrollers (e.g. the Raspberry Pi Pico). It's why I was seeing weirdness with function calls and pointers that I emailed @drowe67 offline about. I'm open to alternatives as the [.got segment](https://en.wikipedia.org/wiki/Global_Offset_Table) doesn't seem to be getting included on SM1000 builds without that change.

drowe67:
> FYI, [e15e591](https://github.com/drowe67/codec2/commit/e15e591f5d369e680bbfd99206b377941177b514) is needed to forego -fPIC when building Codec2 on microcontrollers (e.g. the Raspberry Pi Pico). It's why I was seeing weirdness with function calls and pointers that I emailed @drowe67 offline about. I'm open to alternatives as the [.got segment](https://en.wikipedia.org/wiki/Global_Offset_Table) doesn't seem to be getting included on SM1000 builds without that change.

Perhaps lets wait to see how the Pico port goes before committing to supporting it in the codec2 build system.  I'm not sure it's necessary for the codec2 build system to support every platform, or something we want to commit to maintaining.

drowe67:
> tst_api_demod_700D_AWGN_test

I've raised a [PR](https://github.com/drowe67/codec2/pull/247) for this and I'll see if I can remove the need for [gdb_stdio](https://github.com/drowe67/codec2/pull/242#pullrequestreview-793473740)

tmiw:
> > FYI, [e15e591](https://github.com/drowe67/codec2/commit/e15e591f5d369e680bbfd99206b377941177b514) is needed to forego -fPIC when building Codec2 on microcontrollers (e.g. the Raspberry Pi Pico). It's why I was seeing weirdness with function calls and pointers that I emailed @drowe67 offline about. I'm open to alternatives as the [.got segment](https://en.wikipedia.org/wiki/Global_Offset_Table) doesn't seem to be getting included on SM1000 builds without that change.
> 
> Perhaps lets wait to see how the Pico port goes before committing to supporting it in the codec2 build system. I'm not sure it's necessary for the codec2 build system to support every platform, or something we want to commit to maintaining.

Still working through some USB to serial issues but hopefully I'll have some timings soon.

Regardless, fixed point might be nice to have long term but IIRC that's being discussed in one of the discussion threads.

src/dump.c
drowe67: OK this isn't needed now that https://github.com/drowe67/codec2/pull/247 is merged.

drowe67:
I wanted to summarise on this PR where we are at with Flex support of FreeDV.

@amcdermond - we are kind of stuck from our side.  Despite using 10-20% of the CPU real time operation doesn't seem to be possible.  On several other platforms (some with lower resources and much higher CPU load) we can run the same code smoothly in real time.

One clue: as @tmiw reduced CPU load via optimisations or disabling features, the signal break up issue improved, but was still there.  In most real time systems, once you get your code running in just under the available resources - you get smooth real time operation.  Further optimisation doesn't make it work better.  This suggests there is some process that is causing break up regardless of how fast the 3rd party waveform signal processing code is running.

What would be very helpful is a test/demo from Flex that demonstrates smooth real time operation without the FreeDV code.  For example transmitting a sine wave or other audio while the CPU is under a simulated 30% CPU load.

I appreciate you're pretty busy so no rush.  @tmiw and I may be able to help if you are resource constrained and can discuss operating under NDA if there is commercially sensitive IP involved - perhaps emails us directly re that.

tmiw:
FYI @drowe67, I had modified the code to transmit a sine wave before (see [here](https://github.com/tmiw/smartsdr-codec2/blob/embedded-optimizations/src/freedv-processor.c#L266)) and noted that I had no issues. I'm working on moving the Codec2 processing to a separate device that can be plugged into one of the USB ports on the back; if I can get that working with reasonable performance, I imagine we'll get similar results to the sine wave test. After all, serial port I/O shouldn't take much CPU time at all.

drowe67:
> FYI @drowe67, I had modified the code to transmit a sine wave before (see [here](https://github.com/tmiw/smartsdr-codec2/blob/embedded-optimizations/src/freedv-processor.c#L266)) and noted that I had no issues. 

Understood - however I imagine that was executing the payload 3rd party waveform (sine wave generation) in approximately 0 CPU.  I'm suggesting a test/demo from the Flex side that simulates a 3rd party waveform, including non-trival CPU.  There may be other ways to do this - a test sine wave is just one idea.  Clearly there is a gap in the sine wave demo you have demonstrated above and the inability of FreeDV to run on this platform.  I'm looking into a way to probe that gap, and feel there is value in exploring it from the Flex side.

>I'm working on moving the Codec2 processing to a separate device that can be plugged into one of the USB ports on the back; if I can get that working with reasonable performance, I imagine we'll get similar results to the sine wave test. After all, serial port I/O shouldn't take much CPU time at all.

Cool.  However I feel the FreeDV stack should also be able to run nicely in real time on the Flex CPU, if we can get to the bottom of the current issues.

> Anyway, I'll remove the changes to dump.c later today as they're no longer needed.

:+1:

tmiw:
> Understood - however I imagine that was executing the payload 3rd party waveform (sine wave generation) in approximately 0 CPU. I'm suggesting a test/demo from the Flex side that simulates a 3rd party waveform, including non-trival CPU. There may be other ways to do this - a test sine wave is just one idea. Clearly there is a gap in the sine wave demo you have demonstrated above and the inability of FreeDV to run on this platform. I'm looking into a way to probe that gap, and feel there is value in exploring it from the Flex side.

I could probably update the waveform to add additional load. Perhaps something where the waveform basically runs a while(1) {} and only performs its main tasks once every X iterations? Not sure what value of X to use here (10,000-100,000 maybe?)

> Cool. However I feel the FreeDV stack should also be able to run nicely in real time on the Flex CPU, if we can get to the bottom of the current issues.

Sure. Just more options in any case.

BTW the ESP32-S2 board I'm playing with might be a good candidate for use in an SM1000-like device. There's even a battery charger and Wi-Fi support so it's theoretically possible to interface it to radios with Wi-Fi support (e.g. the Icom IC-705). That's probably a ways off though.

> > Anyway, I'll remove the changes to dump.c later today as they're no longer needed.
> 
> 👍

Done.

drowe67:
> FYI @drowe67, I had modified the code to transmit a sine wave before (see [here](https://github.com/tmiw/smartsdr-codec2/blob/embedded-optimizations/src/freedv-processor.c#L266)) and noted that I had no issues. 

Another thought on this result.  It anchors the problem at one end - if the 3rd party waveform consumes roughly 0 CPU, the system works.  Another above that, and there is something breaking up the signal. This is consistent with what you noticed with optimisations - as optimisations were added, the glitches reduced, but didn't go away.   This suggests some process is occasionally consuming 100% CPU for enough time to prevent samples from the 3rd party waveform getting through.

tmiw:
> > FYI @drowe67, I had modified the code to transmit a sine wave before (see [here](https://github.com/tmiw/smartsdr-codec2/blob/embedded-optimizations/src/freedv-processor.c#L266)) and noted that I had no issues.
> 
> Another thought on this result. It anchors the problem at one end - if the 3rd party waveform consumes roughly 0 CPU, the system works. Anything above that, and there is something breaking up the signal. This is consistent with what you noticed with optimisations - as optimisations were added, the glitches reduced, but didn't go away. This suggests some process is occasionally consuming 100% CPU for enough time to prevent samples from the 3rd party waveform getting through.

Yep, although it's hard to see that for sure from running "top" even at 1x/sec.

BTW, am I right in assuming that we currently only use CMSIS in ofdm.c (and only if __EMBEDDED__ is defined)? On the ESP32 I've had to hack up the CMSIS CMake files a bit to get it to compile and link, so I'm wondering if that usage is really only to support the SM1000. If so, perhaps that could be reworked a bit to support non-ARM embedded platforms as well.

drowe67:
> Yep, although it's hard to see that for sure from running "top" even at 1x/sec.

Yes that's another interesting clue.  Does suggest the average load is quite low.  Might just need some redesign or buffering to distribute those peaks in CPU neatly.
  
> BTW, am I right in assuming that we currently only use CMSIS in ofdm.c (and only if __EMBEDDED__ is defined)? On the ESP32 I've had to hack up the CMSIS CMake files a bit to get it to compile and link, so I'm wondering if that usage is really only to support the SM1000. If so, perhaps that could be reworked a bit to support non-ARM embedded platforms as well.

Yes I think it's something @srsampson and I added for the SM1000 a few years ago.

tmiw:
> Yes that's another interesting clue. Does suggest the average load is quite low. Might just need some redesign or buffering to distribute those peaks in CPU neatly.

Yeah, I think it was averaging like 15% free CPU before. Still, enough from a RT thread could prevent the waveform from getting enough time to run well.

> Yes I think it's something @srsampson and I added for the SM1000 a few years ago.

Since the STM32 doesn't have NEON, I'm presuming CMSIS is simply doing loop unrolling of the dot product calculations. Could we write our own version that does similar and thus get rid of that architecture-specific optimization altogether? Though maybe that should be a separate PR, come to think of it.

tmiw:
So, I ended up rewriting large parts of the Flex waveform code tonight and it looks like I may have gotten it working without needing any external hardware. I'm having some others test it to be sure, though, but I think we're good to merge this PR now.

CMakeLists.txt
```diff
@@ -112,7 +112,7 @@ else()
 endif()
 
 # -fPIC is implied on MinGW...
-if(NOT WIN32)
+if((NOT WIN32) AND (NOT MICROCONTROLLER_BUILD))
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
```
drowe67, line 116: Is this still needed, given the experimental nano build didn't work out?

tmiw: Yep, it's still required for any Codec2 builds intended for microcontrollers. I just tried disabling MICROCONTROLLER_BUILD on the Teensy version of freedv_dongle and the test program wouldn't send/receive anything over USB. Reenabling MICROCONTROLLER_BUILD (which disables -fPIC) resolves the problem.

BTW, on Teensy 4.1 (600MHz Cortex M7 w/FPU):

```
$ time ./test -m 700D -t -p /dev/cu.usbmodem106388901 -i ~/devel/codec2/raw/ve9qrp_10s.raw -o tx.raw 

real	0m2.225s
user	0m0.005s
sys	0m0.022s
```

drowe67:
Re https://github.com/drowe67/codec2/pull/242#discussion_r747765379, would it be possible/useful to add an example of building and running codec2 on a microcontroller platform like the teensy?  I wouldn't know how and when to use `MICROCONTROLLER_BUILD`.

Perhaps a section in ~/codec2/README.md, I'm not sure if there is a better place for it.

tmiw:
@drowe67, I added a blurb about Codec2 use on microcontrollers. Hopefully that's sufficient?

tmiw:
I'm not sure we want to keep \_\_REAL\_\_ and FDV_ARM_MATH around long-term but I added explanations for those for now.

README.md
```diff
@@ -243,7 +243,7 @@ Codec 2 requires a hardware Floating Point Unit (FPU) to run in real time.
 Two build options have been added to support building on microcontrollers:
 1. Setting the `cmake` variable MICROCONTROLLER_BUILD disables position independent code (-fPIC is not used).  This was required for the IMRT1052 used in Teensy 4/4.1).
```
tmiw, line 244: Actually, removing -fPIC was required for every microcontroller I tried (Teensy, Pi Pico and ESP32). I suspect it's required for most if not all of them but then again, the SM1000 doesn't need this option so 🤷

src/codec2_fft.h
```diff
@@ -14,10 +14,11 @@
...
 
-#ifdef FDV_ARM_MATH
-  #include "fdv_arm_math.h"
+#ifndef FDV_ARM_MATH
+  #define USE_KISS_FFT
 #else
-    #define USE_KISS_FFT
+  #include "arm_math.h"
+  #include "arm_const_structs.h"
 #endif
```
drowe67, line 22: @tmiw can you pls tell me what these includes are for?

tmiw: These are the includes that were needed to have the CMSIS functions used by FDV_ARM_MATH work. I'm not sure if fdv_arm_math.h had anything else previously.

drowe67: Thanks.  So is `FDV_ARM_MATH` disabled for the stm32 build?  I think that's the case, I'm poking around the stm32 folders and can't see it enabled :thinking:

tmiw: Correct, it's not used for the SM1000.

drowe67: I ran the stm32 ctests, once I merged with master they passed OK :+1

PR #243, opened 2021-10-19
Change the variance process to possibly fine-tune FreeDV display

srsampson:
I created a c2->variance value in CODEC2 structure, and pre-compute the value every codec frame.

### DIFF
--- a/src/codec2.c
+++ b/src/codec2.c
@@ -205,7 +205,11 @@ struct CODEC2 * codec2_create(int mode)
     c2->xq_dec[0] = c2->xq_dec[1] = 0.0;
 
     c2->smoothing = 0;
-    c2->se = 0.0; c2->nse = 0;
+
+    c2->se = 0.0f;
+    c2->variance = 0.0f;
+    c2->nse = 0;
+    
     c2->user_rate_K_vec_no_mean_ = NULL;
     c2->post_filter_en = 1;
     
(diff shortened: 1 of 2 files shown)


srsampson:
The variance numbers are smoother displaying, but I don't think the differences are still great between EQ on and off.

The values do get larger now, as I'm seeing 10.5 as a value.

srsampson:
Alas, variance values are different with these changes:
```
./c2enc 700C ../../raw/ve9qrp_10s.raw test.c2 --var
../../raw/ve9qrp_10s.raw var: 13.13 std:  3.62

./c2enc 700C ../../raw/ve9qrp_10s.raw test.c2 --var --eq
../../raw/ve9qrp_10s.raw var:  9.24 std:  3.04
```
```
./c2enc 700C ../../raw/kristoff.raw test.c2 --var
../../raw/kristoff.raw var: 12.63 std:  3.55

./c2enc 700C ../../raw/kristoff.raw test.c2 --var --eq
../../raw/kristoff.raw var: 10.47 std:  3.24
```

drowe67:
@srsampson - could you pls describe in the title why we need this PR?  For example what problem does it solve or what bug does it fix?

drowe67:
Thanks for that explanation @srsampson - I'll take a closer look.  It might be useful for us to brainstorm ideas for PRs before too much coding is done, so we can agree on a plan and goals :slightly_smiling_face:

PR #246, opened 2021-10-28
Precalculate c for OFDM dft/idft.

tmiw:
This PR is to review the dft/idft optimizations originally introduced in #242.

*gprof output:*

[freedv_tx_dft.txt](https://github.com/drowe67/codec2/files/7431151/freedv_tx_dft.txt) (no difference w/master)
[freedv_tx_master.txt](https://github.com/drowe67/codec2/files/7431154/freedv_tx_master.txt) (no difference w/master)
[freedv_rx_dft.txt](https://github.com/drowe67/codec2/files/7431153/freedv_rx_dft.txt) (6.49us/call)
[freedv_rx_master.txt](https://github.com/drowe67/codec2/files/7431152/freedv_rx_master.txt) (11.69us/call)

I used the following calls for test: 

```
./freedv_tx 700D ve9qrp.raw ve9qrp_processed.raw
./freedv_rx 700D ve9qrp_processed.raw /dev/null
```

### DIFF
--- a/src/ofdm.c
+++ b/src/ofdm.c
@@ -401,6 +401,36 @@ struct OFDM *ofdm_create(const struct OFDM_CONFIG *config) {
         }
     }
 
+#ifndef STM32F40_41xxx   
+    // Precalculate "c" for use in dft() and idft().
+    ofdm->tx_c = (complex float*)calloc(ofdm->m * (ofdm->nc + 2), sizeof(complex float));
+    ofdm->rx_c = (complex float*)calloc(ofdm->m * (ofdm->nc + 2), sizeof(complex float));
+    assert(ofdm->tx_c != NULL && ofdm->rx_c != NULL);
+
+    complex float* c_ptr = ofdm->tx_c;
+    for (int row = 1; row < ofdm->m; row++) {
+        complex float c = cmplx(ofdm->tx_nlower * ofdm->doc * row);
+        complex float delta = cmplx(ofdm->doc * row);
+
+        for (int col = 0; col < (ofdm->nc + 2); col++) {
+            *c_ptr++ = c;
+            c *= delta;
+        }
+    }
+
+    c_ptr = ofdm->rx_c;
+    for (int col = 0; col < (ofdm->nc + 2); col++) {
+        float tval = (ofdm->rx_nlower + col) * ofdm->doc;
+        complex float c = cmplxconj(tval);
+        complex float delta = c;
+
+        for (int row = 1; row < ofdm->m; row++) {
+            *c_ptr++ = c;
+            c *= delta;
+        }
+    }
+#endif // STM32F40_41xxx
+
     /* default settings of options and states */
 
     ofdm->verbose = 0;
@@ -529,7 +559,7 @@ struct OFDM *ofdm_create(const struct OFDM_CONFIG *config) {
         ofdm_generate_preamble(ofdm, ofdm->tx_postamble, 3);
     }
     ofdm->postambledetectoren = !strcmp(ofdm->data_mode,"burst");
-    
+
     return ofdm; /* Success */
 }
 
@@ -563,6 +593,11 @@ static void deallocate_tx_bpf(struct OFDM *ofdm) {
 
 void ofdm_destroy(struct OFDM *ofdm) {
     int i;
+
+#ifndef STM32F40_41xxx   
+    free(ofdm->tx_c);
+    free(ofdm->rx_c);
+#endif // STM32F40_41xxx
 
     if (strlen(ofdm->data_mode)) {
         free(ofdm->tx_preamble);
@@ -606,15 +641,25 @@ static void idft(struct OFDM *ofdm, complex float *result, complex float *vector
 
     result[0] *= ofdm->inv_m;
 
+#ifndef STM32F40_41xxx
+    complex float* c_ptr = ofdm->tx_c;
+#endif // STM32F40_41xxx
+
     for (row = 1; row < ofdm->m; row++) {
+#ifdef STM32F40_41xxx
         complex float c = cmplx(ofdm->tx_nlower * ofdm->doc *row);
         complex float delta = cmplx(ofdm->doc * row);
+#endif // STM32F40_41xxx
 
         result[row] = 0.0f;
 
         for (col = 0; col < (ofdm->nc + 2); col++) {
+#ifdef STM32F40_41xxx
             result[row] += (vector[col] * c);
             c *= delta;
+#else
+            result[row] += vector[col] * (*c_ptr++);
+#endif // STM32F40_41xxx
         }
 
         result[row] *= ofdm->inv_m;
@@ -633,14 +678,24 @@ static void dft(struct OFDM *ofdm, complex float *result, complex float *vector)
         result[col] = vector[0];                 // conj(cexp(j0)) == 1
     }
 
+#ifndef STM32F40_41xxx
+    complex float* c_ptr = ofdm->rx_c;
+#endif // STM32F40_41xxx
+
     for (col = 0; col < (ofdm->nc + 2); col++) {
+#ifdef STM32F40_41xxx
         float tval = (ofdm->rx_nlower + col) * ofdm->doc;
         complex float c = cmplxconj(tval);
         complex float delta = c;
+#endif // STM32F40_41xxx
 
         for (row = 1; row < ofdm->m; row++) {
+#ifdef STM32F40_41xxx
             result[col] += (vector[row] * c);
             c *= delta;
+#else
+            result[col] += vector[row] * (*c_ptr++);
+#endif // STM32F40_41xxx
         }
     }
 }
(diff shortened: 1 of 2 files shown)


drowe67:
@srsampson do you have any comments on this PR?

@tmiw I'd like to avoid arch specific optimisations unless there is a really good reason.  Could we use the code in the PR all the time?

This optimisation deals with the ofdm modem, to best to use the ofdm test programs to characterise it, quite a few example command lines in README_ofdm.md, results for my machine:
```
$ git checkout master
$ make ofdm_mod ofdm_demod
$ time ./src/ofdm_mod --in /dev/zero --testframes 100 | ./src/ofdm_demod --out /dev/null --testframes

real	0m0.306s
user	0m0.449s
sys	0m0.000s

$ git checkout ms-optimization-dft
$ make ofdm_mod ofdm_demod
$ time ./src/ofdm_mod --in /dev/zero --testframes 100 | ./src/ofdm_demod --out /dev/null --testframes

real	0m0.264s
user	0m0.366s
sys	0m0.008s
```
I guess its running in two CPUs so real < user, looks like a 20% improvement :+1:

tmiw:
@drowe67, it was restricted to non-SM1000 due to concerns over the amount of heap available. If we do have enough, though, I don't see why this couldn't be enabled for all builds.

srsampson:
I tested with SM1000 and it fails the tst_ofdm_demod_quick test.

So I created a cval.h file with the data as it would go in constant flash.

The test worked after that.

Alas, you'd need a constant file for every Rs, Nc, and m combo. I just tried ```19``` and ```144``` for mode 700D.

Pretty good performance increase out of the old clunky DFT. Who needs an FFT with CPU's today? ha.

srsampson:
PS: the cval.h file I created was 5,400+ lines of complex numbers.

```
$ out 2>cval.h
```

Code:

```
#include <complex.h>
#include <math.h>
#include <stdio.h>

#define TAU         (2.0f * M_PI)

#define cmplx(value) (cosf(value) + sinf(value) * I)
#define cmplxconj(value) (cosf(value) + sinf(value) * -I)


int main(int argc, char **argv) {
    complex float tx_c[144 * 19];
    complex float rx_c[144 * 19];

    float doc = (TAU / (8000.0f / (1.0f / .018f)));

    float tval = ((float) 17.0f / 2.0f);
    float tx_nlower = roundf((1500.0f / (1.0f / .018f)) - tval) - 1.0f;
    float rx_nlower = roundf((1500.0f / (1.0f / .018f)) - tval) - 1.0f;

    fprintf(stderr,"const complex float tx_c[] = {\n");

    complex float* c_ptr = tx_c;

    for (int row = 1; row < 144; row++) {
        complex float c = cmplx(tx_nlower * doc * row);
        complex float delta = cmplx(doc * row);

        for (int col = 0; col < 19; col++) {
            *c_ptr++ = c;
            fprintf(stderr,"%ff %+ff * I,\n", crealf(c), cimagf(c));
            c *= delta;
        }
    }

    fprintf(stderr,"};\n\nconst complex float rx_c[] = {\n");

    c_ptr = rx_c;

    for (int col = 0; col < 19; col++) {
        float tval = (rx_nlower + col) * doc;
        complex float c = cmplxconj(tval);
        complex float delta = c;

        for (int row = 1; row < 144; row++) {
            *c_ptr++ = c;
            fprintf(stderr,"%ff %+ff * I,\n", crealf(c), cimagf(c));
            c *= delta;
        }
    }

    fprintf(stderr,"};\n\n");
}
```

drowe67:
> I tested with SM1000 and it fails (never completes until timeout) the tst_ofdm_demod_quick test.

Thanks for those tests @srsampson - do you know why it failed?  I'd prefer to create the table on the fly. Perhaps it was memory if it requires 5400 complex constants (20k) :thinking:

srsampson:
I tried reducing the 4000 number in tst_api_demod.c to 1000 and it still looked like it was going to timeout.

I didn't investigate further as I'm sure it is out of memory. So I tried to see if moving it to flash would work. It passed then, but I remembered that 700E would need a different table, so it wasn't very practical.

Maybe need to search for other common data to move to constant flash memory.

tmiw:
FWIW, the changes in this PR do work on the Pi Pico. However, I'm only seeing about a ~10 second improvement (and when it's taking about 2 minutes to process a 10 second audio clip, that's not enough). Definitely need a FPU for anywhere near reasonable performance.

srsampson:
I tried an idea of converting the table to shorts:
```
    ofdm->tx_c = (int16_t*)calloc(ofdm->m * (ofdm->nc + 2), sizeof(int16_t) * 2);       // COMP
    ofdm->rx_c = (int16_t*)calloc(ofdm->m * (ofdm->nc + 2), sizeof(int16_t) * 2);
    assert(ofdm->tx_c != NULL && ofdm->rx_c != NULL);

    int16_t* c_ptr = ofdm->tx_c;
    
    for (int row = 1; row < ofdm->m; row++) {
        complex float c = cmplx(ofdm->tx_nlower * ofdm->doc * row);
        complex float delta = cmplx(ofdm->doc * row);

        for (int col = 0; col < (ofdm->nc + 2); col++) {
            *c_ptr++ = (int16_t)(crealf(c) * 32767.);
            *c_ptr++ = (int16_t)(cimagf(c) * 32767.);
            c *= delta;
        }
    }
```
same changes to rx_c table...

The tests all passed on Ubuntu, but 18 tests failed on STM32. For example ```tst_ofdm_mod_plain``` failed for:
```
      rms error = 113103.149141
```
Just an idea... Don't know why they passed in Ubuntu and not on STM32.

drowe67:
Thanks @srsampson for taking a look at this one.

OK so thinking about this one against a couple of criteria:
1. A 20% CPU gain is nice, so good to have if it ticks other boxes.
1. Code quality and maintenance: The arch specific compiles aren't desirable as it means maintaining two code paths and makes the code harder to follow.
1. It works well on large machines, but on small machines we have a memory problems.  We could maybe get it to work on small memory machine using ROM tables (if they have enough flash) but this is mode-specific and also means maintaining another code path.
1. On large machines that have enough memory, we tend to have plenty of CPU.  The 100 frames I tested above was 18 seconds of 700D.
1. There are no use cases (with the possible exception of Flex) that need this optimisation right now.

So I'm not sure about this one.  Just because we can - should we? There may be other techniques that give us a similar optimisation (e.g. related to FFTs, or trig recursions).

tmiw:
Could we perhaps restructure the #ifdefs as "low memory" ones instead (e.g. -DLOW_MEMORY=1 would use the alternate path when passed to gcc)? That would make them less architecture specific and more based on properties of the system it'll run on.

(IIRC that specific #ifdef currently isn't used much in the code so there shouldn't be much rework involved if we go that route.)

srsampson:
I'm not really into monoliths, so I think the Flex radio should do one thing well, and then provide the IO for something else to do their thing well. Otherwise you might as well put an Intel Core I5 in there, with 16 Gig of RAM, and not a microprocessor better suited to a kitchen appliance. I have one of their 3000 models, and it required Firewire and Windows 7. Since I moved away from Windows I've been tempted to salvage enough parts out of it to interface it to an Arduino.

tmiw:
> I'm not really into monoliths, so I think the Flex radio should do one thing well, and then provide the IO for something else to do their thing well. Otherwise you might as well put an Intel Core I5 in there, with 16 Gig of RAM, and not a microprocessor better suited to a kitchen appliance. I have one of their 3000 models, and it required Firewire and Windows 7. Since I moved away from Windows I've been tempted to salvage enough parts out of it to interface it to an Arduino.

One of the things I'm hoping to get working is some sort of "accelerator" dongle for FreeDV that can plug into the USB ports on the back. The waveform that runs on the Flex itself would simply communicate with the dongle and have the latter do all the heavy lifting. Unfortunately, the Raspberry Pi Pico definitely wasn't that device (at least as long as there's no fixed-point port), but I'm going to try an ESP32-S2 based device next.

tmiw:
BTW, I do have the previously mentioned dongle code in https://github.com/tmiw/freedv_dongle.

But yeah, have we figured out how we want to proceed with this PR?

drowe67:
> But yeah, have we figured out how we want to proceed with this PR?

I'm not inclined to merge atm for the [reasons explained above](https://github.com/drowe67/codec2/pull/246#issuecomment-955784209).  Happy to keep it open for now until the Flex situation resolves.

tmiw:
Actually, thinking more, would there be possibly room on the SM1000 to only have one of the c arrays cached at a time? After all, usually we call either freedv_tx or _rx repeatedly (e.g. not interleaving them), so it shouldn't be too big of a performance impact to cache the values right when we switch between RX and TX.

drowe67:
@tmiw sorry but I think I'll close this PR (umerged) for now.  If the need for optimisation in this area comes up again we can revisit and try to address the concerns I outlined above.  I do appreciate your work in this area :slightly_smiling_face:

PR #256, opened 2021-11-08
MSYS2/MinGW Compiler problem using alloca.h

srsampson:
These changes modify two ```fsk``` files using ```alloca()``` and changes them to ```malloc()/free()```.

The MSYS compiler header file ```stdlib.h``` also includes ```malloc.h```.

The use of ```alloca()``` would require a ```#ifdef```around the ```#include <alloca.h>```

### DIFF
--- a/unittest/test_700c_eq.sh
+++ b/unittest/test_700c_eq.sh
@@ -4,9 +4,9 @@
 
 results=$(mktemp)
 
-../build_linux/src/c2enc 700C ../raw/kristoff.raw /dev/null --var 2> $results
+c2enc 700C ../raw/kristoff.raw /dev/null --var 2> $results
 var=$(cat $results | sed -n "s/.*var: \([0-9..]*\) .*/\1/p")
-../build_linux/src/c2enc 700C ../raw/kristoff.raw /dev/null --var --eq 2> $results
+c2enc 700C ../raw/kristoff.raw /dev/null --var --eq 2> $results
 var_eq=$(cat $results | sed -n "s/.*var: \([0-9..]*\) .*/\1/p")
 printf "var: %5.2f var_eq: %5.2f\n" $var $var_eq
 python -c "import sys; sys.exit(0) if $var_eq<=$var else sys.exit(1)"
--- a/unittest/test_700c_eq.sh
+++ b/unittest/test_700c_eq.sh
@@ -4,9 +4,9 @@
 
 results=$(mktemp)
 
-c2enc 700C ../raw/kristoff.raw /dev/null --var 2> $results
+../build_linux/src/c2enc 700C ../raw/kristoff.raw /dev/null --var 2> $results
 var=$(cat $results | sed -n "s/.*var: \([0-9..]*\) .*/\1/p")
-c2enc 700C ../raw/kristoff.raw /dev/null --var --eq 2> $results
+../build_linux/src/c2enc 700C ../raw/kristoff.raw /dev/null --var --eq 2> $results
 var_eq=$(cat $results | sed -n "s/.*var: \([0-9..]*\) .*/\1/p")
 printf "var: %5.2f var_eq: %5.2f\n" $var $var_eq
 python -c "import sys; sys.exit(0) if $var_eq<=$var else sys.exit(1)"
--- a/src/fmfsk.c
+++ b/src/fmfsk.c
@@ -196,7 +196,7 @@ void fmfsk_demod(struct FMFSK *fmfsk, uint8_t rx_bits[],float fmfsk_in[]){
     memcpy (&oldsamps[nold], &fmfsk_in[0]        , sizeof(float)*nin );
     
     /* Allocate memory for filtering */
-    float *rx_filt = alloca(sizeof(float)*(nsym+1)*Ts);
+    float *rx_filt = malloc(sizeof(float)*(nsym+1)*Ts);
     
     /* Integrate over Ts input symbols at every offset */
     for(i=0; i<(nsym+1)*Ts; i++){
@@ -364,4 +364,6 @@ void fmfsk_demod(struct FMFSK *fmfsk, uint8_t rx_bits[],float fmfsk_in[]){
     
     modem_probe_samp_f("t_norm_rx_timing",&norm_rx_timing,1);
     modem_probe_samp_f("t_rx_filt",rx_filt,(nsym+1)*Ts);
+
+    free(rx_filt);
 }
--- a/src/fmfsk_demod.c
+++ b/src/fmfsk_demod.c
@@ -28,6 +28,7 @@
 */
 
 #include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include "fmfsk.h"
 #include "modem_stats.h"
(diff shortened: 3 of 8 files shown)


srsampson:
I did a grep in ```src``` and these files also use ```alloca()``` and don't include ```alloca.h```:
```
fmfsk.c:    float *rx_filt = alloca(sizeof(float)*(nsym+1)*Ts);
fmfsk_demod.c:    bitbuf = (uint8_t*)alloca(sizeof(uint8_t)*fmfsk->nbit);
fmfsk_demod.c:    rawbuf = (int16_t*)alloca(sizeof(int16_t)*(fmfsk->N+fmfsk->Ts*2));
fmfsk_demod.c:    modbuf = (float*)alloca(sizeof(float)*(fmfsk->N+fmfsk->Ts*2));
fmfsk_mod.c:    bitbuf = (uint8_t*)alloca(sizeof(uint8_t)*fmfsk->nbit);
fmfsk_mod.c:    rawbuf = (int16_t*)alloca(sizeof(int16_t)*fmfsk->N);
fmfsk_mod.c:    modbuf = (float*)alloca(sizeof(float)*fmfsk->N);
freedv_fsk.c:    tx_float = alloca(sizeof(float)*f->n_nom_modem_samples);
```
As far as I can tell, using ```alloca()``` under gcc linux, should include ```alloca.h```, however the compiler may include this header when setting ```__USE_MISC``` in ```features.h```:
```
#if defined _BSD_SOURCE || defined _SVID_SOURCE
# define __USE_MISC     1
#endif
```
Evidently using ```alloca()``` doesn't seem to create a warning in all cases, given the files above seem to work fine.

srsampson:
More compiler info from GNU:
```
Macro: _DEFAULT_SOURCE
If you define this macro, most features are included apart from X/Open, LFS and GNU extensions:   
the effect is to  enable features from the 2008 edition of POSIX, as well as certain   
BSD and SVID features without a separate feature test macro to control them.

Be aware that compiler options also affect included features:

* If you use a strict conformance option, features beyond those from the compiler’s language   
version will be disabled, though feature test macros may be used to enable them.
* Features enabled by compiler options are not overridden by feature test macros.
```
Doesn't sound like a macro you would want to use just to get ```alloca.h``` included.

srsampson:
Added info from ```features.h```
```
#ifdef _GNU_SOURCE
# undef  _DEFAULT_SOURCE
# define _DEFAULT_SOURCE        1
#endif
```
I kind of edited that, but basically defining ```_GNU_SOURCE``` includes everything.

I haven't found out how the linker resolves the ```alloca.h``` if you don't include it yet. Maybe a bit of linker magic.

srsampson:
Note: these files are not used in SM1000 or STM32, so no testing was performed there.

srsampson:
alloca man page notes:
```
Normally, gcc translates calls to alloca() with inlined code.

This is not done when either the -ansi, -std=c89, -std=c99, or the -std=c11 option is given   
and the header <alloca.h> is not included.

Otherwise, (without an -ansi or -std=c* option) the glibc version of <stdlib.h> includes   
<alloca.h> and that contains the lines:

      #ifdef  __GNUC__
       #define alloca(size)   __builtin_alloca (size)
      #endif
```
I believe the make is using ```-std=c11``` or ```-std=c99``` for compiles:

In cmake file CMakeLists.txt:
```
# Check for what C standard is supported.
include(CheckCCompilerFlag)
CHECK_C_COMPILER_FLAG("-std=gnu11" COMPILER_SUPPORTS_GNU11)
CHECK_C_COMPILER_FLAG("-std=gnu99" COMPILER_SUPPORTS_GNU99)

if(COMPILER_SUPPORTS_GNU11)
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu11")
elseif(COMPILER_SUPPORTS_GNU99)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
else()
   message(SEND_ERROR "Compiler doesn't seem to support at least gnu99, might cause problems" )
endif()
```

drowe67:
> I did a grep in `src` and these files also use `alloca()` and don't include `alloca.h`:
> 
> ```
> fmfsk.c:    float *rx_filt = alloca(sizeof(float)*(nsym+1)*Ts);
> fmfsk_demod.c:    bitbuf = (uint8_t*)alloca(sizeof(uint8_t)*fmfsk->nbit);
> fmfsk_demod.c:    rawbuf = (int16_t*)alloca(sizeof(int16_t)*(fmfsk->N+fmfsk->Ts*2));
> fmfsk_demod.c:    modbuf = (float*)alloca(sizeof(float)*(fmfsk->N+fmfsk->Ts*2));
> fmfsk_mod.c:    bitbuf = (uint8_t*)alloca(sizeof(uint8_t)*fmfsk->nbit);
> fmfsk_mod.c:    rawbuf = (int16_t*)alloca(sizeof(int16_t)*fmfsk->N);
> fmfsk_mod.c:    modbuf = (float*)alloca(sizeof(float)*fmfsk->N);
> freedv_fsk.c:    tx_float = alloca(sizeof(float)*f->n_nom_modem_samples);

Thanks @srsampson - I guess it was Brady's style at the time (he did the C port of the FSK modem).  Please go ahead and remove the `alloca()` calls.

srsampson:
I had to make a change to:

```unittest/test_700c_eq.sh```

In order to get it to pass. FYI

File: unittest/test_700c_eq.sh [hunk 1]
```diff
@@ -4,9 +4,9 @@
 
 results=$(mktemp)
 
-c2enc 700C ../raw/kristoff.raw /dev/null --var 2> $results
+../build_linux/src/c2enc 700C ../raw/kristoff.raw /dev/null --var 2> $results
 var=$(cat $results | sed -n "s/.*var: \([0-9..]*\) .*/\1/p")
```
drowe67, line 8: That will break on systems that don't used `build_linux` or a different path to `c2sim`.  The path to `c2sim` should be set up in the ctest section of `CMakeLists.txt`:
```
     add_test(NAME test_700c_eq
             COMMAND sh -c "cd ${CMAKE_CURRENT_SOURCE_DIR}/unittest;
                            PATH=$PATH:${CMAKE_CURRENT_BINARY_DIR}/src ./test_700c_eq.sh")
```

This test has worked fine to date.  Can you pls describe why we need this change, or why it isn't working for you?

srsampson:
If I type ```c2enc``` on the command line on my system, it doesn't know where it is. When I type ```make test``` it's the only one that fails.

Somethings fishy, because my ```*.float``` in ```raw``` don't get built either. Anyway, I reverted the change, so will see how github handles it.

drowe67:
The $PATH to c2sim is set up in CMakeLists.txt.

`ctest -V -R test_700c_eq` will let you run this test in verbose mode and see what's going on with your machine.  It would be good to get to the bottom of it as it might be a subtle bug, like a path assumption.  You can also add commands to the script like "-x" at the top:
```
#!/bin/bash -x
```
 to echo each line as it executes.

Looks like it passes when run from GitHub actions (and on mine and other machines).

Re fading files, some of the `codec2/unittest` scripts look for the files and generate them if they are not there, e.g. `unittest/ofdm_fade.sh`, which is called when we run the `test_OFDM_modem_fading_BER` ctest.  Once again, running the ctest in verbose mode with bash "-x" will let you debug it.

srsampson:
The github build went well, so it's just a personal problem here, I have some mods in my /etc directory that may be at the bottom of it.

The malloc/free stuff is ready to merge and we can close this.

drowe67:
What do you think of these:

```
david@boomer:~/codec2/src ss-alloca $ grep "alloca.h" *
grep: codebook: Is a directory
_kiss_fft_guts.h:#include <alloca.h>
david@boomer:~/codec2/src ss-alloca $ grep "alloca(" *
grep: codebook: Is a directory
fsk.c:    kiss_fft_cpx *fftin  = (kiss_fft_cpx*)alloca(sizeof(kiss_fft_cpx)*Ndft);
fsk.c:    kiss_fft_cpx *fftout = (kiss_fft_cpx*)alloca(sizeof(kiss_fft_cpx)*Ndft);
_kiss_fft_guts.h:#define  KISS_FFT_TMP_ALLOC(nbytes) alloca(nbytes)
david@boomer:~/codec2/src ss-alloca $ 
```
1. I don't think `KISS_FFT_USE_ALLOCA` is defined, so that one probably doesn't matter.  Given it's a 3rd party library it's nice to make minimal changes to it.
1. However I think `DEMOD_ALLOC_STACK` might be set at the top of `fsk.c`?  I'm OK with removing all the `alloca()` code from `fsk.c`.

srsampson:
I saw those two, I didn't want to mess with the FFT like you, since it is not enabled anyway, but I've made the changes to fsk.c. It passed the tests here locally.

PR #264, opened 2021-12-09
Update CML patch to support build on Clang.

tmiw:
macOS can't build CML due to its use of Clang instead of GCC. This PR updates the patch inside Codec2 to enable building on systems that use Clang.

@drowe67, @hobbes1069, does this work any better for you guys?

### DIFF
--- a/octave/cml.patch
+++ b/octave/cml.patch
@@ -59,52 +59,6 @@ diff -uNr cml.orig/mat/InitializeWiMaxLDPC.m cml/mat/InitializeWiMaxLDPC.m
      P=inv(circshift( eye(z),[0,floor(80*z/z0)]));
  end
  
-Binary files cml.orig/mex/Capacity.dll and cml/mex/Capacity.dll differ
-Binary files cml.orig/mex/Capacity.mex and cml/mex/Capacity.mex differ
-Binary files cml.orig/mex/CapacityTableLookup.dll and cml/mex/CapacityTableLookup.dll differ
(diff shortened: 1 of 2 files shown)


drowe67:
@tmiw still plenty of warnings but it builds OK for me :+1: 

Actually I think it gets tested every time we [push to github](https://github.com/drowe67/codec2/blob/master/.github/workflows/cmake.yml#L35).

tmiw:
> @tmiw still plenty of warnings but it builds OK for me 👍
> 
> Actually I think it gets tested every time we [push to github](https://github.com/drowe67/codec2/blob/master/.github/workflows/cmake.yml#L35).

It doesn't look like it gets moved to ~/cml/ on push, unless I missed it at first glance?

drowe67:
> > @tmiw still plenty of warnings but it builds OK for me +1
> > Actually I think it gets tested every time we [push to github](https://github.com/drowe67/codec2/blob/master/.github/workflows/cmake.yml#L35).
> 
> It doesn't look like it gets moved to ~/cml/ on push, unless I missed it at first glance?

I think it's OK, I can see test 4 (CML ldpc ut) running in https://github.com/drowe67/codec2/runs/4466805746?check_suite_focus=true, plus a bunch of other tests that use the CML library.

drowe67:
@tmiw is it OK with you to merge this?

tmiw:
> @tmiw is it OK with you to merge this?

Yep, go for it. 👍

PR #265, opened 2021-12-10
FreeData TNC support

drowe67:
Some maintenance, new demos, and tests to support the [FreeData Xmas Sprint](https://github.com/DJ2LS/FreeDATA/discussions/87):

1. tnc1_high_snr.sh and tnc4_high_snr_ping.sh virtual sound card tests working with latest burst data modes. :heavy_check_mark: 
2. Clean up `freedv_data_raw_tx` to separate testframe and burst modes. :heavy_check_mark: 
3. Retire old streaming mode from `freedv_data_raw_tx` :heavy_check_mark: 
4. Demonstrate running two different raw data mode receivers at the same time. :heavy_check_mark: 
5. Add capability for frames_per_burst > 1 when using stdin back into `freedv_data_raw_tx`.
6. Dig up 48k-8k resamplers (float and short), write tests :heavy_check_mark:


drowe67:
```
time ctest -R test_demo_datac0c1
```
runs in 10s for 54s of data, so 18% of real time on my laptop.  The acquisition code is pretty slow and could use some speed ups such as vectorisation, however this should be enough to get us started.

PR #266, opened 2021-12-12
Fix bugs found with GCC 11

fmoessbauer:
This PR fixes bugs found when compiling with gcc 11.2.
Commits 1,2 fix array-size bugs (potential buffer overflows).
Commits 3,4 fix undefined behavior glitches.

### DIFF
--- a/src/cohpsk.c
+++ b/src/cohpsk.c
@@ -437,7 +437,7 @@ void qpsk_symbols_to_bits(struct COHPSK *coh, float rx_bits[], COMP ct_symb_buf[
 \*---------------------------------------------------------------------------*/
 
 void tx_filter_and_upconvert_coh(COMP tx_fdm[], int Nc, const COMP tx_symbols[],
-                                 COMP tx_filter_memory[COHPSK_NC*COHPSK_ND][COHPSK_NSYM],
+                                 COMP tx_filter_memory[][COHPSK_NSYM],
                                  COMP phase_tx[], COMP freq[],
                                  COMP *fbb_phase, COMP fbb_rect)
 {
--- a/src/cohpsk.c
+++ b/src/cohpsk.c
@@ -962,7 +962,7 @@ void rate_Fs_rx_processing(struct COHPSK *coh, COMP ch_symb[][COHPSK_NC*ND], COM
     int   r, c, i, ch_fdm_frame_index;
     COMP  rx_fdm_frame_bb[COHPSK_M+COHPSK_M/P];
     COMP  rx_baseband[COHPSK_NC*ND][COHPSK_M+COHPSK_M/P];
-    COMP  rx_filt[COHPSK_NC*ND][P+1];
+    COMP  rx_filt[COHPSK_NC*(ND+1)][P+1];
     float env[NT*P], rx_timing;
     COMP  rx_onesym[COHPSK_NC*ND];
     float beta, g;
--- a/src/cohpsk.c
+++ b/src/cohpsk.c
@@ -809,7 +809,7 @@ typedef float float4 __attribute__ ((vector_size (16)));
 
 \*---------------------------------------------------------------------------*/
 
-inline void rx_filter_coh(COMP rx_filt[COHPSK_NC+1][P+1], int Nc, COMP rx_baseband[COHPSK_NC+1][COHPSK_M+COHPSK_M/P], COMP rx_filter_memory[COHPSK_NC+1][+COHPSK_NFILTER], int nin)
+inline void rx_filter_coh(COMP rx_filt[COHPSK_NC*COHPSK_ND][P+1], int Nc, COMP rx_baseband[COHPSK_NC*COHPSK_ND][COHPSK_M+COHPSK_M/P], COMP rx_filter_memory[COHPSK_NC*COHPSK_ND][COHPSK_NFILTER], int nin)
 {
     int c,i,j,k,l;
     int n=COHPSK_M/P;
@@ -961,8 +961,8 @@ void rate_Fs_rx_processing(struct COHPSK *coh, COMP ch_symb[][COHPSK_NC*ND], COM
     struct FDMDV *fdmdv = coh->fdmdv;
     int   r, c, i, ch_fdm_frame_index;
     COMP  rx_fdm_frame_bb[COHPSK_M+COHPSK_M/P];
-    COMP  rx_baseband[COHPSK_NC*ND][COHPSK_M+COHPSK_M/P];
-    COMP  rx_filt[COHPSK_NC*(ND+1)][P+1];
+    COMP  rx_baseband[COHPSK_NC*COHPSK_ND][COHPSK_M+COHPSK_M/P];
+    COMP  rx_filt[COHPSK_NC*COHPSK_ND][P+1];
     float env[NT*P], rx_timing;
     COMP  rx_onesym[COHPSK_NC*ND];
     float beta, g;
--- a/src/cohpsk.c
+++ b/src/cohpsk.c
@@ -752,7 +752,7 @@ void cohpsk_clip(COMP tx_fdm[], float clip_thresh, int n)
 
 \*---------------------------------------------------------------------------*/
 
-void fdm_downconvert_coh(COMP rx_baseband[COHPSK_NC][COHPSK_M+COHPSK_M/P], int Nc, COMP rx_fdm[], COMP phase_rx[], COMP freq[], int nin)
+void fdm_downconvert_coh(COMP rx_baseband[][COHPSK_M+COHPSK_M/P], int Nc, COMP rx_fdm[], COMP phase_rx[], COMP freq[], int nin)
 {
     int   i,c;
     float mag;
--- a/src/cohpsk.c
+++ b/src/cohpsk.c
@@ -60,7 +60,7 @@ static COMP qpsk_mod[] = {
 static int sampling_points[] = {0, 1, 6, 7};
 
 void corr_with_pilots(float *corr_out, float *mag_out, struct COHPSK *coh, int t, float f_fine);
-void update_ct_symb_buf(COMP ct_symb_buf[][COHPSK_NC*ND], COMP ch_symb[][COHPSK_NC*ND]);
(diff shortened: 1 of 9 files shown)


src/cohpsk_internal.h
```diff
@@ -113,7 +113,7 @@ struct COHPSK {
 void bits_to_qpsk_symbols(COMP tx_symb[][COHPSK_NC*COHPSK_ND], int tx_bits[], int nbits);
 void qpsk_symbols_to_bits(struct COHPSK *coh, float rx_bits[], COMP ct_symb_buf[][COHPSK_NC*COHPSK_ND]);
 void tx_filter_and_upconvert_coh(COMP tx_fdm[], int Nc, const COMP tx_symbols[],
-                                 COMP tx_filter_memory[COHPSK_NC][COHPSK_NSYM],
+                                 COMP tx_filter_memory[COHPSK_NC*ND][COHPSK_NSYM],
```
tmiw, line 116: Shouldn't this be COHPSK_NC * COHPSK_ND to match line 114?

unittest/compare_ints.c
```diff
@@ -126,7 +127,7 @@ int main(int argc, char *argv[]) {
         uint64_t err = llabs(data1 - data2);
         if (err > tol) {
             errors ++;
-            printf("%d %lld %lld\n", count, data1, data2);
+            printf("%d %" PRId64 " %" PRId64 "\n", count, data1, data2);
```
tmiw, line 130: Seems weird that gcc10 doesn't support %lld, no?

src/cohpsk_internal.h
srsampson: It should be, as you say. The "ND" appears to be a redundant define, and not as specific as "COHPSK_ND". Probably should track down "ND" everywhere and update it.

srsampson: cohpsk_create() seems to use both and even compare each other in assert().

tmiw: If they're the same value, no point in having both. Otherwise, we need to figure out why they're different and at least document that accordingly.

src/cohpsk.c
```diff
@@ -962,7 +962,7 @@ void rate_Fs_rx_processing(struct COHPSK *coh, COMP ch_symb[][COHPSK_NC*ND], COM
     int   r, c, i, ch_fdm_frame_index;
     COMP  rx_fdm_frame_bb[COHPSK_M+COHPSK_M/P];
     COMP  rx_baseband[COHPSK_NC*ND][COHPSK_M+COHPSK_M/P];
-    COMP  rx_filt[COHPSK_NC*ND][P+1];
+    COMP  rx_filt[COHPSK_NC*(ND+1)][P+1];
     float env[NT*P], rx_timing;
```
drowe67, line 966: why ND+1?  ND is the number of diversity (copies of signal).  It can only be 2.

src/cohpsk_internal.h
drowe67: I can't remember why we have both 'ND' and 'COHPSK_ND'.  Hopefully the cests will pick up any issues if we decide to reconcile them

unittest/compare_ints.c
drowe67: Actually I'm seeing a warning for this when I build on gcc 9.3.0:
```
/home/david/codec2/unittest/compare_ints.c:129:27: warning: format ‘%lld’ expects argument of type ‘long long int’, but argument 3 has type ‘int64_t’ {aka ‘long int’} [-Wformat=]
  129 |             printf("%d %lld %lld\n", count, data1, data2);

```

drowe67: So it's not lack of support for %lld, just the wrong data type

drowe67:
Thanks @fmoessbauer they all look good except I'm puzzled by https://github.com/drowe67/codec2/pull/266/commits/9b4831fbab72dcdfdc05785446a0623bb3a9f466, as per [comment above](https://github.com/drowe67/codec2/pull/266#pullrequestreview-829808953). TBH I'm having trouble following the function that fills that array, as it's been several years and it has had several optimisations applied to it.  Do your tools give us any clue as to the potential buffer overflow?

@tmiw - do you have any comments? IIRC you worked on the vector optimisation of that function 6 months ago.

srsampson:
The +1 may be due to a probable type in:

https://github.com/drowe67/codec2/blob/master/src/cohpsk.c#L812

Which doesn't use anything other than NC internally.

I would correct line 965 and line 812 to [COHPSK_NC * COHPSK_ND][P + 1]

tmiw:
> The +1 may be due to a probable typo in:
> 
> https://github.com/drowe67/codec2/blob/master/src/cohpsk.c#L812
> 
> Which doesn't use anything other than ND internally.
> 
> I would correct line 965 and line 812 to [COHPSK_NC * COHPSK_ND][P + 1]

Shouldn't rx_filter_coh() use a pointer for rx_filt instead? Seems that the array sizing depends on the second parameter, really.

Or possibly get rid of the second parameter entirely since rx_filter_coh() is only called from one place in the code.

drowe67:
Thanks @tmiw and @srsampson - both those comment make sense to me and might be the source of https://github.com/drowe67/codec2/commit/9b4831fbab72dcdfdc05785446a0623bb3a9f466

unittest/compare_ints.c
fmoessbauer: > So it's not lack of support for %lld, just the wrong data type

That depends on the arch (on x64 `int64_t` normally maps to long, on i386 to long long). That's why you should use the macros to make it portable.

src/cohpsk.c
fmoessbauer: I just set it to that value to make the types "compatible" / same. GCC told me that the accessed buffer `rx_filt` is smaller than the type it is implicitly cast to. If it helps, I can build current master and past the warning here.

Unfortunately, I'm not deep enough into this implementation to understand the semantics of the buffer.

drowe67:
@srsampson - good point - pls feel free to add that comment.  I found instructions on pushing to someone elses PR [here](https://tighten.co/blog/adding-commits-to-a-pull-request/).

srsampson:
I upgraded the ND to COHPSK_ND in both src and unittest. I think I found them all.

drowe67:
Thanks @srsampson - I'm attempting to get the ctests to run automagically on Pull Requests like this.

drowe67:
@srsampson - yep it caught something.  Maybe `make` and `ctest` on your local machine to track it down :+1:

srsampson:
I must be going blind...

fmoessbauer:
With 321ccaae22315c73bc0cd96014be49a6b2e03624 we still have the following warning (same for 68edc0e):

```
[2/63] Building C object src/CMakeFiles/codec2.dir/cohpsk.c.o
../src/cohpsk.c: In function ‘rate_Fs_rx_processing’:
../src/cohpsk.c:986:21: warning: ‘rx_est_timing’ accessing 840 bytes in a region of size 560 [-Wstringop-overflow=]
  986 |         rx_timing = rx_est_timing(rx_onesym, fdmdv->Nc, rx_filt, fdmdv->rx_filter_mem_timing, env, nin, COHPSK_M);
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/cohpsk.c:986:21: note: referencing argument 3 of type ‘COMP (*)[5]’
In file included from ../src/cohpsk_internal.h:41,
                 from ../src/cohpsk.c:42:
../src/fdmdv_internal.h:186:7: note: in a call to function ‘rx_est_timing’
  186 | float rx_est_timing(COMP  rx_symbols[], int Nc,
      |       ^~~~~~~~~~~~~
[3/63] Building C object unittest/CMakeFiles/tcohpsk.dir/__/src/cohpsk.c.o
../src/cohpsk.c: In function ‘rate_Fs_rx_processing’:
../src/cohpsk.c:986:21: warning: ‘rx_est_timing’ accessing 840 bytes in a region of size 560 [-Wstringop-overflow=]
  986 |         rx_timing = rx_est_timing(rx_onesym, fdmdv->Nc, rx_filt, fdmdv->rx_filter_mem_timing, env, nin, COHPSK_M);
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/cohpsk.c:986:21: note: referencing argument 3 of type ‘COMP (*)[5]’
In file included from ../src/cohpsk_internal.h:41,
                 from ../src/cohpsk.c:42:
../src/fdmdv_internal.h:186:7: note: in a call to function ‘rx_est_timing’
  186 | float rx_est_timing(COMP  rx_symbols[], int Nc,
      |       ^~~~~~~~~~~~~
[63/63] Linking C executable misc/vq_mbest
```

srsampson:
rx_est_timing() and rx_filt() are used by two programs: fdmdv and cohpsk. These two call the functions with different size first-dimension. We can just delete the dimension size, as the C compiler can figure it out. There's many more examples of this, but cohpsk doesn't call them.

I'm not sure how this will fly on MSYS.

See if my changes have cleared-up your compiler warnings.

drowe67:
@srsampson thanks for that work.  Are their any compiler options we can enable to see the warnings @fmoessbauer is seeing?  Or perhaps we could just try gcc 11

srsampson:
> @srsampson thanks for that work. Are their any compiler options we can enable to see the warnings @fmoessbauer is seeing? Or perhaps we could just try gcc 11

It's new to me, but maybe the ```-Warray-bounds=2``` along with ```-O2``` would enable it in GCC 10?

srsampson:
OK, I've cleaned-up the warnings listed in this PULL. They were just warnings, not affecting anything functionally.

drowe67:
@srsampson - thanks that looks good. Did you work out a way to reproduce the warnings @fmoessbauer reported?  I just want to confirm the gcc 11 warnings are gone.

srsampson:
I installed GCC-11 on my Ubuntu workstation. I figured that would be the easiest route. Then cloned his repository and made the changes to clear up the warnings.

PR #275, opened 2021-12-29
Makefile for CML

drowe67:
`codec2` side of https://github.com/drowe67/cml/pull/3 - prevent CML building mex files on every `codec2 make`.

We'll need to remove the `unittest/CMakeLists.txt GIT_TAG dr-makefile` before merging.

@tmiw can you please try this and tell me if it works OK for you?


tmiw:
All good on macOS 12.0.1:

```
100% tests passed, 0 tests failed out of 111
```

PR #274, opened 2021-12-29
Experimental 2020A and 2020B Modes

drowe67:
Experimental 2020A and 2020B modes as discussed in https://github.com/drowe67/codec2/discussions/217.  Also see associated [LPCNet PR](https://github.com/drowe67/LPCNet/pull/42).

Octave simulation (MPP channel at 10dB SNR):
```
octave:68> ofdm_tx("test_2020A.raw", "2020A",60,10,'mpp')
octave:68> ofdm_rx("test_2020A.raw", "2020A")
``` 

C simulation (AWGN channel at 4dB SNR):
```
 ./src/ofdm_mod --in /dev/zero --testframes 10 --mode 2020A | ./src/cohpsk_ch - - -22 --Fs 8000 | ./src/ofdm_demod --mode 2020A --testframes -v1 > /dev/null
```
C simulation with compression, we can drop the noise level for similar BERs with the same peak modulator output:
```
./src/ofdm_mod --in /dev/zero --testframes 30 --mode 2020A --clip | ./src/cohpsk_ch - - -18 --Fs 8000 | ./src/ofdm_demod --mode 2020A --testframes -v1 > /dev/null
```
LPCNet passing through 2020A modem, on a MPP (1 Hz fading, 2ms delay) channel at around 14dB SNR.  Audio is rough but intelligible, also works on MPD channel (2 Hz fading, 4ms delay):
```
cat ~/LPCNet/wav/wia.wav | ~/LPCNet/build_linux/src/lpcnet_enc -s | ./ofdm_mod --mode 2020A -v1 --clip | ./cohpsk_ch - - -26 --Fs 8000 --mpp --fading_dir ../unittest | ./ofdm_demod --mode 2020A --verbose 1 | ~/LPCNet/build_linux/src/lpcnet_dec -s | aplay -f S16_LE -r 16000
```

TODO:
1. Waveform design on spreadsheet :heavy_check_mark: 
1. Octave and C simulations :heavy_check_mark: 
2. Compression for 2020A (and 2020) :heavy_check_mark: 
3. Run with LPCNet on command line :heavy_check_mark: 
[...]

PR #278, opened 2022-01-01
Channel simulator maintenance

drowe67:
1. Change name from `cohpsk_ch` to `ch` as it's used for all modems.
2. Make default Fs 8000
3. Use `getopt()` for command line arguments

`getopt` appears to choke on negative no-optional arguments like;
```
ch in.raw out.raw -30
```
So I had to change everything to:
```
ch in.raw out.raw --No -30
```
Hmm, on second thoughts that's not so bad, makes the usage a bit cleaner in some cases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -396,7 +396,7 @@ if(UNITTEST)
     # check channel simulator measures correct Peak to Average Power Ratio (about 0dB) with a sine wave input signal
     add_test(NAME test_ch_papr
              COMMAND sh -c "cd ${CMAKE_CURRENT_BINARY_DIR};
-                            ./misc/mksine - 1000 10 | ./src/ch - /dev/null -100  --ctest")
+                            ./misc/mksine - 1000 10 | ./src/ch - /dev/null --ctest")
 
     add_test(NAME test_codec2_700c_octave_port
              COMMAND sh -c "
@@ -418,11 +418,11 @@ if(UNITTEST)
              set_tests_properties(test_COHPSK_modem_octave_port PROPERTIES PASS_REGULAR_EXPRESSION "fails: 0")
 
     add_test(NAME test_COHPSK_modem_AWGN_BER
-             COMMAND sh -c "$<TARGET_FILE:cohpsk_get_test_bits> - 5600 | $<TARGET_FILE:cohpsk_mod> - - | $<TARGET_FILE:ch> - - -30 --Fs 7500 | $<TARGET_FILE:cohpsk_demod> - - | $<TARGET_FILE:cohpsk_put_test_bits> -"
+             COMMAND sh -c "$<TARGET_FILE:cohpsk_get_test_bits> - 5600 | $<TARGET_FILE:cohpsk_mod> - - | $<TARGET_FILE:ch> - - --No -30 --Fs 7500 | $<TARGET_FILE:cohpsk_demod> - - | $<TARGET_FILE:cohpsk_put_test_bits> -"
              )
 
     add_test(NAME test_COHPSK_modem_freq_offset
-             COMMAND sh -c "set -x; $<TARGET_FILE:cohpsk_get_test_bits> - 5600 | $<TARGET_FILE:cohpsk_mod> - - | $<TARGET_FILE:ch> - - -40 -f -30 --Fs 7500 | $<TARGET_FILE:cohpsk_demod> -v - - 2>log.txt | $<TARGET_FILE:cohpsk_put_test_bits> - ; ! grep 'lost sync' log.txt"
+             COMMAND sh -c "set -x; $<TARGET_FILE:cohpsk_get_test_bits> - 5600 | $<TARGET_FILE:cohpsk_mod> - - | $<TARGET_FILE:ch> - - --No -40 -f -30 --Fs 7500 | $<TARGET_FILE:cohpsk_demod> -v - - 2>log.txt | $<TARGET_FILE:cohpsk_put_test_bits> - ; ! grep 'lost sync' log.txt"
              )
 
     # -------------------------------------------------------------------------
@@ -523,7 +523,7 @@ if(UNITTEST)
                             ./fsk_put_test_bits - -q")
 
     add_test(NAME test_OFDM_modem_AWGN_BER
-             COMMAND sh -c "$<TARGET_FILE:ofdm_mod> --in /dev/zero --ldpc --testframes 60 --txbpf | $<TARGET_FILE:ch> - - -20  -f -50 | $<TARGET_FILE:ofdm_demod> --out /dev/null --testframes --ldpc --verbose 1"
+             COMMAND sh -c "$<TARGET_FILE:ofdm_mod> --in /dev/zero --ldpc --testframes 60 --txbpf | $<TARGET_FILE:ch> - - --No -20  -f -50 | $<TARGET_FILE:ofdm_demod> --out /dev/null --testframes --ldpc --verbose 1"
              )
 
     add_test(NAME test_OFDM_modem_fading_BER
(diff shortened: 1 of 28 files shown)


drowe67:
@srsampson @tmiw when convenient can you pls review this PR?

CMakeLists.txt
```diff
@@ -394,9 +394,9 @@ if(UNITTEST)
              set_tests_properties(test_CML_ldpcut PROPERTIES PASS_REGULAR_EXPRESSION "Nerr: 0")
 
     # check channel simulator measures correct Peak to Average Power Ratio (about 0dB) with a sine wave input signal
-    add_test(NAME test_cohpsk_ch_papr
+    add_test(NAME test_ch_papr
              COMMAND sh -c "cd ${CMAKE_CURRENT_BINARY_DIR};
-                            ./misc/mksine - 1000 10 | ./src/cohpsk_ch - /dev/null -100 --Fs 8000 --ctest")
+                            ./misc/mksine - 1000 10 | ./src/ch - /dev/null --ctest")
```
tmiw, line 399: Should this be `$<TARGET_FILE:ch>` as well?

tmiw: As well as the equivalent for mksine, if built as part of the build process.

CMakeLists.txt
```diff
@@ -552,7 +552,7 @@ endif()
     add_test(NAME test_OFDM_modem_700E_AWGN
              COMMAND sh -c "cd ${CMAKE_CURRENT_BINARY_DIR}/src;
                             ./ofdm_mod --in /dev/zero --testframes 10 --ldpc --mode 700E |
-                            ./cohpsk_ch - - -22 --Fs 8000 |
+                            ./ch - - --No -22 |
```
tmiw, line 555: `$<TARGET_FILE:ch>`, `$<TARGET_FILE:ofdm_mod>` and `$<TARGET_FILE:ofdm_demod>`.

tmiw: Similar comment for the rest of this file.

README_data.md
```diff
@@ -240,7 +240,7 @@ Some notes on this example:
...
 1. Our first tone `--tone` is at 10kHz, and each tone is spaced `--shift` by 10kHz, so we have FSK tones at 10,20,30, and 40 kHz.  For good performance, FSK tones must be spaced by at least the symbol rate Rs.
-1. Although the `cohpsk_ch` utility is designed for 8kHz sample rate operation, it just operates on sampled signals, so it's OK to use at higher sample rates.  It does have some internal filtering so best to keep your signal well away from 0 and (sample rate)/2.  The SNR measurement is calibrated to a 3000 Hz noise bandwidth, so won't make much sense at other sample rates.  The third argument `-12` sets the noise level of the channel.
+1. Although the `ch` utility is designed for 8kHz sample rate operation, it just operates on sampled signals, so it's OK to use at higher sample rates.  It does have some internal filtering so best to keep your signal well away from 0 and (sample rate)/2.  The SNR measurement is calibrated to a 3000 Hz noise bandwidth, so won't make much sense at other sample rates.  The third argument `-12` sets the noise level of the channel.
```
tmiw, line 243: If designed for 8KHz, wouldn't we want to keep our signals away from 4KHz regardless of the sample rate?

src/ch.c
```diff
@@ -273,25 +330,24 @@ int main(int argc, char *argv[])
...
             float mag = sqrt(ch_in[i].real*ch_in[i].real + ch_in[i].imag*ch_in[i].imag);
-            //fprintf(stdout, "%f\n",mag);
             float angle = atan2(ch_in[i].imag, ch_in[i].real);
             if (mag > clip) {
               mag = clip;
               nclipped++;
             }
             tx_pwr += mag*mag;
-            if (mag > peak) peak = mag;
+            if (mag > peak) { peak = mag; /*fprintf(stderr, "%f\n",mag);*/ }
```
tmiw, line 343: Does the commented code here need to be removed?

unittest/ofdm_phase_est_bw.sh
```diff
@@ -17,7 +17,7 @@ results=$(mktemp)
 fading_dir=$1
```
tmiw, line 17: Per https://github.com/drowe67/freedv-gui/pull/200, should we be using `/usr/bin/env bash` for all of the shell scripts in codec2 as well?

unittest/ofdm_phase_est_bw.sh
```diff
@@ -17,7 +17,7 @@ results=$(mktemp)
...
 # BER should be < 5% for this test
 ofdm_mod --in /dev/zero --testframes 300 --mode 2020 --ldpc -p 312 --verbose 0 | \
-cohpsk_ch - - -40 --Fs 8000 -f 10 --ssbfilt 1 --mpp --fading_dir $fading_dir | \
+ch - - --No -40 -f 10 --ssbfilt 1 --mpp --fading_dir $fading_dir | \
```
tmiw, line 20: These scripts were working before from what I remember, but I wonder if we need to be more specific with the paths to `ofdm_mod`, `ch`, etc. anyway. Thoughts?

CMakeLists.txt
drowe67: The `cd ${CMAKE_CURRENT_BINARY_DIR};` puts us in the correct directory so the following lines work.  I've adopted this style as the same code can be pasted to the cmd line and run manually while developing/debugging tests.  The `$<TARGET_FILE:ch>` style only works inside cmake (you can't past the line to the cmd line for manual development) and is less intuitive for people not very familiar with cmake.

This and the other tests seem to run OK.

CMakeLists.txt
drowe67: As per above the `cd ${CMAKE_CURRENT_BINARY_DIR}/src;` puts in the correct directory so the following code works.  It should be "build dir" agnostic - we've been writing/running tests like this for a while now, although some earlier tests still use the `$<TARGET_FILE:ch>` style.

tmiw:
Tests pass 100% on macOS here. 👍

CMakeLists.txt
tmiw: Makes sense. For some reason I thought CMake applied its own interpolation before sh.

CMakeLists.txt
tmiw: Yep, no worries. 👍

README_data.md
drowe67: Well the filters are all designed on a normalized scale, e.g. we might design one with a cut off of 2000Hz/8000Hz.  So the utility just works on this normalized scale - it can't tell the difference between a 2000 Hz signal at an 8000Hz sample rate and a 4000 Hz signal at a 16000 Hz sample rate.

It only gets scaled back to a real word sampling rate at the ADC and DAC :slightly_smiling_face:

src/ch.c
drowe67: I've actually been using it to debug just today over in https://github.com/drowe67/codec2/pull/274 !  We can remove it a little bit down the track.  I ended up putting a Tx BPF into 2020 to prevent some weirdness in this magnitude :thinking:

unittest/ofdm_phase_est_bw.sh
drowe67: One thing I like about using script files is once you set up the `$PATH` you can just use the application name without any path wrangling.

unittest/ofdm_phase_est_bw.sh
drowe67: Good idea: https://github.com/drowe67/codec2/pull/278/commits/01010b7fddfcb771615cece20b467a8afe61e137

I really hope `env` always live sin `/usr/bin` on all systems!

tmiw: Fortunately /usr/bin does seem to be a common place for it on the platforms we typically support. I think there are some oddball Unices that have it elsewhere but I doubt anyone's using FreeDV on those.

unittest/ofdm_phase_est_bw.sh
tmiw: I thought the changes to PATH weren't there but I checked again and they're there. 👍

PR #309, opened 2022-02-18
New function: freedv_set_fmin_fmax

DJ2LS:
My first freedv api function :partying_face: 

### New api function for **setting fmin and fmax** for OFDM modems.

**Use-case:** FreeDATA / TNC environment, where adjusting or changing the frequency is not always an option, specially for remote operation. The default +/- 50Hz are also a little bit too sensitive, if a radio is not that precise regarding to its frequency. Playing with XIT and RIT of the radio would also be possible, but an increased range of decoding at cost of not too much increased CPU - why not, as if its making things easier. 

**Usage:**
`int freedv_set_tuning_range(struct freedv *freedv, float val_fmin, float val_fmax)`
returns 1 , if is ofdm mode
returns 0 , if not ofdm mode

**Example Python + ctypes with fmin and fmax +/- 150Hz :**
```
freedv_set_tuning_range.restype = c_int
freedv_set_tuning_range.argype = [c_void_p, c_float, c_float]
freedv_set_tuning_range(freedv, c_float(-150.0), c_float(150.0))
```

### DIFF
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1372,6 +1372,25 @@ int freedv_get_bits_per_modem_frame       (struct freedv *f) {return f->bits_per
 int freedv_get_rx_status                  (struct freedv *f) {return f->rx_status;}
 void freedv_get_fsk_S_and_N               (struct freedv *f, float *S, float *N) { *S = f->fsk_S[0]; *N = f->fsk_N[0]; }
 
+
+/*---------------------------------------------------------------------------*\
+
+  FUNCTIONS...: freedv_set_fmin_fmax
+  AUTHOR......: Simon Lang - DJ2LS
+  DATE CREATED: 18 feb 2022
+
+  Set fmin / fmax for ofdm modem
+
+\*---------------------------------------------------------------------------*/
+void freedv_set_fmin_fmax(struct freedv *freedv, float val_fmin, float val_fmax) {
+    freedv->ofdm->fmin = val_fmin;
+    freedv->ofdm->fmax = val_fmax;
+}
+
+
+  
+
+
 int freedv_get_n_max_speech_samples(struct freedv *f) {
     /* When "passing through" demod samples to the speech output
        (e.g. no sync and squelch off) f->nin bounces around with
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1378,13 +1378,18 @@ void freedv_get_fsk_S_and_N               (struct freedv *f, float *S, float *N)
   FUNCTIONS...: freedv_set_fmin_fmax
   AUTHOR......: Simon Lang - DJ2LS
   DATE CREATED: 18 feb 2022
-
-  Set fmin / fmax for ofdm modem
-
+  DEFAULT.....: fmin: -50.0Hz fmax: 50.0Hz
+  DESCRIPTION.:
+  
+  |<---fmin - | rx centre frequency | + fmax--->|
+  
+  Useful for handling frequency offsets, 
+  e.g. caused by an imprecise VFO, the trade off is more CPU power is required.  
+  
 \*---------------------------------------------------------------------------*/
-int freedv_set_fmin_fmax(struct freedv *freedv, float val_fmin, float val_fmax) {
+int freedv_set_tuning_range(struct freedv *freedv, float val_fmin, float val_fmax) {
 
-    if (is_ofdm_data_mode(freedv)) {
+    if (is_ofdm_mode(freedv)) {
         freedv->ofdm->fmin = val_fmin;
         freedv->ofdm->fmax = val_fmax;
         return 1;
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -1389,7 +1389,7 @@ void freedv_get_fsk_S_and_N               (struct freedv *f, float *S, float *N)
 \*---------------------------------------------------------------------------*/
 int freedv_set_tuning_range(struct freedv *freedv, float val_fmin, float val_fmax) {
 
-    if (is_ofdm_data_mode(freedv) && (strcmp(freedv->ofdm->data_mode, "burst"))) {
+    if (is_ofdm_data_mode(freedv) && (strcmp(freedv->ofdm->data_mode, "burst") == 0)) {
         freedv->ofdm->fmin = val_fmin;
         freedv->ofdm->fmax = val_fmax;
         return 1;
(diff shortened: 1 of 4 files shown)


src/freedv_api.c
```diff
@@ -1372,6 +1372,28 @@ int freedv_get_bits_per_modem_frame       (struct freedv *f) {return f->bits_per
...
+  AUTHOR......: Simon Lang - DJ2LS
+  DATE CREATED: 18 feb 2022
+
+  Set fmin / fmax for ofdm modem
+
+\*---------------------------------------------------------------------------*/
+int freedv_set_fmin_fmax(struct freedv *freedv, float val_fmin, float val_fmax) {
+
+    if (is_ofdm_data_mode(freedv)) {
+        freedv->ofdm->fmin = val_fmin;
```
drowe67, line 1388: Should we also be checking if this is "burst mode"?  This API function just adjusts the acquisition range of the burst data modems (used for raw data mode).

src/freedv_api.c
```diff
@@ -1372,6 +1372,28 @@ int freedv_get_bits_per_modem_frame       (struct freedv *f) {return f->bits_per
...
 
+
+/*---------------------------------------------------------------------------*\
+
+  FUNCTIONS...: freedv_set_fmin_fmax
+  AUTHOR......: Simon Lang - DJ2LS
+  DATE CREATED: 18 feb 2022
+
+  Set fmin / fmax for ofdm modem
+
```
drowe67, line 1383: Could you pls add some comments here, similar to the top of the PR?  The source code is all that will be left after we merge.

src/freedv_api.c
```diff
@@ -1372,6 +1372,28 @@ int freedv_get_bits_per_modem_frame       (struct freedv *f) {return f->bits_per
 int freedv_get_rx_status                  (struct freedv *f) {return f->rx_status;}
 void freedv_get_fsk_S_and_N               (struct freedv *f, float *S, float *N) { *S = f->fsk_S[0]; *N = f->fsk_N[0]; }
 
+
+/*---------------------------------------------------------------------------*\
+
+  FUNCTIONS...: freedv_set_fmin_fmax
+  AUTHOR......: Simon Lang - DJ2LS
```
drowe67, line 1379: Do we actually need to set both fmin and fmax?  Maybe one argument is enough.  Then again we would need to make it clear that the total acquisition range is twice the single argument :thinking: 

Re naming how about:
`freedv_set_tuning_range(fmin, fmax)` or `freedv_set_freq_acquisition_range(fmin, fmax)`

drowe67:
@tmiw - I heard some new FreeDV users complain about "sensitive tuning" on air a few nights ago. So increasing the acquisition freq range using this function may be of interest for the "ease of use" meme.  I have the usual trepidation over increasing CPU usage in acquisition, but hey, this is experimental radio .... :slightly_smiling_face:

src/freedv_api.c
DJ2LS: My thoughts with using both - fmin AND fmax are, that it could be interesting, keeping this settings flexible. Maybe if two stations need to run near together, it could be interesting, if the tuning to the next station is "less sensitive / less wide" so they are not affecting each other. Maybe its like a "filter" or "out filtering" traffic of the neighbor station. But this are just theoretical thoughts for being flexible for possible future use-cases 🤔 

Yes, renaming the function to` freedv_set_tuning_range(fmin, fmax)` sounds good.

src/freedv_api.c
DJ2LS: Sure! I would suggest, I'll do this as soon we finished our discussion about the function names and final design

src/freedv_api.c
drowe67: Using `fmin` and `fmax` also makes the entire range (e.g. -100 to +100, 200Hz total) explicit, so I agree - lets keep both arguments :+1:

DJ2LS: I also did tests with unequal fmin-fmax values like  fmin = -150 and fmax = 50, which where successful. Or shouldn't they?

src/freedv_api.c
DJ2LS: so you are talking about `is_ofdm__mode(*f)` instead of `is_ofdm_data_mode(*f)` ?
This would add additional support for modes 2020, 700D, 700E - right?

tmiw: Burst mode is defined by `ofdm->data_mode` being set to "burst" (see line 1310 in ofdm.c).

tmiw:
> @tmiw - I heard some new FreeDV users complain about "sensitive tuning" on air a few nights ago. So increasing the acquisition freq range using this function may be of interest for the "ease of use" meme. I have the usual trepidation over increasing CPU usage in acquisition, but hey, this is experimental radio .... 🙂

I think it should be okay for most users with somewhat modern systems. Worst-case, we can add a checkbox in freedv-gui to disable this behavior.

src/freedv_api.c
tmiw: If we're allowing unequal values, then yes, having it work successfully is expected. 👍

src/freedv_api.c
```diff
@@ -1372,6 +1372,33 @@ int freedv_get_bits_per_modem_frame       (struct freedv *f) {return f->bits_per
 int freedv_get_rx_status                  (struct freedv *f) {return f->rx_status;}
 void freedv_get_fsk_S_and_N               (struct freedv *f, float *S, float *N) { *S = f->fsk_S[0]; *N = f->fsk_N[0]; }
 
+
+/*---------------------------------------------------------------------------*\
+
+  FUNCTIONS...: freedv_set_fmin_fmax
```
drowe67, line 1378: :point_up: function name in doc

DJ2LS: ooops, done with https://github.com/drowe67/codec2/pull/309/commits/a45310aff1402d0bd86da31c4fd2d3812156b283

src/freedv_api.c
DJ2LS: thanks for help, @tmiw !  I pushed a commit with this changes and I hope this is what you and @drowe67 where talking about.

tmiw: Yep, but the check should be if `strcmp(...) == 0` (it'll return non-zero if not equal).

DJ2LS: oh, sorry! I expected some returned "bool" value. Should have read C docs first!. However, I updated the corresponding line, recompiled and ran all ctests with success.

src/freedv_api.c
tmiw: Is this still going to be renamed to freedv_set_tuning_range?

DJ2LS: @tmiw this should already be solved by 
https://github.com/drowe67/codec2/commit/a45310aff1402d0bd86da31c4fd2d3812156b283

tmiw: Yep, for some reason I had an older version displayed.

Issue #312, opened 2022-03-02
Definition different from Declaration


maddin200:
src/gp_interleaver.h line 38 / 39 vs src/gp_interleaver.c line 90 / 100
void gp_interleave_float(float frame[], float interleaved_frame[], int Nbits);

void gp_interleave_float(float interleaved_frame[], float frame[], int Nbits) {

void gp_deinterleave_float(float interleaved_frame[], float frame[], int Nbits);

void gp_deinterleave_float(float frame[], float interleaved_frame[], int Nbits) {

drowe67:
Thanks @maddin200 - can you please post a Pull Request to fix this issue?

maddin200:
fixed

PR #313, opened 2022-03-03
fix declarations  #312

maddin200:


drowe67:
thanks @maddin200 :slightly_smiling_face:

PR #317, opened 2022-04-03
Allow use of __EMBEDDED__ on non-ARM systems.

tmiw:
Currently, enabling `__EMBEDDED__` results in failures on non-ARM systems due to the lack of CMSIS. This PR tweaks the code so that `__EMBEDDED__` can still be used on systems that need it without requiring CMSIS if not on ARM. Tested with ESP32 (compiling, not running).

Status:
1. 28/28 stm32 ctests passing :heavy_check_mark: 
2. Tested on SM1000 :heavy_check_mark: 
3. git tag `stm32_PR#317` to mark this PR as a point where stm32 tests are passing :heavy_check_mark:


tmiw:
@drowe67, here's an example of using the wrapper on my ESP32-S3 FreeDV project:

```
// Implement required Codec2 math methods below as CMSIS doesn't work on ESP32.

void codec2_dot_product_f32(float* left, float* right, size_t len, float* result)
{
    dsps_dotprod_f32(left, right, result, len);
}

void codec2_complex_dot_product_f32(float* left, float* right, size_t len, float* resultReal, float* resultImag)
{
    // Complex number math: (a + bi)(c + di) = ac + bci + adi + (bi)(di) = ac - bd + (bc + ad)i
    float realTimesRealResult; // ac
    float realTimesImag1Result; // bci
    float realTimesImag2Result; // adi
    float imagTimesImagResult; // bi * di
    
    dsps_dotprode_f32(left, right, &realTimesRealResult, len, 0, 0);
    dsps_dotprode_f32(left, right, &realTimesImag1Result, len, 1, 0);
    dsps_dotprode_f32(left, right, &realTimesImag2Result, len, 0, 1);
    dsps_dotprode_f32(left, right, &imagTimesImagResult, len, 1, 1);
    
    *resultReal = realTimesRealResult - imagTimesImagResult;
    *resultImag = realTimesImag1Result + realTimesImag2Result;
}
```

I'm open to suggestions for this, though.

drowe67:
@tmiw - great idea :100: This is a bit like the FFT library from @db4ple.  I also like the way it gets rid of the nested #ifdefs - makes the code much cleaner.

src/codec2_math.h
```diff
@@ -0,0 +1,50 @@
...
+//  along with this program; if not, see <http://www.gnu.org/licenses/>.
+//
+//==========================================================================
+
+#if defined(__ARM_ARCH)
+#include "arm_math.h"
+
+#define codec2_dot_product_f32(left, right, len, result) (arm_dot_prod_f32(left, right, len, result))
+#define codec2_complex_dot_product_f32(left, right, len, resultReal, resultImag) (arm_cmplx_dot_prod_f32(left, right, len, resultReal, resultImag))
+
```
drowe67, line 32: Suggest the ARM case goes inside codec2_math.c, making the header file content just the function headers.

drowe67: Oh I see, there is no codec2_math.c by default, that's user provided :thinking:

src/codec2_math.h
```diff
@@ -0,0 +1,50 @@
...
+#ifdef __cplusplus
+extern "C"
+{
+#endif // __cplusplus
+
+// Your embedded project must define these two methods if it is not running on ARM.
+extern void codec2_dot_product_f32(float* left, float* right, size_t len, float* result);
+extern void codec2_complex_dot_product_f32(float* left, float* right, size_t len, float* resultReal, float* resultImag);
+
```
drowe67, line 43: Can we please make the variable names in the function headers a little more informative?  It would be nice to make it clear that the first two arguments are complex number vectors in floatReal, floatImag format (I think?).  Is `len` the number length in floats or complex numbers?  It would be nice to understand those sorts of details from the header alone.

src/codec2_math.h
drowe67: Hmm, so should we add platform specific implementation (like your ESP32 example above) in a codec_math.h?  That way other people can build codec2 on those platforms out of the box.

tmiw: > Hmm, so should we add platform specific implementation (like your ESP32 example above) in a codec_math.h? That way other people can build codec2 on those platforms out of the box.

At least for ESP32, some of the implementation of ESP-DSP (especially complex math as the dot product functions are real-only) doesn't translate over to a single function call so we probably can't unless it can be represented as an inline function. I'm not sure we necessarily want to support every single embedded architecture out there, either, though just simple wrappers to the platform's existing DSP library probably won't be a huge deal.

src/codec2_math.h
tmiw: I updated the .h file to hopefully make the parameters more clear. Let me know if further clarifications are needed.

BTW I also added a `codec2_math_arm.c` so we don't need to use the preprocessor for the STM32 version of these.

tmiw:
Also, it looks like the SM1000 doesn't currently build but it's not because of this PR. I'll create a separate issue for that.

tmiw:
OK, with the changes in https://github.com/drowe67/codec2/pull/323 and this PR, the SM1000 builds successfully. 👍

src/codec2_math.h
drowe67: Thanks.  So the complex vectors have type float which is ambiguous.  Should we use the COMP struct or make it clear in the comments that they are floatReal,floatImag format?

tmiw: OK, updated comments accordingly.

drowe67: Taking another look at this one - I'm thinking we should use the COMP structure for the interface to make it consistent with other abstraction code like `codec2_fft.h` (and indeed the convention for the rest of libcodec2), and to make the interface clearer.  All that's required is to include `comp.h`.

I guess we also need to run all the stm32 ctests again.

tmiw: Haven't rerun the ctests yet (and probably won't have time to this weekend) but SM1000 does compile with the change I just pushed.

tmiw:
See https://github.com/drowe67/codec2/pull/317 for the fix for the build failures.

drowe67:
@tmiw - I tried running the stm32 ctests:

1. `master` is not passing some tests, for example:
```
ctest -V -R tst_ofdm_demod_quick
: BER checks
7: Reference
7: BER......: 0.0000 Tbits: 14518 Terrs:     0 Tpackets:    61 SNR3kdB: 38.80
7: BER2.....: 0.0000 Tbits:  9758 Terrs:     0
7: Target
7: BER......: 0.0000 Tbits: 14518 Terrs:     0
7: BER2.....: 0.0000 Tbits:  9758 Terrs:     0
7: PASS: max BER
7: PASS: BER compare
7: 
7: Output checks
7: FAIL Output, length mismatch
7: 
7: Test FAILED!
```
So this looks OK except for the output length mismatch, which might be an `ofdm_demod` (the reference tool) thing.

2. With this branch the stm32 side doesn't look right:
```
7: BER checks
7: Reference
7: BER......: 0.0000 Tbits: 14518 Terrs:     0 Tpackets:    61 SNR3kdB: 38.80
7: BER2.....: 0.0000 Tbits:  9758 Terrs:     0
7: Target
7: BER......:   nan Tbits:     0 Terrs:     0
7: BER2.....:   nan Tbits:     0 Terrs:     0
7
```

However it's been a while since I did any stm32 tests - so I may have messed up somewhere.  Anyhoo, no rush on this one. Next weekend is fine.

drowe67:
Have made some progress https://github.com/drowe67/codec2/pull/339/files

drowe67:
@tmiw - I had to make quite a few changes to get the stm32 ctests to work, including removing the `weak` attribute and moving `codec2_math_arm.c` to the `stm32/CMakeLists.txt`.  There may be other ways to do this - but please run the stm32 ctests if you make any changes.  

It would be useful to try this code on a SM1000 too.

src/ofdm.c
```diff
@@ -746,29 +746,33 @@ static int est_timing(struct OFDM *ofdm, complex float *rx, int length,
...
         codec2_dot_product_f32(&rx_real[i], wvec_pilot_real, ofdm->samplespersymbol, &re);
-        codec2_dot_product_f32(&rx_real[i], wvec_pilot_imag, ofdm->samplespersymbol, &im);
-        corr_st = re + im * I;
+	codec2_dot_product_f32(&rx_real[i], wvec_pilot_imag, ofdm->samplespersymbol, &im);
+	corr_st = re + im * I;
 
-        codec2_dot_product_f32(&rx_real[i+ ofdm->samplesperframe], wvec_pilot_real, ofdm->samplespersymbol, &re);
-        codec2_dot_product_f32(&rx_real[i+ ofdm->samplesperframe], wvec_pilot_imag, ofdm->samplespersymbol, &im);
-        corr_en = re + im * I;
-#elif defined(__EMBEDDED__)
```
tmiw: Wait, does the ARM version of gcc not like #elif for some reason or is this a style thing?

src/codec2_math.h
```diff
@@ -31,8 +31,7 @@ extern "C"
 {
 #endif // __cplusplus
 
-// Forward declare COMP struct (use comp.h to refer to items inside the struct).
-struct COMP;
+#include "comp.h"
```
tmiw, line 34: Wouldn't this already be included in theory by ofdm.c (even if only implicitly)? Though it probably wouldn't hurt to just include it anyway.

src/ofdm.c
drowe67: Not sure.  When it wasn't working I started with master/ofdm.c and made small changes, so ended up with the original #if ... #endif structure.

src/codec2_math.h
drowe67: I feel it's good practice to include data types used by a header file.

src/ofdm.c
tmiw: In theory both should be equivalent but if it's working now, it's not a big deal to keep as is.

src/codec2_math.h
tmiw: That works.

tmiw:
Just built and tested this PR on my SM1000, no issues receiving and transmitting all supported modes (1600, 700D, 700E).

drowe67:
Thanks @tmiw  you beat me to it :slightly_smiling_face: 

Do you think it would be useful to tag this point, something like "stm32 working" or "sm1000_v7.1?  Might help the next time the stm32 gets bit rot.

tmiw:
> Thanks @tmiw you beat me to it 🙂
> 
> Do you think it would be useful to tag this point? Something like:
> 
> 1. "stm32 working"
> 2. "sm1000_v7.1"
> 3. "stm32_ctests_pass"
> 4. "stm32_PR#317"
> 
> Might help the next time the stm32 gets bit rot. I'm not inclined to make an official sm1000 release, as it's all internal refactoring, no new features.

Sure (either 2 or 4 above). Longer term it becomes less important if we get STM32 ctests to run as part of PRs (even if only emulated) as we'll find breakages sooner.

PR #321, opened 2022-05-10
FreeDATA low SNR signalling mode

DJ2LS:
Hi,
after some testing with Octave, I designed 3 additional modes which might be useful for FreeDATA. My intention was, using the existing OFDM modem, but decreasing the number of carriers, which allows operation on really bad SNR channels. Before starting to work on a FSK mode which fits for HF channels, I think we should give the following ofdm modes a try for testing so we can use already existing resources. 
Focus is on a new signalling mode.

<hr>

- [x] 1. SNR requirements
- [x] 2. Tab for mode in doc/modem_codec_frame_design.ods
- [x] 3. Octave or C Pre/Post acquistion works at target operating points (e.g. test_OFDM_modem_octave_datac0_mpp)
- [ ] 4. Octave ctests at AWGN and MPP operating points.
- [ ] 5. Tune compression.
- [ ] 6. C Port to ofdm_mod/ofdm_demod level, ctest to show Octave and C working together, e.g. test_OFDM_modem_datac0_octave_burst
- [ ] 7. FreeDV API integration and ctest e.g. test_freedv_data_raw_ofdm_datac0_burst
- [x] 8. Scripted OTA test campaign between HF Tx and KiwiSDR e.g. unittest/ota_test.sh, demo mode works over real channels, gather some statistics.
- [ ] 9. Mode documented in README_data.md table.

## 1. Signalling mode requirements

- SNR: about -5dB MPP ( or as low as possible, with considering the signalling-mode specific tradeoff )
- shorter than 1.5seconds frame duration
- Minimum Payload: 10Byte
- Reduced carriers with the idea of "punching" through QRM signals or operation at low SNR, because of higher PAPR than existing datac0
- LDPC ratio min 0.5

### 1.1 Prototype candidates

| Mode  | payload data bit/s  | payload  | FEC | Np  | Nc  | Ns  | UW  | duration |
|---|---|---|---|---|---|---|---|---|
| datac2_1  | 106.06  |  112 | (112/224) | 6 | 3 | 8 | 28 | 1.056s |
| datac2_2  | 127.27  |  112  | (112/224) | 8 | 4 | 5 | 32 | 0.88s |
| datac2_3  | 103.9  |  112  | (112/224) | 7 | 3 | 7 | 28 | 1.078 |

### 1.2 Run Prototype datac2
`ofdm_tx("test_datac2.raw","datac2",1,-5,"awgn","bursts",1);ofdm_rx("test_datac2.raw", "datacX0", "packetsperburst", 1)`

`ofdm_ldpc_tx("test_datac2.raw","datac2",1,-5,"awgn","bursts",1);ofdm_ldpc_rx("test_datac2.raw", "datac2", "packetsperburst", 1)`

`ofdm_tx("test_datac2.raw","datac2",1,-5,"mpp","bursts",1);ofdm_rx("test_datac2.raw", "datac2", "packetsperburst", 1)`

`ofdm_ldpc_tx("test_datac2.raw","datac2",1,-5,"mpp","bursts",1);ofdm_ldpc_rx("test_datac2.raw", "datac2", "packetsperburst", 1)`

### 1.3 Prototype candidates test results
| Mode  | PER AWGN (SNR3k) | PER MPP -5dB (SNR3k)| PER MPD -5dB (SNR3k) | 
|---|---|---|---|
| datac2_1   | 95/100 @-5.8dB (-6.02) | 70/100 ( -5.74) | 58/100 (-6.62) | 
| datac2_2  | 91/100 @-5.6dB (-5.53) | 54/100 (-4.46) | 48/100 (-5.01) | 
| datac2_3  | 93/100 @-5.8dB (-6.09) | 73/100 (-5.24) | 56/100 (-5.89) |

### DIFF
--- a/src/fdmdv.c
+++ b/src/fdmdv.c
@@ -1021,7 +1021,7 @@ void rxdec_filter(COMP rx_fdm_filter[], COMP rx_fdm[], COMP rxdec_lpf_mem[], int
 
 \*---------------------------------------------------------------------------*/
 
-static void fir_filter2(float acc[2], float mem[], const float coeff[], const unsigned int dec_rate) {
+static void fir_filter2(float acc[], float mem[], const float coeff[], const unsigned int dec_rate) {
     acc[0] = 0.0;
     acc[1] = 0.0;
 


drowe67:
@DJ2LS - so I tried:

```
octave:2> ofdm_tx("test_datacX0.raw","datacX0",1,-6.6,"awgn","bursts",1);
error: vec(1): out of bound 0 (note: variable 'vec' shadows function)
error: called from
    ofdm_init_mode at line 381 column 7
    ofdm_tx at line 48 column 10
```

drowe67:
> > ofdm_tx("test_datacX0.raw","datacX0",1,-6.6,"awgn","bursts",1);
> 
> Mh, interesting. I'm running the tests with Octave 6.3 and/or 7.1 thinking I will setup a test run on a Ubuntu virtual machine.

The previous data modes work OK, it might be a namespace thing.

DJ2LS:
@drowe67 , for some reason my gitclient added all of the cmake files 🤔 - sorry!

drowe67:
@DJ2LS re a plan for development of a new mode - can we please start with updating the requirements table above?  It might be better to start with a single mode.

Here is a draft task list:

1. SNR requirements for mode at top of PR above.
1. Tab for mode in `doc/modem_codec_frame_design.ods`
1. Octave or C Pre/Post acquistion works at target operating points (e.g. `test_OFDM_modem_octave_datac0_mpp)`
1. Octave ctests at AWGN and MPP operating points.
1. Tune compression.
1. C Port to ofdm_mod/ofdm_demod level, ctest to show Octave and C working together, e.g. `test_OFDM_modem_datac0_octave_burst`
1. FreeDV API integration and ctest e.g.  `test_freedv_data_raw_ofdm_datac0_burst`
1. Scripted OTA test campaign between HF Tx and KiwiSDR e.g. `unittest/ota_test.sh`, demo mode work sover real channels, gather some statistics.
1. Mode documented in README_data.md table.

DJ2LS:
Thanks, @drowe67 , this makes the test procedure more clear and structured and helps introducing new modes. I will update my initial comment, soon. 
I also agree with starting with one mode as my focus is also on datacX0. The other two modes are a proof of concept while testing to see how the modes behave under different parameter constellation .

drowe67:
Hi @DJ2LS, thanks for the update, a couple of questions/comments:
1.  Re Section 1.1, the table at the top, I tend to bump the rev number when I try different versions, e.g. datac0_1, datac0_2. In the table above both waveforms are called datac2.
1. Is the "data rate" the payload data rate, or uncoded (raw) data rate?
1. Ns, the pilot insertion rate, sets the fading channel performance.  Ns=8 will struggle with fast fading channels like MPP, as it won't be able to track the fast phase changes.  I'd recommend Ns=5.  The pilots let us sample the phase of the channel, so Ns is the "sample period" for phase.
1. What SNR where the tests conducted at for AWGN/MPP? Best to present a test/table row for each case, as the SNR operating  points will be different.
1. What AWGN SNR does theory (spreadsheet) estimate for 10% PER?  This is a good check.

DJ2LS:
> Hi @DJ2LS, thanks for the update, a couple of questions/comments:
> 
> 1. Re Section 1.1, the table at the top, I tend to bump the rev number when I try different versions, e.g. datac0_1, datac0_2. In the table above both waveforms are called datac2.

:heavy_check_mark:

> 2. Is the "data rate" the payload data rate, or uncoded (raw) data rate?

payload data rate --> I updated the tables above
> 3. Ns, the pilot insertion rate, sets the fading channel performance.  Ns=8 will struggle with fast fading channels like MPP, as it won't be able to track the fast phase changes.  I'd recommend Ns=5.  The pilots let us sample the phase of the channel, so Ns is the "sample period" for phase.

okay, I'll set up an additional test candidate with lower Ns and lower Nc. However, we had overall a usable performance, I think on MPP 🤔 
> 4. What SNR where the tests conducted at for AWGN/MPP? Best to present a test/table row for each case, as the SNR operating  points will be different.

I let the test run at -5dB AWGN & MPP, are you talking about the returned SNR value after the check has completed?
> 5. What AWGN SNR does theory (spreadsheet) estimate for 10% PER?  This is a good check.

Good question, which calculation do you specially mean? I can only find `MPP operating point SNR3k (dB)` and `AWGN operating post SNR3k (db)`

drowe67:
> payload data rate --> I updated the tables above

Cool, so about 1/3 the data rate of DATAC3.  That's a good increment I think :+1: 
 
> okay, I'll set up an additional test candidate with lower Ns and lower Nc. However, we had overall a usable performance, I think on MPP thinking

MPP is about 1 Hz, maybe try MPD for a tougher channel.  From FreeDV users in the UK, I understand the Winter channel in Northern Europe often has fading faster that 1 Hz.  FreeDV 700D uses Ns=8, and struggles in Winter there, but 700E works well.
 
> > 4. What SNR where the tests conducted at for AWGN/MPP? Best to present a test/table row for each case, as the SNR operating  points will be different.
> 
> I let the test run at -5dB AWGN & MPP, are you talking about the returned SNR value after the check has completed?

Suggest you add an additional test at the SNR that results in 10% PER for AWGN (it might be -9dB).  The current 100/100 result for AWGN at -5dB doesn't tell us very much.  It's more interesting to see where it breaks :slightly_smiling_face: 
 
> > 5. What AWGN SNR does theory (spreadsheet) estimate for 10% PER?  This is a good check.
> 
> Good question, which calculation do you specially mean? I can only find `MPP operating point SNR3k (dB)` and 
`AWGN operating post SNR3k (db)`

Yes the `AWGN operating post SNR3k (db)` calculation on the spreadsheet, e.g. cell C40 on the datac0 tab.  This works from the Eb/No where the code breaks.  Actually I found a small bug in these calculations when playing with 2020B recently, but it should still be within a dB of the AWGN simulation 10% PER SNR.

drowe67:
> Will add this to the tests! I'm wondering, how Ns=8 behaves with short frames. So when using long data frames, I would expect more fading related interference than compared to a short frame like we have it in this scenario?

Longer frames work better as we spread errors over time and have a better chance of correcting them.  Longer LDPC codewords also work better.  Ideally, we want the frame duration to be > 2x the fade period.
 
> I'm not sure, but it seems there are some missing formulas?` Multipath uncoded operating point Eb/No (dB)` for example thinking

Yes the formula is missing. I don't have a good theoretical model for fading channels.

drowe67:
`
> Raw BER..: 0.1407 Tbits: 17640 Terrs:  2482 SNR3k: -7.08
> Coded BER: 0.0907 Tbits:  7840 Terrs:   711
> Coded PER: 0.6000 Pckts:    70 Perrs:    42 Npre: 49 Npost: 29

A PER of 60% is very high, I normally shoot for 10%.

> As you already told, Ns > 5 will result in a poor performance on fading channels. Just to be sure, I didn't forget anything: Fading will be simulated by the Octave tests? 

We can simulate fading (multpath channels MPP, MPD etc) in Octave or C.  The next step after that is OTA, but best to do as a formal, controlled test.

> Can the data modes compared with voice modes even if they use the same modulation technique, so we can benefit from the experiences made with them? Say we have a continuous 700D stream, which struggles at UK winter channels and compare it with a short, 1-1.5 seconds datac with equal Ns value thinking

At the modem frame level yes. At the packet level - the data modes have longer packets so have a better PER at a given SNR compared to the short voice packets (180ms and less).  However voice is more tolerant to errors than data.  We decode voice packets and present them to the listener even if they have bit errors.

DJ2LS:
Oh it seems I focused too much on "Pckts" .... so if we look at it from a PER point of view,  I get a PER of 0.15 on -3dB MPP (0.19 on -4db MPP) for datac2_2, which outperforms datac2_1 and datac2_3 in this case. 
I'm a little bit confused -  sorry. But its a learning curve and we are getting closer to a usable candidate which is worth testing OTA, I think.

DJ2LS:
```
Raw BER..: 0.0786 Tbits: 17408 Terrs:  1369 SNR3k: -4.19
Coded BER: 0.0319 Tbits:  7616 Terrs:   243
Coded PER: 0.1912 Pckts:    68 Perrs:    13 Npre: 49 Npost: 30
```

This is the result with datac2_2, running at -4dB MPP  Octave test. PER increases with increasing dB. But a PER of 20% at this low SNRs may be okay, as we have to expect higher PER. Theres most interesting, if we still could operate or not. So if we have a higher PER, but still able communicating with some attempts, it may be worth a try.

drowe67:
Yeah the slope of the PER v SNR curve is very flat for HF multipath channels.  I used to joke that "you can add as much power as you want, the BER will always be 10%".

I don't think it's a given that we should expect higher PER at lower SNR.  Everything should scale with bit rate.

I note in codec2/README_data.md the data points I presented were for 30% PER at 0dB - so you might still be 5dB in front of the earlier modes :slightly_smiling_face: 

It would be nice to test the protocol layer with different PERs, e.g. measure throughput versus PER.  This could be done with a simulated modem (just drop packets in software).  It would help tune to protocol code too.

drowe67:
> Coded PER: 0.1912 Pckts:    68 Perrs:    13 Npre: 49 Npost: 30

Hmm, I think this says we acquired 49+30=79 packets, out of them 68 packets were decoded (some might have a UW fail), and 13 of the 68 had errors.  So we successfully received 55/100 packets  of the original 100 transmitted.  However it's been a year since  I played with these simulations, I'll have to check the code again to confirm.

drowe67:
OK here's a sample run of datac0 at MPP 0dB:
```
ofdm_ldpc_tx('test_datac0.raw','datac0',1,0,'mpp','bursts',100);
ofdm_ldpc_rx('test_datac0.raw','datac0','packetsperburst',1)
<snip>
Coded PER: 0.2826 Pckts:    92 Perrs:    26 Npre: 82 Npost: 12
```
So we acquired 82+12=94, out of that we attempted to decode 92, but had 26 errors, so a PER of 28%.  Out of the total of 100 packets sent, (92-26)=66 were received OK, so a throughput of 66%.

I guess there are several ways to lose a packet;
1. We might fail to acquire it (we can't see the pre or post amble)
2. UW doesn't match, this can be due to bit errors, or acquisition making a mistake about the packet location.
3. FEC can't correct bit errors

datac3 is a little better (longer packets), (87-13)=74/100 packets made it through MPP 0dB:
```
Coded PER: 0.1494 Pckts:    87 Perrs:    13 Npre: 73 Npost: 20
```

As a long term idea (not for now), a protocol that doesn't depend on short signalling packets would have performance gains.

DJ2LS:
> Yeah the slope of the PER v SNR curve is very flat for HF multipath channels. I used to joke that "you can add as much power as you want, the BER will always be 10%".
> 
> I don't think it's a given that we should expect higher PER at lower SNR. Everything should scale with bit rate.
> 

I wrote this from a point of view regarding to a signalling mode. But yes, you're right 👍 

> I note in codec2/README_data.md the data points I presented were for 30% PER at 0dB - so you might still be 5dB in front of the earlier modes 🙂
> 
> It would be nice to test the protocol layer with different PERs, e.g. measure throughput versus PER. This could be done with a simulated modem (just drop packets in software). It would help tune to protocol code too.

I could run a test here locally - just compiling the different modes and testing them under more or less real circumstances

DJ2LS:
> OK here's a sample run of datac0 at MPP 0dB:
> 
> ```
> ofdm_ldpc_tx('test_datac0.raw','datac0',1,0,'mpp','bursts',100);
> ofdm_ldpc_rx('test_datac0.raw','datac0','packetsperburst',1)
> <snip>
> Coded PER: 0.2826 Pckts:    92 Perrs:    26 Npre: 82 Npost: 12
> ```
> 
> So we acquired 82+12=94, out of that we attempted to decode 92, but had 26 errors, so a PER of 28%. Out of the total of 100 packets sent, (92-26)=66 were received OK, so a throughput of 66%.
> 
> I guess there are several ways to lose a packet;
> 
> 1. We might fail to acquire it (we can't see the pre or post amble)
> 2. UW doesn't match, this can be due to bit errors, or acquisition making a mistake about the packet location.
> 3. FEC can't correct bit errors
> 
> datac3 is a little better (longer packets), (87-13)=74/100 packets made it through MPP 0dB:
> 
> ```
> Coded PER: 0.1494 Pckts:    87 Perrs:    13 Npre: 73 Npost: 20
> ```
> 
> As a long term idea (not for now), a protocol that doesn't depend on short signalling packets would have performance gains.

This might be possible somehow and somewhen, I think. Actually I'm trying to find a tradeoff signalling mode which fits for both - low and high SNR operation.

However, I need some more time here with doing theoretical/octave experiments.

DJ2LS:
## OTA proof of concept test run

### Idea 
Testing DATAC0 against DATAC2 with different parameter constellation to figure out the benefits of a reduced Nc ( Number of carriers ).

### Test
**1. DATAC0 vs DATAC2_2** 
DATAC2_2 is most comparable with DATAC0, but with decreased Nc. We are going to determine the benefits of reduced Nc

**2. DATAC2_2 vs DATAC2_1**
DATAC2_1 has a higher Ns (pilot tone interval) and lower Nc. We want to figure out - as long as DATAC2_2 has benefits over DATAC0 - if a higher Ns has negative effects on a data mode

**3. DATAC2_1 vs DATAC2_3**
If DATAC2_1 outperforms  DATAC2_2 we might consider a test run for "fine tuning" for checking another parameter constellation.

DJ2LS:
Latest, most promising test results of datac2:
  ```
    Ns=5; config.Np=8; Tcp = 0.006; Ts = 0.016; Nc = 4; config.data_mode = "burst"; config.Nuwbits = 32; 
    config.Ntxtbits = 0; config.bad_uw_errors = 9;
    config.state_machine = "data";
    config.ftwindow_width = 80; config.amp_est_mode = 1; config.EsNodB = 10;
    config.edge_pilots = 0; config.timing_mx_thresh = 0.08;
    config.tx_uw = zeros(1,config.Nuwbits);
    config.tx_uw(1:24) = [1 1 0 0  1 0 1 0  1 1 1 1  0 0 0 0  1 1 1 1  0 0 0 0];
    config.tx_uw(end-24+1:end) = [1 1 0 0  1 0 1 0  1 1 1 1  0 0 0 0  1 1 1 1  0 0 0 0];
    config.amp_scale = 300E3; config.clip_gain1 = 2.2; config.clip_gain2 = 0.8;
```

Issue #322, opened 2022-05-16
SM1000 build failure on Ubuntu 21.10 


tmiw:
While working on #317 I tried to compile the SM1000 firmware on Ubuntu 21.10. Unfortunately I got the following:

```
[ 68%] Building C object CMakeFiles/codec2_prof.dir/codebooknewamp1_energy.c.obj
[ 68%] Building C object CMakeFiles/codec2_prof.dir/codebooknewamp1.c.obj
[ 68%] Linking C static library libcodec2_prof.a
[ 69%] Built target codec2_prof
Scanning dependencies of target sm1000base
[ 69%] Building C object CMakeFiles/sm1000base.dir/src/sm1000_leds_switches.c.obj
[ 70%] Building C object CMakeFiles/sm1000base.dir/src/debugblinky.c.obj
[ 70%] Building C object CMakeFiles/sm1000base.dir/src/system_stm32f4xx.c.obj
[ 70%] Building ASM object CMakeFiles/sm1000base.dir/src/startup_stm32f4xx.s.obj
[ 70%] Linking C static library libsm1000base.a
[ 70%] Built target sm1000base
[ 71%] Building C object CMakeFiles/dac_ut.dir/src/dac_ut.c.obj
[ 71%] Linking C executable dac_ut.elf
Creating binary for dac_ut
/usr/bin/objcopy: Unable to recognise the format of the input file `dac_ut.elf'
make[2]: *** [CMakeFiles/dac_ut.dir/build.make:102: dac_ut.elf] Error 1
make[2]: *** Deleting file 'dac_ut.elf'
make[1]: *** [CMakeFiles/Makefile2:1084: CMakeFiles/dac_ut.dir/all] Error 2
make: *** [Makefile:101: all] Error 2
```

Anyway, it seems that it's using the x86_64 version of objcopy instead of `arm-none-eabi-objcopy` as expected.

drowe67:
Hi @tmiw - can we close this one now?

tmiw:
Yep!

PR #323, opened 2022-05-16
Add override for objcopy to the toolchain file.

tmiw:
1. Add building for the stm32 to the set of GitHub actions we run.
1. Per #322, CMake is improperly choosing the host version of `objdump` instead of the ARM version. This PR adds an override to the toolchain file to force CMake to choose the correct version.

### DIFF
--- a/.github/workflows/cmake-sm1000.yml
+++ b/.github/workflows/cmake-sm1000.yml
@@ -23,11 +23,6 @@ jobs:
            sudo apt-get update
            sudo apt-get install octave octave-common octave-signal liboctave-dev gnuplot sox p7zip-full python3-numpy valgrind 
 
-    - name: Run ctests
-      working-directory: ${{github.workspace}}/build_linux
-      shell: bash
-      run: ctest --output-on-failure
-
     - name: Install ST Standard Peripheral Library (SM1000)
       working-directory: ${{github.workspace}}/stm32
       shell: bash
--- a/.github/workflows/cmake-sm1000.yml
+++ b/.github/workflows/cmake-sm1000.yml
@@ -1,0 +1,48 @@
+name: Build SM1000
+
+on: [pull_request]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Debug
+
+jobs:
+  build:
+    # The CMake configure and build commands are platform agnostic and should work equally
+    # well on Windows or Mac.  You can convert this to a matrix build if you need
+    # cross-platform coverage.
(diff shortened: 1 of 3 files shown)


tmiw:
> Thanks @tmiw - so how did this problem creep in, I am wondering why we didn't see it before?

Honestly, I don't think I ever tried building the SM1000 code since before I upgraded my VM to 21.10. Perhaps now is a good time to add a GitHub action for it so this doesn't happen in the future?

drowe67:
> OH, hmm. I wonder if it was something I forgot to do that caused the compile error. From the README:

Indeed, however that was written a few Ubuntu versions ago, so the situation may have improved.

> BTW I also just noticed that the STM32 libraries need registration on TI's website to get. That might have been why we never had a GitHub action for the SM1000.

Ahh, yes, that's right.

tmiw:
> > OH, hmm. I wonder if it was something I forgot to do that caused the compile error. From the README:
> 
> Indeed, however that was written a few Ubuntu versions ago, so the situation may have improved.
> 
> > BTW I also just noticed that the STM32 libraries need registration on TI's website to get. That might have been why we never had a GitHub action for the SM1000.
> 
> Ahh, yes, that's right.

Would [this version](https://github.com/whimsicalraps/STM32F4xx_DSP_StdPeriph_Lib) potentially work for CI builds? I'm not sure what's different between that and 1.8.0.

.github/workflows/cmake-sm1000.yml
```diff
@@ -0,0 +1,43 @@
...
+      working-directory: ${{github.workspace}}/stm32
+      shell: bash
+      run: git clone https://github.com/whimsicalraps/STM32F4xx_DSP_StdPeriph_Lib
+
+    - name: Install SM1000 prerequisites
+      working-directory: ${{github.workspace}}/stm32
+      shell: bash
+      run: sudo apt install gcc-arm-none-eabi
+
+    - name: Build SM1000
```
drowe67, line 36: Do we need to build x86 codec2 first?  I always thought that was part of the sm1000 build procedure, although perhaps that was just to support running the sm1000 ctests.

tmiw: IIRC it's just for ctests.

drowe67:
@tmiw - is this one good to merge?  I think building for the stm32 on each push is a pretty good idea, will catch any obvious issues.

tmiw:
> @tmiw - is this one good to merge? I think building for the stm32 on each push is a pretty good idea, will catch any obvious issues.

Yep, we can merge. 👍

PR #325, opened 2022-05-23
Handle ftwindow being odd better

srsampson:
I noticed that ftwindow was cast to a float on one line, but not on  another.

This is just to align the logic and also cast back to an integer after floor() and ceil().

P.S. It is normally even, and equal to 32

### DIFF
--- a/src/ofdm.c
+++ b/src/ofdm.c
@@ -1391,7 +1391,7 @@ static void ofdm_demod_core(struct OFDM *ofdm, int *rx_bits) {
     if (ofdm->timing_en == true) {
         /* update timing at start of every frame */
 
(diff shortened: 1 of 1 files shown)


drowe67:
Thanks @srsampson.  Just curious - is the `(int)` in front of the ceilf()/floorf() necessary?  Perhaps it generates a warning?

tmiw:
I'm going to assume no issues with ctests passing before and after PR, right?

srsampson:
Yes, I started with a fresh clone, built and ctest, then changes and built and ctest again. Then of course it built and ctested in the PR.

The (int) isn't really required in C, as it will cast the float correctly. I could have left it off. Other languages do give an error (such as Java). Which led me to this change.

After the floor() and ceil() it is essentially an integer without a fractional part.

PR #326, opened 2022-06-01
Add new optimized TX BPF implementation for OFDM modes.

tmiw:
This PR contains a potential optimized TX BPF implementation that basically ignores the complex component and only operates on real components. As a result, there is a significant improvement in performance on x86_64:

*Original implementation*:

```
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 1 700D hrlongsinewave.raw tx.raw

real	0m17.933s
user	0m17.444s
sys	0m0.095s
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 1 700D hrlongsinewave.raw tx.raw

real	0m17.501s
user	0m17.326s
sys	0m0.092s
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 1 700D hrlongsinewave.raw tx.raw

real	0m18.805s
user	0m18.241s
sys	0m0.138s
HT-TM05:build_osx mooneer$ 
```

*This PR*:

```
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 1 700D hrlongsinewave.raw tx.raw

real	0m13.571s
user	0m13.470s
sys	0m0.091s
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 1 700D hrlongsinewave.raw tx.raw

real	0m13.621s
user	0m13.491s
sys	0m0.092s
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 1 700D hrlongsinewave.raw tx.raw

real	0m13.513s
user	0m13.415s
sys	0m0.090s
HT-TM05:build_osx mooneer$
```

SNR per `freedv_rx` also appears similar to the current implementation:

```
HT-TM05:build_osx mooneer$ src/freedv_rx -v 700D tx.raw rx.raw
frame: 1  demod sync: 0  nin: 1280 demod snr: -5.00 dB  bit errors: 0 clock_offset: 0.000000
frame: 2  demod sync: 0  nin: 370 demod snr: -5.00 dB  bit errors: 0 clock_offset: 0.000000
frame: 3  demod sync: 1  nin: 1280 demod snr: 6.76 dB  bit errors: 0 clock_offset: 0.000000
frame: 4  demod sync: 1  nin: 1280 demod snr: 11.43 dB  bit errors: 0 clock_offset: 0.000000
frame: 5  demod sync: 1  nin: 1280 demod snr: 11.10 dB  bit errors: 0 clock_offset: 0.000000
frame: 6  demod sync: 1  nin: 1280 demod snr: 10.86 dB  bit errors: 0 clock_offset: 0.000000
frame: 7  demod sync: 1  nin: 1280 demod snr: 10.53 dB  bit errors: 0 clock_offset: 0.000000
frame: 8  demod sync: 1  nin: 1280 demod snr: 10.36 dB  bit errors: 0 clock_offset: 0.000000
frame: 9  demod sync: 1  nin: 1280 demod snr: 10.09 dB  bit errors: 0 clock_offset: 0.000000
frame: 10  demod sync: 1  nin: 1280 demod snr: 9.97 dB  bit errors: 0 clock_offset: 0.000000
```


tmiw:
@drowe67, are the assumptions made here reasonable? I'm also not sure if the 100 pole filter the PR is currently using for 700D is sufficient. From the code:

```
// Alternate version of above BPF generated using the following:
//
//     # taps = 100
//     0- 750 Hz: -60dB attenuation
//  1000-2000 Hz: 0.1dB ripple
//  2250-4000 Hz: -60dB attenuation
//
// This filter provides less steep skirts, which allows us to use fewer poles to
// provide bandpass filtering. As a result, there is a significant performance 
// improvement over the 100 pole Quisk complex filter above.
```

I had also tried the following BPF but there was no performance improvement (and might have had slightly worse runtime, even):

```
// BPF generated from http://t-filter.engineerjs.com/ using the following:
//
//     # taps = 256
//     0- 900 Hz: -60dB attenuation
//  1000-2000 Hz: 0.1dB ripple
//  2100-4000 Hz: -60dB attenuation
//
// Note that the coefficients are reversed. This is because after we add the current x[t] to the
// lookback buffer, we start calculating from x[t - N] and loop back around, making b[0] * x[t] the
// the last multiplication done. Doing it this way enables easier vectorization by the compiler 
// (or via e.g. CMSIS).
//
// Additionally, this filter is used as-is by fir_quick_*. That is, the coefficients are not made
// complex before applying them. This works as the TX BPF for OFDM modes is the last operation executed,
// after which the complex component is thrown out.
```

If the less steep skirts are okay, I can migrate 700E, etc. over too. 👍

tmiw:
One of the memory leak ctests is failing, will need to investigate tonight. Other than that, all other ctests pass. 👍

src/filter.c
```diff
@@ -279,3 +280,60 @@ void quisk_ccfFilter(complex float * inSamples, complex float * outSamples, int
...
+void fir_quick_destroy(struct fir_quick** filter)
+{
+    assert(filter != NULL);
+    assert(*filter != NULL);
+    
+    FREE((*filter)->lookbackBuffer);
+    FREE(*filter);
+    *filter = NULL;
+}
+
```
drowe67, line 308: Do we really need new init functions?  Could the original ones be use?  Perhaps this saves some memory.

src/filter_coef.h
```diff
@@ -614,3 +614,389 @@ float quiskFilt120t480[480] = {
...
 
+// BPF generated from http://t-filter.engineerjs.com/ using the following:
+//
+//     # taps = 256
+//     0- 900 Hz: -60dB attenuation
+//  1000-2000 Hz: 0.1dB ripple
+//  2100-4000 Hz: -60dB attenuation
+//
+// Note that the coefficients are reversed. This is because after we add the current x[t] to the
+// lookback buffer, we start calculating from x[t - N] and loop back around, making b[0] * x[t] the
```
drowe67, line 625: I think the coeffs are symmetrical, so it might run just fine without reversal.  As per my comments I would prefer not to change coeffs at this time - it may affect the compressor/PAPR.

drowe67:
Hi @tmiw - thanks for taking a look at optimising the BPF.  It is quite intensive in terms of CPU.

Some thoughts/questions:
1. On what platforms do we actually have a problem with CPU for this filter?  As per the Flex work - I'm uncomfortable about adding complexity to the ofdm code just for one particular platform.
1. I'm not that keen on alternate code paths - two pieces of code doing the same thing.  It means more to maintain and test.  I'm flexible on this if there is a good reason for it.  
1. Could this be done with a simple(r) change to `quisk_ccfFilter()` to just output the real part of the signal?
1. Does this break `freedv_comptx()` for 700D?
1. The shape factor of the filter is really up to us, I don't have any firm specs.  I guess stop band  >50dB as IIRC that's a typical Ham spurious spec.  Actually that 50dB is perhaps the product of the Tx sidebands (which might already be attenuated) and filter response. Perhaps before changing the filter coeffs we should establish some requirements.  It's all been a bit ad-hoc until now :thinking: 
1. Re suggested filter coeff changes - what is more important, CPU or shape factor?  Shape factor effectively sets the bandwidth of the on-air signal.  Perhaps we should set some requirements here.
1. However the filter coeffs also affect PAPR and hence performance so good to avoid changing the filter coeffs unless there is a really good reason, and we have checked the PAPR.  We could perhaps consider that step in another PR if there is a good reason for it.
1. Do you have a feel/measurement for the time spent in the filter code with the optimisation?  for example compare `--txbpf 0` to `--txbpf 1` with the optimisations in this PR.  So we can measure the CPU used by the filter alone.
1. What does the tx spectrum look like compared to the original, complex implementation?  Good to check it's still filtering as expected.

An idea I had along these lines (to really minimise the CPU) is use an IIR filter.  This could be very short (5th order).  They have a bit of phase distortion around the transitions but that's probably OK, as we have per-carrier phase estimation.  That might result in a universal solution (one code path, trivial storage for coeffs) that is very efficient.  Lot of work though ... many filters to develop/test and the idea might not work, e.g. IIR filter are short but don't vectorise well.

tmiw:
> Hi @tmiw - thanks for taking a look at optimising the BPF. It is quite intensive in terms of CPU.
> 
> Some thoughts/questions:
> 
> 1. On what platforms do we actually have a problem with CPU for this filter?  As per the Flex work - I'm uncomfortable about adding complexity to the ofdm code just for one particular platform.

The most immediate improvement would be for the embedded platforms (ESP32, SM1000). The ESP32S3, in particular, needs something like 20 extra milliseconds to run the current BPF, resulting in ~120ms total time generating 160ms of audio. In other words, it can still handle FreeDV in real time but a bit extra buffer would be nice.

In any case, the changes are intended to be used for all platforms, not just the embedded ones. This should help with maintainability.

> 2. I'm not that keen on alternate code paths - two pieces of code doing the same thing.  It means more to maintain and test.  I'm flexible on this if there is a good reason for it.

It looks like outside ofdm.c, the existing Quisk filter code is being used by freedv_700.c to go to/from 7500 and 8000 Hz sample rates. Keeping the old code around for 700C would be good to prevent the possibility of breaking that mode.

Other than that, once we figure out what to do for 700D, much of the work in this PR can easily be applied to 700E, etc. (and I intend to do so).

> 3. Could this be done with a simple(r) change to `quisk_ccfFilter()` to just output the real part of the signal?

The existing filters seem to be LPFs that are converted to complex BPFs by `quisk_cfTune()`. I'm not sure if it's sufficient to use only the real components of the complex BPFs generated by it.

> 4. Does this break `freedv_comptx()` for 700D?

Looks like `freedv_comptx()` gets used unless you're transmitting 2400A/B or 800XA. As I'm able to decode audio generated by `./freedv_tx` with `./freedv_rx` (and most ctests pass), I'm confident this didn't break it.

> 5. The shape factor of the filter is really up to us, I don't have any firm specs.  I guess stop band  >50dB as IIRC that's a typical Ham spurious spec.  Actually that 50dB is perhaps the product of the Tx sidebands (which might already be attenuated) and filter response. Perhaps before changing the filter coeffs we should establish some requirements.  It's all been a bit ad-hoc until now 🤔
> 6. Re suggested filter coeff changes - what is more important, CPU or shape factor?  Shape factor effectively sets the bandwidth of the on-air signal.  Perhaps we should set some requirements here.

I know for 700D we have been saying that it's 1 KHz wide. Perhaps we should make a best effort to steepen the skirts unless it causes a lot of trouble with CPU usage?

> 7. However the filter coeffs also affect PAPR and hence performance so good to avoid changing the filter coeffs unless there is a really good reason, and we have checked the PAPR.  We could perhaps consider that step in another PR if there is a good reason for it.

Is there something I can run to check this?

> 8. Do you have a feel/measurement for the time spent in the filter code with the optimisation?  for example compare `--txbpf 0` to `--txbpf 1` with the optimisations in this PR.  So we can measure the CPU used by the filter alone.

Using this PR:

```
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 1 700D hrlongsinewave.raw tx.raw

real	0m13.778s
user	0m13.578s
sys	0m0.097s
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 0 700D hrlongsinewave.raw tx.raw

real	0m11.543s
user	0m11.333s
sys	0m0.097s
HT-TM05:build_osx mooneer$
```

and with master:

```
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 1 700D hrlongsinewave.raw tx.raw

real	0m18.776s
user	0m17.653s
sys	0m0.104s
HT-TM05:build_osx mooneer$ time src/freedv_tx --clip 1 --txbpf 0 700D hrlongsinewave.raw tx.raw

real	0m12.572s
user	0m11.810s
sys	0m0.123s
HT-TM05:build_osx mooneer$
```

> 9. What does the tx spectrum look like compared to the original, complex implementation?  Good to check it's still filtering as expected.

700D spectrum according to Audacity (using master):

Screen Shot 2022-06-01 at 8 04 42 PM

and using this PR:

Screen Shot 2022-06-01 at 8 03 52 PM

The version generated with this PR seems a tad bit narrower with no change in peak.

> An idea I had along these lines (to really minimise the CPU) is use an IIR filter. This could be very short (5th order). They have a bit of phase distortion around the transitions but that's probably OK, as we have per-carrier phase estimation. That might result in a universal solution (one code path, trivial storage for coeffs) that is very efficient. Lot of work though ... many filters to develop/test and the idea might not work, e.g. IIR filter are short but don't vectorise well.

Considering the spectrum above, we might be able to adjust the new FIR further to more closely match master's results. That will likely reduce the number of taps (though I'm not sure by how much, so it's possible IIR could still improve performance even further).

src/filter.c
tmiw: I'm not fully sure we can remove the original ones at the moment. See below.

drowe67:
Hi @tmiw - lets discuss the goals and design:

1. So I think we both agree that improving the CPU load of the tx BPF filters is desirable. However the current CPU load isn't a practical problem on any platform (e.g. no end user would notice any difference).
1. From my side, I'd like to minimise any code changes, and new code paths.  I'm not inclined to make big changes just to lower CPU.  Happy to make modest changes, and accept a modest improvement in CPU.
1. I'm uncomfortable with changes to filter coefficients as this affects the PAPR tuning which does have an impact on SNR performance.  While we could work through this if there was a good reason too - I can't see a good reason given (1).  I'm also time poor at present, and this would require careful review.
1. We can't break `freedv_comptx()` - so I think the current Quisk implementation needs to stay.  Note that although `freedv_comptx()` is called by `freedv_tx()`, this is not a test that the complex valued version is working, as `freedv_tx` only works with real valued signals.  This is an oversight - there should be a ctest for complex sample operation of these FreeDV API functions.
1. The key thing you've discovered is that in the real valued `freedv_tx()` use case we can save a lot of CPU.  That's pretty cool.

So here is a suggested design:

1. Modify `quisk_ccfFilter()` to just perform calculations to output the real part.  No changes to the the filter init/tuning functions (minimal code changes).  As you suggest above this might need some investigation.  My first read of the function suggests real and imag samples are just parallel operations.
1. Enable this "real mode" when called from `freedv_tx()` somehow. 
1. Or alternatively code a `quisk_crfFilter()` function that just outputs real samples.
1.  So this is a new code path we need to test.  It should output the same samples as master - we are just not calculating the imag samples that are discarded anyway.
1. Add ctests to check `freedv_comptx()` and `freedv_comprx()` are working correctly with single sided, complex valued signals.  Need to think abut an appropriate test for this :thinking: Maybe a small exp(jw) frequency shift? Might need a new `codec2/unittest` application, rather than adding a complex code to `freedv_tx`.

@tmiw - would this design work for you :point_up:? I'm looking to come up with a design that meets both of our needs.  Please feel free to comment/suggest alternatives/spot any issues in my reasoning.

tmiw:
> 4. We can't break `freedv_comptx()` - so I think the current Quisk implementation needs to stay.  Note that although `freedv_comptx()` is called by `freedv_tx()`, this is not a test that the complex valued version is working, as `freedv_tx` only works with real valued signals.  This is an oversight - there should be a ctest for complex sample operation of these FreeDV API functions.

It looks like freedv-gui uses `freedv_comptx()` directly. However, the complex components of the input samples are likely still 0. There is a call to `freq_shift_coh()` immediately after which I'm not sure about though. Do you know of anything else from third parties that uses `freedv_comptx()`?

> 5. The key thing you've discovered is that in the real valued `freedv_tx()` use case we can save a lot of CPU.  That's pretty cool.

Yep 👍 

> So here is a suggested design:
> 
> 1. Modify `quisk_ccfFilter()` to just perform calculations to output the real part.  No changes to the the filter init/tuning functions (minimal code changes).  As you suggest above this might need some investigation.  My first read of the function suggests real and imag samples are just parallel operations.

Considering that `(a + bi)(c + di) = ac + bci + adi - bd`, we'd likely still need to calculate `ac - bd` but could possibly not calculate `ad + bc`. Still a 50% reduction in calculations and would still be a significant decrease in CPU load, unless I'm missing something?

> 2. Enable this "real mode" when called from `freedv_tx()` somehow.

New FreeDV API call the application uses on start? Perhaps something called `freedv_enable_real_filter(dvObj, 1)` to enable this mode? Doing so would also cause an assertion in `freedv_comptx()` to ensure that code doesn't accidentally provide complex samples.

> 3. Or alternatively code a `quisk_crfFilter()` function that just outputs real samples.
> 4. So this is a new code path we need to test.  It should output the same samples as master - we are just not calculating the imag samples that are discarded anyway.

It'd probably be neater/easier to understand if there was a separate function, IMO. Since there are already ctests for `freedv_tx()`, we'd likely have coverage of it, leaving below to implement.

> 5. Add ctests to check `freedv_comptx()` and `freedv_comprx()` are working correctly with single sided, complex valued signals.  Need to think abut an appropriate test for this 🤔 Maybe a small exp(jw) frequency shift? Might need a new `codec2/unittest` application, rather than adding a complex code to `freedv_tx`.

Yeah, I'm not sure on this either. Maybe we could just take what `freedv_comptx()` returns for a simple sine wave and pass it back into `freedv_comprx()`?

> @tmiw - would this design work for you ☝️? I'm looking to come up with a design that meets both of our needs. Please feel free to comment/suggest alternatives/spot any issues in my reasoning.

Will need to try it out and see if it works but it seems doable.

drowe67:
> It looks like freedv-gui uses `freedv_comptx()` directly. However, the complex components of the input samples are likely still 0. There is a call to `freq_shift_coh()` immediately after which I'm not sure about though. Do you know of anything else from third parties that uses `freedv_comptx()`?
>
Not that I know of, but this being open source it's hard to say.  The obvious use case is for SDR Tx.

The complex version is used in the raw data section of the API e.g. `freedv_data_raw_tx.c`, and in `freedv_rx.c` (rx side).

IIRC the use of  `freedv_comptx()` in freedv-gui followed by `freq_shift_coh()` is for "click to tune" split freq operation.  It's easy to add small freq offsets with complex signals.  This was added very early on, I'm not sure if it's used by anyone.

> Considering that `(a + bi)(c + di) = ac + bci + adi - bd`, we'd likely still need to calculate `ac - bd` but could possibly not calculate `ad + bc`. Still a 50% reduction in calculations and would still be a significant decrease in CPU load, unless I'm missing something?

Good catch - this function is doing complex float operations.  Interesting to ponder the difference between this and a purely real bandpass filter as you have prototyped.  My gut feel is there is a "no free lunch"  trade off there - e.g. roughly the same number of cycles using either approach for a given freq response.  But this is just a guess :thinking: 

> > 2. Enable this "real mode" when called from `freedv_tx()` somehow.
> 
> New FreeDV API call the application uses on start? Perhaps something called `freedv_enable_real_filter(dvObj, 1)` to enable this mode? Doing so would also cause an assertion in `freedv_comptx()` to ensure that code doesn't accidentally provide complex samples.

Can we do it without adding a FreeDV API call?  Given we are already calling freedv_tx() we know that only real samples are required.  This also means we can call freedv_tx() or freedv_comptx() at any time, no mode specific "gotchas".


> > 5. Add ctests to check `freedv_comptx()` and `freedv_comprx()` are working correctly with single sided, complex valued signals.  Need to think abut an appropriate test for this thinking Maybe a small exp(jw) frequency shift? Might need a new `codec2/unittest` application, rather than adding a complex code to `freedv_tx`.
> 
> Yeah, I'm not sure on this either. Maybe we could just take what `freedv_comptx()` returns for a simple sine wave and pass it back into `freedv_comprx()`?

I'm thinking `comptx -> 10 Hz shift -> comprx`, as small freq shifts only work with complex samples. With real samples you get two copies at +/- 10Hz which would break the modem. I'm happy to code that up if it helps.

> > @tmiw - would this design work for you point_up? I'm looking to come up with a design that meets both of our needs. Please feel free to comment/suggest alternatives/spot any issues in my reasoning.
> 
> Will need to try it out and see if it works but it seems doable.

:+1:

tmiw:
So I tried the following:

```
void quisk_ccfFilter_realonly(complex float * inSamples, complex float * outSamples, int count, struct quisk_cfFilter * filter) {
    for (int i = 0; i < count; i++) 
    {    
        float* coeffPtr = (float*)filter->cpxCoefs;
        int numValsInTaps = filter->nTaps * 2;
        float filterOutAC = 0;
        float filterOutBD = 0;
        
        *filter->ptcSamp = inSamples[i];
        
        complex float * ptSample = filter->ptcSamp;
        for (int k = 0; k < numValsInTaps; k += 2) 
        {
            // Note: The below only calculates the real component of the result,
            // not the complex one.
            float* ptSampleReal = (float*)ptSample;
            filterOutAC += ptSampleReal[0] * coeffPtr[k];
            filterOutBD -= ptSampleReal[1] * coeffPtr[k + 1];
            
            if (--ptSample < filter->cSamples)
                ptSample = filter->cSamples + filter->nTaps - 1;
        }

        outSamples[i] = filterOutAC + filterOutBD;
        
        if (++filter->ptcSamp >= filter->cSamples + filter->nTaps)
            filter->ptcSamp = filter->cSamples;
    }
}
```

with it being enabled via a new `ofdm_set_realonlybpf()` function (only called by `freedv_tx()` just before and after the `freedv_comptx()` call). The result in Audacity seems to match master:

Screen Shot 2022-06-06 at 6 02 20 PM

However, there wasn't nearly as much time savings as before:

```
Mooneers-16-MacBook-Pro-2325:build_osx mooneer$ time src/freedv_tx --txbpf 1 --clip 1 700D hrlongsinewave.raw tx.raw

real	0m17.399s
user	0m16.811s
sys	0m0.198s
Mooneers-16-MacBook-Pro-2325:build_osx mooneer$ time src/freedv_tx --txbpf 1 --clip 1 700D hrlongsinewave.raw tx.raw

real	0m17.684s
user	0m16.623s
sys	0m0.184s
Mooneers-16-MacBook-Pro-2325:build_osx mooneer$ time src/freedv_tx --txbpf 1 --clip 1 700D hrlongsinewave.raw tx.raw

real	0m16.977s
user	0m16.544s
sys	0m0.167s
Mooneers-16-MacBook-Pro-2325:build_osx mooneer$ 
```

(avg. 17.35s)

I suspect this is because the compiler is unable to use certain optimizations (such as SIMD) due to having a reverse running component (`filter->ptcSamp`). Perhaps we can change that (assuming we're okay with changing `quisk_ccfFilter()` as well) by having the input samples saved bottom to top instead of top to bottom as per now?

tmiw:
Looks like the below (as well as the appropriate adjustments in the rest of filter.c to make sure coefficients/samples are inserted/run through in the correct order when going between the complex and real-only versions) does better time-wise:

```
void quisk_ccfFilter_realonly(complex float * inSamples, complex float * outSamples, int count, struct quisk_cfFilter * filter) {
    for (int i = 0; i < count; i++) 
    {    
        float* coeffPtr = (float*)filter->cpxCoefs;
        int numValsInTaps = filter->nTaps * 2;
        float filterOutAC = 0;
        float filterOutBD = 0;
        
        *filter->ptcSamp = inSamples[i];
        
        float* ptSample = (float*)filter->ptcSamp;
        float* endPtSample = (float*)(filter->cSamples + filter->nTaps);
        for (int k = 0; k < numValsInTaps; k += 2) 
        {
            // Note: The below only calculates the real component of the result,
            // not the complex one. This works because:
            // (a + bi)(c + di) = ac + bci + adi + bdi^2 = ac - bd + (ad + bc)i.
            filterOutAC += *ptSample++ * *coeffPtr++;
            filterOutBD -= *ptSample++ * *coeffPtr++;
            
            if (ptSample == endPtSample)
                ptSample = (float*)filter->cSamples;
        }

        outSamples[i] = filterOutAC + filterOutBD;
        
        if (--filter->ptcSamp < filter->cSamples)
            filter->ptcSamp = filter->cSamples + filter->nTaps - 1;
    }
}
```

Results:

```
Mooneers-16-MacBook-Pro-2325:build_osx mooneer$ time src/freedv_tx --txbpf 1 --clip 1 700D hrlongsinewave.raw tx.raw

real	0m16.313s
user	0m15.665s
sys	0m0.116s
Mooneers-16-MacBook-Pro-2325:build_osx mooneer$ time src/freedv_tx --txbpf 1 --clip 1 700D hrlongsinewave.raw tx.raw

real	0m15.531s
user	0m15.340s
sys	0m0.095s
Mooneers-16-MacBook-Pro-2325:build_osx mooneer$ time src/freedv_tx --txbpf 1 --clip 1 700D hrlongsinewave.raw tx.raw

real	0m15.587s
user	0m15.409s
sys	0m0.101s
Mooneers-16-MacBook-Pro-2325:build_osx mooneer$ 
```

(avg. 15.81s)

I haven't checked this version in yet so whatever's preferred I suppose. 👍

drowe67:
Hi @tmiw - thanks for trying those experiments.  Putting some of the results above in a table:

| Test | User time | 
|--- | --- |
| master --txbpf 1 | 17.6 |
| master --txbpf 0 | 11.8 |
| 7 June --txbpf 1 | 15.4 |

Curious that the 7 June result is not halving the CPU load, it's perhaps 1/3 reduction of the entire filtering CPU load. I guess that's useful though. Perhaps the CPU is not dominated by the multiply accumulates.

Interesting point about the https://github.com/drowe67/codec2/pull/326/commits/ece77b1fcc4ff4d2c60212e90fc7c72c9732e01f prototype enabling various automatic optimisations like SIMD to kick in.  However these are on a platform (e.g. x86 Desktops) where we don't need any acceleration (a delta of a few seconds out of an hour).  These platforms also tend to have more options available for optimsation (like powerful SIMD). So while convenient, I'm wondering if this is the best platform to evaluate this work :thinking: I guess ideally we'd see some gains across a sample of platforms.

Would it be worth measuring the CPU using the June 7 prototype on one of the "small" embedded targets?

tmiw:
> Curious that the 7 June result is not halving the CPU load, it's perhaps 1/3 reduction of the entire filtering CPU load. I guess that's useful though. Perhaps the CPU is not dominated by the multiply accumulates.

I think `master` was closer to 18.5s average on the hour-long test file but IIRC there was a lot of variance. 3s improvement is still pretty good. 👍 

> Would it be worth measuring the CPU using the June 7 prototype on one of the "small" embedded targets?

[Apparently the ESP32S3 has vector instructions](https://esp32.com/viewtopic.php?t=22144) so I went ahead and ran some tests on ezDV with 700D:

`ms-embedded-noarm` + clip/BPF off:

```
I (9305) FreeDVTask: freedv_tx ran in 99794 us on 1280 samples and generated 1280 samples
I (9469) FreeDVTask: freedv_tx ran in 100733 us on 1280 samples and generated 1280 samples
I (9632) FreeDVTask: freedv_tx ran in 100240 us on 1280 samples and generated 1280 samples
I (9796) FreeDVTask: freedv_tx ran in 100298 us on 1280 samples and generated 1280 samples
I (9960) FreeDVTask: freedv_tx ran in 99740 us on 1280 samples and generated 1280 samples
```

`ms-embedded-noarm` + clip/BPF on:

```
I (6940) FreeDVTask: freedv_tx ran in 144465 us on 1280 samples and generated 1280 samples
I (7105) FreeDVTask: freedv_tx ran in 144146 us on 1280 samples and generated 1280 samples
I (7271) FreeDVTask: freedv_tx ran in 144159 us on 1280 samples and generated 1280 samples
I (7437) FreeDVTask: freedv_tx ran in 143565 us on 1280 samples and generated 1280 samples
I (7604) FreeDVTask: freedv_tx ran in 145596 us on 1280 samples and generated 1280 samples
```

`ms-embedded-noarm` + clip/BPF on + making complex version of `quisk_ccfFilter` run forward-only:

```
I (5545) FreeDVTask: freedv_tx ran in 143754 us on 1280 samples and generated 1280 samples
I (5710) FreeDVTask: freedv_tx ran in 144133 us on 1280 samples and generated 1280 samples
I (5876) FreeDVTask: freedv_tx ran in 144016 us on 1280 samples and generated 1280 samples
I (6042) FreeDVTask: freedv_tx ran in 143889 us on 1280 samples and generated 1280 samples
I (6208) FreeDVTask: freedv_tx ran in 144180 us on 1280 samples and generated 1280 samples
```

`ms-embedded-noarm` + clip/BPF on + using `quisk_ccfFilter_realonly()`:

```
I (8392) FreeDVTask: freedv_tx ran in 116747 us on 1280 samples and generated 1280 samples
I (8550) FreeDVTask: freedv_tx ran in 115777 us on 1280 samples and generated 1280 samples
I (8708) FreeDVTask: freedv_tx ran in 116114 us on 1280 samples and generated 1280 samples
I (8867) FreeDVTask: freedv_tx ran in 115906 us on 1280 samples and generated 1280 samples
I (9027) FreeDVTask: freedv_tx ran in 116638 us on 1280 samples and generated 1280 samples
```

In short:

* No difference between what's effectively `master`'s implementation of the BPF and the complex-only optimization that I attempted (but still not checked in).
* ~30ms per `freedv_tx()` call improvement between `master` and the current PR.

It's possible there could be further improvements by also using the new codec2 math stuff introduced in #317 as part of `quisk_ccfFilter_realonly()` but I'm not sure by how much. [Espressif makes it sound like it'd be significant](https://docs.espressif.com/projects/esp-dsp/en/latest/esp-dsp-benchmarks.html) depending on the functions used, though.

tmiw:
> using what I have denoted the "June 7" version above? 

Actually, I just pushed the version that I used for those results. Figured that'd cut down on the confusion. :)

> I feel we need to see meaningful improvement on a small machine to consider merging any filter optimisations. For faster machines alone the case is not strong (e.g. 3 seconds in an hour = 3/3600 = 0.08% improvement).

Yep. Fortunately there does seem to be a significant difference on the embedded side, just need to button up how that should look.

src/filter.h
```diff
@@ -36,6 +36,7 @@ extern void quisk_filt_cfInit(struct quisk_cfFilter *, float *, int);
 extern void quisk_filt_destroy(struct quisk_cfFilter *);
 extern void quisk_cfTune(struct quisk_cfFilter *, float);
 extern void quisk_ccfFilter(complex float *, complex float *, int, struct quisk_cfFilter *);
+extern void quisk_ccfFilter_realonly(complex float *, complex float *, int, struct quisk_cfFilter *);
```
drowe67, line 40: How about using the Quisk nomenclature, I think `cc` means complex in complex out, so the function could be named: `quisk_crfFilter`.

drowe67:
Thanks @tmiw - much clearer :slightly_smiling_face: 

So is this table correct for the ESP32? I'd like to make sure we clearly record the outcome of the PR here.

| Version | --txbpf | cycles |
| --- | --- | --- |
| master | 0 | 100E3 |
| master | 1 |144E3 |
| https://github.com/drowe67/codec2/pull/326/commits/bd74df7e1b20fde1ef2e0941e480d24ab1ab2ea6 | 1 | 115E3 |

I was thinking about the way we select the real output filter but can't figure out a better way than your `ofdm_set_realonlybpf`.  An alternative would be a function argument in `freedv_comptx()`, but there are too many layers to drill through.  It's opaque and stateless to the caller which was the main aim :+1:

src/filter.h
tmiw: Works for me. I'll rename it.

tmiw:
> So is this table correct for the ESP32? 

Yep. 👍 

> It's opaque and stateless to the caller which was the main aim 👍

Speaking of which, I pushed some additional changes so users can use both real and complex TX in a single session. Otherwise, https://github.com/drowe67/codec2/commit/bd74df7e1b20fde1ef2e0941e480d24ab1ab2ea6 would have locked the user into one or the other (lest they get incorrect results).

drowe67:
Hi @tmiw re https://github.com/drowe67/codec2/pull/326/commits/80350dc7364407e712471af9503da2c69edf6571 the complex filter has been changed and there is no (c)test to determine the output it is exactly the same as the previous version. It is very easy to introduce subtle errors with these sorts of changes.  So we need some way to verify this.  For example I'd suggest sample by sample comparison with master running in complex mode, e.g. `load_raw()` output files, plot difference in Octave.  Perhaps you have performed such a test already?

Can I pls confirm - is it still necessary to reverse the storage order?  I saw your notes above but just wanted to double check as there have been many versions explored in this PR. 

We also need a way to determine that `quisk_crfFilter` produces the same samples as master.  Subtle glitches in the samples may not show up in the current ctests or a spectrum plot from Audacity.

One approach for a ctest:
1. Keep `quisk_ccfFilter` unmodified to use as a reference (and accept we can't use complex/real in one session for the moment).
1. Make a run of ofdm_mod or freedv_tx with complex output samples A.
1. Make a run of ofdm_mod or freedv_tx with real output samples B.
1. Use some Octave to compare the real part of A to B. Ensure they are the same, e.g. a MSE of 1 sample or something small.

drowe67:
BTW I think I can see why these filters are fairly slow.  If I'm reading the code correctly they have an `if` adjusting the pointer to the previous input samples part way through the Ntap MAC loop.  I tend to favour block based approaches where the previous and new samples are arranged like:

`|Ntap memory of previous input samples|new input samples|`

This allows the Ntap dot product to be performed without interruption.  You just have one block based memory update right at the end.  This approach needs some more storage I guess.

But this is just FYI - no need to worry about this sort of change for this PR.

src/freedv_api.c
```diff
@@ -340,9 +340,20 @@ void freedv_tx(struct freedv *f, short mod_out[], short speech_in[]) {
         }
         freedv_tx_fsk_voice(f, mod_out);
     } else {
+        // Temporarily enable real-only BPF to improve performance.
+        if (FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_700E, f->mode)) {
+            ofdm_set_realonlybpf(f->ofdm, 1);
```
drowe67, line 345: Any particular reason why we only run this for 700D & E?

tmiw: Those were the only OFDM modes that came to mind offhand, actually. We can add additional ones if there's a need but given that those are the only two we can run on the embedded platforms at the moment, it probably wouldn't hurt to keep as is?

tmiw:
> Can I pls confirm - is it still necessary to reverse the storage order? I saw your notes above but just wanted to double check as there have been many versions explored in this PR.

I would say yes, especially for platforms that can do SIMD (such as some ESP32 chips). It's harder for the compiler to determine whether SIMD can be used otherwise.

> We also need a way to determine that `quisk_crfFilter` produces the same samples as master. Subtle glitches in the samples may not show up in the current ctests or a spectrum plot from Audacity.
> 
> One approach for a ctest:
> 
> 1. Keep `quisk_ccfFilter` unmodified to use as a reference (and accept we can't use complex/real in one session for the moment).
> 2. Make a run of ofdm_mod or freedv_tx with complex output samples A.
> 3. Make a run of ofdm_mod or freedv_tx with real output samples B.
> 4. Use some Octave to compare the real part of A to B. Ensure they are the same, e.g. a MSE of 1 sample or something small.

Unfortunately I'm not an expert on Octave so I'd have to look into how such a test could be done in that language. However, if we revert [80350dc](https://github.com/drowe67/codec2/pull/326/commits/80350dc7364407e712471af9503da2c69edf6571) we can perhaps just compare the generated .raw files using C code instead. Is that a reasonable approach?

drowe67:
Hi @tmiw - no worries, let me think about a way to test it.  I'm traveling atm (try Goggling Coober Pedy!), but I'll take a look at it over the next few days.

tmiw:
> Hi @tmiw - no worries, let me think about a way to test it. I'm traveling atm (try Goggling Coober Pedy!), but I'll take a look at it over the next few days.

Huh, that sounds like an interesting place to visit. Not sure I'd want to live there, though, if they have to build houses underground to avoid the constant heat. Have a good trip!

tmiw:
Interesting, not sure why there's that much of a difference:

```
    Start 59: test_freedv_api_700D_real_comp

59: Test command: /bin/sh "-c" "cd /Users/mooneer/devel/codec2/unittest;
                            PATH=$PATH:/Users/mooneer/devel/codec2/build_osx/demo:/Users/mooneer/devel/codec2/build_osx/unittest;
                            ./check_real_comp.sh"
59: Test timeout computed to be: 1500
59: + cat ../raw/ve9qrp_10s.raw
59: + freedv_700d_tx
59: + cat ../raw/ve9qrp_10s.raw
59: + freedv_700d_comptx
59: + echo 'tx_real=load_raw('\''tx_700d.int16'\''); tx_comp=load_raw('\''tx_700d.iq16'\'');       tx_comp=tx_comp(1:2:end)+j*tx_comp(2:2:end);       diff = sum(real(tx_comp)-tx_real); printf('\''diff: %f\n'\'', diff);       if diff < 1, quit(0), end;       quit(1)'
59: + octave-cli -p ../octave -qf
59: diff: 1093.000000
1/1 Test #59: test_freedv_api_700D_real_comp ...***Failed    1.01 sec
```

tmiw:
@drowe67, I've been investigating why the new ctest in #331 was failing with this PR. It turns out that `ofdm_clip()` is still being executed immediately after the BPF (even with clipping disabled), which relies on complex values. Fortunately, generating real-only coefficients and input samples via `cabsf()` resolved that issue. Additionally, there's a bit more of a time improvement on x86_64 as a result:

```
HT-TM05:build_osx mooneer$ time src/freedv_tx 700D hrlongsinewave.raw hrlongsinewave-tx.raw

real	0m16.618s
user	0m16.002s
sys	0m0.233s
HT-TM05:build_osx mooneer$ time src/freedv_tx 700D hrlongsinewave.raw hrlongsinewave-tx.raw

real	0m16.156s
user	0m15.506s
sys	0m0.190s
HT-TM05:build_osx mooneer$ time src/freedv_tx 700D hrlongsinewave.raw hrlongsinewave-tx.raw

real	0m15.658s
user	0m15.361s
sys	0m0.185s
HT-TM05:build_osx mooneer$
```

(avg. 16.14s)

I haven't tried running on SM1000 yet to make sure the extra heap space for real coefficients won't be a problem. On ezDV at least it shouldn't be a problem, especially if you have one of the ESP32 parts with additional RAM built-in.

drowe67:
Ahh, actually that might be a show stopper for this entire PR.  We reply on Hilbert clipping (clipping of the complex magnitude) to implement PAPR reduction.  This might imply the use of a complex BPF filter due to that final clipping stage (which is intentionally always on).

Integrity of the DSP is more important than CPU reduction.

Suggest you stop development on this PR until this is resolved - I'll take a detailed look at this when I can.

tmiw:
> Ahh, actually that might be a show stopper for this entire PR. We reply on Hilbert clipping (clipping of the complex magnitude) to implement PAPR reduction. This might imply the use of a complex BPF filter due to that final clipping stage (which is intentionally always on).
> 
> Integrity of the DSP is more important than CPU reduction.
> 
> Suggest you stop development on this PR until this is resolved - I'll take a detailed look at this when I can.

Will do. I think the previous change made things worse, anyway, so I'll at least revert that. If it helps, the difference reported by the new ctest was only something like ~1100 (with the vast majority of samples being the same for both real and complex), which may or may not still be okay.

drowe67:
> Will do. I think the previous change made things worse, anyway, so I'll at least revert that. If it helps, the difference reported by the new ctest was only something like ~1100 (with the vast majority of samples being the same for both real and complex), which may or may not still be okay.

Yes that's probably the only samples the final clipper is affecting.  The BPF tends to "ring" a bit, so some output samples exceed the input level.  The final clipper prevents anything > 16384 getting out and tickling the SSB radio ALC (which would reduce the RMS Tx power).  Even occasional peaks will reset the ALC.

Clipping the complex magnitude reduces distortion, which will also gives us a cleaner Tx waveform.  So we don't want to adopt real only clipping. 

So I'm a bit stuck on this one :thinking: Even though it's not affecting any platforms atm im principle I too would like to reduce the BPF CPU - but that PAPR/compression work has given us a a very big system gain (around 4dB) so I'm not inclined to compromise that aspect of performance to improve CPU.  So I can't see a way forward on this PR without significant work - and I'd prefer to put that effort into other aspects of codec2 right now (like improved speech quality which I'm trying to get back into).

So @tmiw how would you feel if we close this one for now?  We've learnt a lot, and perhaps a better solution will appear in the future.

tmiw:
> So @tmiw how would you feel if we close this one for now? We've learnt a lot, and perhaps a better solution will appear in the future.

Maybe add to the list of things we could hire out for with ARDC funds?

Anyway, I might revisit the idea of offloading the BPF to the TLV320 chip for ezDV, at least only for 700E as BPF is mandatory for that mode. I'm currently offloading the HPF to it for the ADC channels and it's doing very well, anyway (basically the background hiss that existed due to AGC has completely gone away now).

drowe67:
> > So @tmiw how would you feel if we close this one for now? We've learnt a lot, and perhaps a better solution will appear in the future.
> 
> Maybe add to the list of things we could hire out for with ARDC funds?

I was thinking along similar lines.  As a first step it would be good to list all our "nice to haves", then triage them at PLC level, based on some metric like potential impact.  One aspect I'd like to manage better is what PRs I say yes and no to reviewing. PRs like this can require much more of my time to review compared to simpler, non DSP ones. That takes me away from other work that might be more core.  So it would be better to decide where to work at PLC level - before the coding/PR starts.

PR #328, opened 2022-06-08
ofdm->m should be rounded not truncated

srsampson:
I noticed while playing with ofdm->doc that it should use ofdm->m for a cleaner value, but then I noticed that ofdm->m was using an (int) cast which would result in 143 rather than 144.

So this change makes it roundf(ofdm-fs / ofdm->rs)

### DIFF
--- a/src/ofdm.c
+++ b/src/ofdm.c
@@ -260,7 +260,7 @@ struct OFDM *ofdm_create(const struct OFDM_CONFIG *config) {
     }
 
     ofdm->rs = (1.0f / ofdm->ts);                 /* Modulation Symbol Rate */
-    ofdm->m = (int) (ofdm->fs / ofdm->rs);        /* 700D: 144 */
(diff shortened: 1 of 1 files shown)


drowe67:
Hi @srsampson:

1. IIRC the intention is for all values of m to be integers.
1. For 700D Ts*Fs = 0.018*8000 = 144.
1. Are the any examples of the current maths giving the wrong value for `ofdm->m`?  Could these be trapped by an assert?

srsampson:
I think I'm just being overly cautious maybe:

#include <math.h>
#include <stdio.h>

int main(int argc, char **argv) {
    float ts = .018f;
    float fs = 8000.0f;
    float rs = 1.0f / ts;
    float m = fs / rs;
    float mi = (int) m;
    float mr = roundf(m);

    fprintf(stdout, "m = %f, mi = %f, mr = %f\n", m, mi, mr);
}

Output:

$ ./test
m = 144.000000, mi = 144.000000, mr = 144.000000

[image: Screenshot from 2022-06-08 16-45-02.png]

PR #319, opened 2022-04-27
FreeDV 2020C

drowe67:
Following on from #274. In the [blog post](https://www.rowetel.com/?p=8014) write up of 2020A/2020B, one of the further work items was to try 2020B with a high rate FEC code that can protect all bits.  The idea is that once a threshold SNR is reached, 2020C will be error free.

@billcowley has kindly designed a suitable (212,158) LDPC code.

Note branch is named dr-2020B as 2020C was initially derived from 2020B.

TODO:
1. Build up mode in FreeDV API :heavy_check_mark: 
2. ctests :heavy_check_mark: 
3. "Ear protection" - some work on reducing loud noises due to uncorrected bit errors (see below).  Applied to all LPCNet (2020) modes  :heavy_check_mark:
4. Test OTA.  One day automated test campaign was inconclusive - didn't show any difference between 2020B and 2020C (see below). :heavy_check_mark:

Simulation on an AWGN channel at about 6dB SNR, 2020C sounds better than 2020B (which has unequal error protection):
```
./src/freedv_tx 2020C ~/LPCNet/wav/all.wav - --clip 1 | ./src/ch - - --No -21 | ./src/freedv_rx 2020C - - --vv | aplay -f S16_LE -r 16000
<snip>
ch: SNR3k(dB):     6.27  C/No....:    41.04
ch: peak.....: 17543.86  RMS.....: 10045.32   CPAPR.....:  4.84
```

PR #334, opened 2022-07-11
remove 2020x index opt support that broke FreeDV API

drowe67:
1. Remove experimental adv.lpcnet_vq_type used for 2020x index opt experiments.
2. Make FreeDV API compatible with earlier versions to support 3rd party applications.
3. Hard coded`vq_type` for 2020 and 2020B, rather tan making it an API selectable option.
4. Support a codec2 1.0.4 release candidate 2 tag/release, to allow an up to date freedv-gui version to be released.

Note that given 2020C is a work in progress https://github.com/drowe67/codec2/pull/319, we are not sure if 2020B will continue to be supported, so it should still be considered experimental.

### DIFF
--- a/src/freedv_2020.c
+++ b/src/freedv_2020.c
@@ -36,7 +36,7 @@
 extern char *ofdm_statemode[];
 
 #ifdef __LPCNET__
-void freedv_2020x_open(struct freedv *f, int vq_type) {
+void freedv_2020x_open(struct freedv *f) {
     f->speech_sample_rate = FREEDV_FS_16000;
     f->snr_squelch_thresh = 4.0;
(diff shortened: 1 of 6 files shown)


drowe67:
@tmiw - when convenient can you please see if this works OK with freedv-gui?  Then we can try another 1.0.04 release if you like.

@hobbes1069 - Richard we have been working on the issue you raised on the mailing list - thanks.

tmiw:
> @tmiw - when convenient can you please see if this works OK with freedv-gui? Then we can try another 1.0.4 release if you like.

I'm able to compile and execute fine with this branch + LPCNet master. Is there another LPCNet tag/branch I should try as well?

PR #337, opened 2022-07-17
update cmake

maitbot:
Some minor refactoring of Codec2 build to better use CMake
features to improve readability ans maintainability.

About this update cmake patch to codec2:

- Follow the convention of cmake_minimum_required() being the first CMake command.
  (Moved CMAKE_OSX_DEPLOYMENT_TARGET definition down next to another OSX setting.)

- Update cmake_minimum_required to 3.13.
  (Since that is where POLICY CMP0079 is introduced. We want to follow NEW usage,
   so that seems best. The cmake_policy() handling can be removed.)

- Use (new since CMake 3.0) project() features.
  This is the best single-point definition of project version information.
  (This handles the CODEC2_VERSION* variables for us, so the other definitions
   can be removed.)

- Better DATE_RESULT setting.
  Using CMake's string(TIMESTAMP... function does this in 1 line rather than 5.
  Also, CMake honors the SOURCE_DATE_EPOCH environment variable needed to support
  reproducible builds.
  https://reproducible-builds.org/specs/source-date-epoch/

- Move the SOVERSION setting next to where it is used in the library target definition.
  (Moved from CMakeLists.txt to src/CMakeLists.txt)

- Handle math library via codec2 library target and executable target definitions.
  No need for CODEC2_REQUIRED_LIBRARIES everywhere.

- Pass along the lpcnet_freedv.h header as a codec2 library header if needed.
  (Any other lpcnet headers stay private.)

73 de aa4hs,
-Maitland

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ if(EXISTS ${CMAKE_SOURCE_DIR}/.git)
     find_package(Git QUIET)
     if(Git_FOUND)
         execute_process(
-            COMMAND "${GIT_EXECUTABLE}" describe --always HEAD
+            COMMAND "${GIT_EXECUTABLE}" git rev-parse --short HEAD
             WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
             RESULT_VARIABLE res
             OUTPUT_VARIABLE FREEDV_HASH
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,19 +5,13 @@
 # Please report questions, comments, problems, or patches to the freetel
 # mailing list: https://lists.sourceforge.net/lists/listinfo/freetel-codec2
 #
-set(CMAKE_OSX_DEPLOYMENT_TARGET "10.9" CACHE STRING "Minimum OS X deployment version")
-
-project(codec2)
-
-cmake_minimum_required(VERSION 3.0)
-
-# Set policies here, probably should move to cmake dir.
-if(POLICY CMP0075)
-    cmake_policy(SET CMP0075 NEW)
-endif()
-if(POLICY CMP0079)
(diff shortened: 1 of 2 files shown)


kantooon:
Thanks Maitland for doing this. I have a question for you if you don't mind: do you intend to backport this Codec2 version for usage in gnuradio 3.8 on Bullseye or will it be exclusively gr 3.10?
Adrian

drowe67:
Thanks @maitbot looks OK to me :slightly_smiling_face: 

 @hobbes1069 - do you have any comments?

hobbes1069:
Nothing jumps out at me as problematic. I think most of this was written during circa CMake 2.4ish so I'm sure a lot has changed. We should of course make sure it builds for all supported configurations.

CMakeLists.txt
```diff
@@ -65,6 +42,7 @@ endif()
 if(BUILD_OSX_UNIVERSAL)
 set(CMAKE_OSX_ARCHITECTURES "x86_64;arm64")
 endif(BUILD_OSX_UNIVERSAL)
+set(CMAKE_OSX_DEPLOYMENT_TARGET "10.9" CACHE STRING "Minimum OS X deployment version")
 
 #
 # Find the git hash if this is a working copy.
```
tmiw, line 48: We should probably use `git rev-parse --short` to get the commit hash. `git describe` gives me something like `v1.0.0-428-g160866fc` for current master.

However, we can do another PR for that particular change if preferred.

drowe67: @tmiw - would you like to try that change in this PR?  I'm happy to wait for you to try that...

tmiw: Done :)

drowe67: Thank you @tmiw :slightly_smiling_face:

PR #338, opened 2022-07-23
Fix build errors with LPCNet and ctests enabled.

tmiw:
Resolves build error discovered due to recent changes to ctests made in master.


drowe67:
Thanks @tmiw

PR #339, opened 2022-07-24
stm32 ctest maintenance

drowe67:
Several stm32 ctests have broken between v1.0.1 and https://github.com/drowe67/codec2/commit/842c566aa33b7e6e0a91e5395cf47fc8889b8d8a, as per commit comments.

The spare RAM was dropped from 4000 to 3000 bytes in `tst_api_demod.c` in order to get `tst_api_demod 700E_AWGN_test` to stop choking when it achieved sync.  Curiously, the other 700D and 700E tests were still running OK with 4000 bytes spare (including the tests that used the codec).  This suggests when we do BER tests with test frames, we need a bit more RAM.  Test frame mode BER tests aren't run on the SM1000 (we always decode to speech), so we are probably OK.

Status:
1. 28/28 stm32 ctests passing :heavy_check_mark: 
2. Try this build on a SM1000.

### DIFF
--- a/stm32/README.md
+++ b/stm32/README.md
@@ -2,7 +2,11 @@
 
 ## Quickstart
 
-1. Build codec2 for your host system, see [codec2/README.md](../README.md)
+1. Build codec2 (with -DUNITTEST=1) for your host system, see [codec2/README.md](../README.md)
+   ```
+   $ cd ~/codec2
+   $ mkdir build_linux && cd build_linux && cmake -DUNITTEST=1 .. && make
+   ```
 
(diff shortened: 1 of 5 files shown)


tmiw:
On that note, it looks like [there's a version of QEMU that can emulate STM32](https://github.com/beckus/qemu_stm32). This might be something that can be integrated as part of the PR auto-build stuff so that tests can at least be run more often (albeit perhaps not 100% indicative of how it'll behave on the actual SM1000).

drowe67:
Merged into #317

PR #340, opened 2022-07-26
Fix build errors with LPCNet and ctests enabled V2.0

drowe67:
Following on from https://github.com/drowe67/codec2/pull/338


tmiw:
Honestly I'm still not sure why this is failing. On the original PR the build seemed to run okay with -DUNITTEST=1 (see [here](https://github.com/drowe67/codec2/runs/7479490595?check_suite_focus=true)) and per the Codec2 CMake files it shouldn't even be going into the unittest folder unless UNITTEST=1, so the line removed in this PR shouldn't matter.

That said, if @Tyrbiter confirms that this PR works okay, I'm fine with the change.

Tyrbiter:
Please see:

https://github.com/drowe67/codec2/commit/0687cb9b6a7670c532db3b799a024ca20a01957e#commitcomment-79525758

.github/workflows/cmake.yml
```diff
@@ -62,3 +62,8 @@ jobs:
       # Execute tests defined by the CMake configuration.  
       # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
       run: ctest --output-on-failure
+
+    - name: Test library installation
+      working-directory: ${{github.workspace}}/build_linux
+      shell: bash
+      run: cmake --install . --prefix "$HOME/codec2_install" && rm -rf "$HOME/codec2_install"
```
drowe67, line 69: @tmiw - how will this trap the problem we encountered in https://github.com/drowe67/codec2/commit/0687cb9b6a7670c532db3b799a024ca20a01957e?

I'd rather not add more GitHub actions unless there is a strong reason, as they are executed on every push.  My understanding is the root cause here is relaying on an include file that was in the wrong path?  That's not a very common problem.

tmiw: I was reviewing again and realized that the error happened on `make install`--not simply while building. That was why I (and the current GH actions) weren't seeing any issues before. The new GH action will also allow us to validate that all the public include and library files get installed as expected in the future.

BTW the new GH action ran in 0 seconds per [the last results](https://github.com/drowe67/codec2/runs/7544022419?check_suite_focus=true).

drowe67:
@Tyrbiter - can I pls confirm this PR has solved the problem for you?

.github/workflows/cmake.yml
drowe67: Cool - good work tracking it down.  The link between the cmake code and `make install` was pretty obscure :thinking:

Tyrbiter:
> @Tyrbiter - can I pls confirm this PR has solved the problem for you?

Yes, it has.

The 2nd lpcnet_freedv.h instance was the problem for the Fedora build process.

Issue #357, opened 2022-10-11
Binary stdin/stdout on Windows


gvanem:
None of the examples I've tried that pipes data between stdout/stdin and to e.g `sox` works on Windows.
E.g. this command (based on `script\train_700c_quant.sh`):
```
c2enc.exe 700C ./raw/hts1a.raw - --var | c2dec.exe 700C - - | sox -q -t .s16 -c 1 -r 8000 -b 16 - hts1a.wav
./raw/hts1a.raw var:  9.33 std:  3.05
```

generated this `hts1a.wav` that sound like:

https://user-images.githubusercontent.com/945271/195076343-4ba984a8-a4c6-4fa6-82b0-f82b2b276995.mp4

Nice! A `setmode(stdin, O_BINARY)` is needed to fix this. Here's my [attempt](https://github.com/drowe67/codec2/files/9754593/bin-mode-fix.diff.txt) to fix it. So after that patch that same sounds like:

https://user-images.githubusercontent.com/945271/195077897-2ec13f9f-0399-48ba-815b-4d8ef29fec98.mp4

*The Navy attacks the V task force*?

drowe67:
As per https://github.com/drowe67/codec2/issues/356#issuecomment-1275161545 - I'm not sure this is an Issue - more a policy that all our development and tool support is on Unix based environments. 

I'm not interested in supporting Windows C/shell incompatibilities, seems like a corner case to me.

@tmiw - we should probably make an official call on this, and document it in the READMEs.

gvanem:
> I'm not interested in supporting Windows C/shell incompatibilities, 

It's not a shell issue. 
And he [front-page](https://github.com/drowe67/codec2#building-for-windows-on-a-linux-machine) mentions *Building for Windows ...*, so that gave me the impression Windows is supported. Not the case I guess.

tmiw:
I would consider Codec2 supported on Windows (otherwise, the main FreeDV application would have issues running on that platform, too), but I'm not sure if we've attempted to run the ctests or helper applications there much. I took a look at `c2enc`/`c2dec` and it seems we do follow best practice when actual filenames are used (e.g. `fopen` using "b" as part of the mode). The main issue is stdin/stdout as previously mentioned.

Anyway, I'd recommend specifying separate filenames as a workaround. For example:

```
c2enc.exe 700C ./raw/hts1a.raw hts1a_temp.raw --var
c2dec.exe 700C hts1a_temp.raw hts1a_out.raw
[...]

PR #358, opened 2022-10-12
Force stdin and stdout to be binary on win32 for helper applications.

tmiw:
Per #357, while we are passing "b" to `fopen()` calls in the applications we provide, we don't do similar for stdin/stdout, causing confusion for Windows users. This PR adds the necessary code to allow piping between applications to work properly.

(Thanks @gvanem for the initial fix!)

### DIFF
--- a/src/c2dec.c
+++ b/src/c2dec.c
@@ -36,6 +36,11 @@
 #include <errno.h>
 #include <getopt.h>
 
+#ifdef _WIN32
+#include <io.h>
+#include <fcntl.h>
+#endif /* _WIN32 */
+
 #define NONE          0  /* no bit errors                          */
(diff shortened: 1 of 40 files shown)


gvanem:
Thanks, but I've not tried it. 
But the case of `fclose(stdin);` is still there. Here MinGW crashes if one closes `stdin` / `stdout`.

tmiw:
> Thanks, but I've not tried it. But the case of `fclose(stdin);` is still there. Here MinGW crashes if one closes `stdin` / `stdout`.

My understanding is that you should be able to close stdin/stdout using `fclose()` without any problems, so the check from your original patch shouldn't be needed. If you're able to run some sort of debugger, it would be good to see exactly where it's crashing on your machine.

Issue #352, opened 2022-09-08
Data mode SNR Estimation


drowe67:
@DJ2LS is using the Codec 2 data modes for FreeDATA, and has an issue with the SNRs reported by the demod:

1. As the actual channel SNR reduces, `datac0` and `datac3` tend to peak at 10dB, whereas `datac1` will peak at over 20dB. 
2. We need reported SNRs to be consistent across modes to support "gear shifting" between modes.

drowe67:
Some notes on the issue:
1. `datac1` doesn't have clipping enabled be default, the other modes do.
2. Clipping increases the (peak power)/noise ratio with maximises SNR at the receiver when using SSB PAs which are defined by their peak power output.
3. Unfortunately clipping is a form of distortion that sets a ceiling on the measured SNR.  At low SNRs, this is not noticeable, as the channel noise dominates.  As the channel SNR reaches about 10dB, the measured SNR flattens out.
4. The fix might be as simple as enabling `datac1` clipping by default, so all modes have consistent behavior.
5. It would be useful to have a test that verifies the SNRs reported can be used for gear shifting.
6. It would also be useful to document the effect, and plot a a curve of measured SNR versus actual channel SNR.

drowe67:
We tuned the peak levels (including the data modes) added a test for peak levels in #341, which is related:
```
ctest -V -R test_peak_levels
```

drowe67:
This issues has been addressed and documented in `README_data.md` as part of https://github.com/drowe67/codec2/pull/362

PR #362, opened 2022-12-01
Tools for raw data mode

drowe67:
Some raw data mode tools and maintenance.

1. `unittest/spectrogram.sh`, plot spectrogram PNGs from wave files.
1. fix mark:space ratio in `freedv_data_raw_tx`
1. Research accuracy of SNR estimates (#352), generate PER/BER v SNR curves, compare Octave and C Tx, compare calibration of SNR between Octave Tx and `ch` tool:
    Usage:
    ```
    cd codec2/unittest/raw_data_curves
    make
    ```
1. While plotting curves found & fixed a bug in pre/post amble levels, which may have affected `datac0` and `datac3`.
1. Added a BPF to `datac1` and set up levels for clean/compressed mode operation.  Previously it was compressed but there was no BPF filter.
1. Update `README_data.md` with PER v SNR & SNRest v SNR curves and related comments.
1. `test_OFDM_modem_datac0_compression` to check compressed raw data waveform


drowe67:
@DJ2LS - I've written some scripts to plot SNR estimates from the raw data modem, against actual SNR.  The estimates don't look very good!  However the slopes look about right, so it might be a y-offset thing.  I'll keep digging :thinking: 

The larger variance for datac0 is kind of expected as it's a shorter frame, so less symbols.

snr_curves

DJ2LS:
interesting, and thanks for having a look at this, @drowe67 

For the FreeDATA protocol I'm now using reported SNR (min 5dB for datac1) + condition of at least two successful decodings at current speed level (datac3) before shifting up to next speed level (datac1). So the SNR offset might cause not too much problems. However, this could become interesting if we maybe have some more modes somewhen with less difference in minimum SNR operation level.

drowe67:
@DJ2LS Here are some better plots, you can see the `datac0` and `datac3` curves bending over at high SNR, as the noise from the waveform compression is greater than the channel noise.  `datac1` is pretty close, about 1dB low, then wanders off at low SNRs, where the waveform falls over.

I still have to explore why `datac0` and `datac3` aren't correct around 0dB.

The PER results are pretty boring in an AWGN channel - that's LPDC codes!

How we gear shift is an open question too, measured packet loss (PER) might be better than SNR.

snr_curves
per_snr

DJ2LS:
> How we gear shift is an open question too, measured packet loss (PER) might be better than SNR.
> 
Yes, for this reason I'm doing both - using SNR and PER (2 decodings) as well for up shifting 
- using PER (2 lost packets) for down shifting

drowe67:
Digging into this a little more there are a couple of mysteries.  Found a bug with the level of pre/post amble in C.

First, lets make sure the Octave (`ofdm_ldpc_tx`) and C (`ch` tool) measure the same SNR when injecting a given noise level.  We trust the Octave version more.  A good results is the PER/BER `o` and `x` curves on top of each other. This means that both tools report the same SNR for a given PER/BER, and either can be used as the SNR "source of truth" for the next tests.

octave_ch_noise

So that we trust `ch` lets compare the C tx (`freedv_data_raw_tx`) to Octave (`ofdm_ldpc_tx`) without compression:

octave_c_tx

Now lets turn on compression in the C Tx, note the performance drops about 1dB compared to the reference uncompressed Octave Tx. However we gain about 4dB in PAPR (not shown on this figure), so we have a net 3dB gain for a typical SSB PA that is peak power limited.

octave_c_tx_comp

Here is the SNR estimates returned by the modem, compared to actual SNR (uncompressed case):

snrest_snr

DJ2LS:
@drowe67 This may not be SNR related in case of datac0, but I'm also wondering about these parts here:
Is there a specific reason, this area doesn't contain all data modes?
https://github.com/drowe67/codec2/blob/5b109c27d359b17a5f1f8c721b91aec12c2426e3/src/ofdm.c#L565-L568
https://github.com/drowe67/codec2/blob/5b109c27d359b17a5f1f8c721b91aec12c2426e3/src/ofdm.c#L958-L963

https://github.com/drowe67/codec2/blob/5b109c27d359b17a5f1f8c721b91aec12c2426e3/src/ofdm_mode.c#L107

drowe67:
> @drowe67 This may not be SNR related in case of datac0, but I'm also wondering about these parts here: Is there a specific reason, this area doesn't contain all data modes?

Yes because I haven't set up compression (clipping) for datac1 yet.  There might be a few more dB if we set that up.

drowe67:
Composite curve showing PER for AWGN (thin) and MPP (thick) against SNR:
c_tx_comp

Throughput of payload data for modes against SNR - where lines cross is the best SNR to "gearshift" between modes,  Unfortunately this is channel dependent.
c_tx_comp_thruput

Updated SNRest v SNR:
snrest_snr

DJ2LS:
So it seems, suggested gear shifting SNR for datac1 is at about 2.5 dB? This would be a great improvement. Actually I'm shifting to datac1 at about 5dB

drowe67:
@DJ2LS the datac0 MPP PER curve is also interesting - it's a poor performer as SNR increases, due to it's short duration I think.  You could get into situations where the fwd packet is fine, but the reverse ACK is missed.

DJ2LS:
@drowe67 the protocol would handle this scenario. 

But when doing tests with the modem and just some carriers, about 2 or 3, it felt like it had problems on high SNR. Behaviour increased with decreasing carriers and therefore an increased signal strength on RX side. 

Independent from the "low snr" topic, maybe worth a look somewhen 🤔

drowe67:
@DJ2LS - I'll merge this in the next few days.  Can you please try this branch with FreeDATA and see if there are any issues?  The `datac0` and `datac1` waveforms have changed, as per top of PR.

DJ2LS:
@drowe67 tested with a 200kb file. Seems running stable for me 👍

PR #365, opened 2022-12-10
Clarify Windows support in the README

### LINKED ISSUE #357: Binary stdin/stdout on Windows

tmiw:
Resolves #357 by clarifying the current extent of the project's support for Windows and providing alternate instructions using the WSL for development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ On Ubuntu Linux:
    make
    ```
    
-This will create a working libcodec2.dll file for use with other applications (e.g. FreeDV GUI which is in wide spread use on Windows).  Please note the utility/development command line applications (e.g. `freedv_rx.exe`) may not work exactly the same on the Windows CLI compared to running on a Unix machine/shell.  For example pipes may not function as expected, and ctests are not supported.  Our primary development and test environment is Unix, and we lack the resources to support and maintain these applications for other operating systems.
+This will create a working `libcodec2.dll` file for use with other applications (e.g. FreeDV GUI which is in wide spread use on Windows).  Please note the utility/development command line applications (e.g. `freedv_rx.exe`) may not work exactly the same on the Windows CLI compared to running on a Unix machine/shell.  For example pipes may not function as expected, and ctests are not supported.  Our primary development and test environment is Unix, and we lack the resources to support and maintain these applications for other operating systems.
 
 We have anecdotal reports that Windows Subsystem For Linux (WSL) provides better support for the Codec 2 command line applications.
 
--- a/README.md
+++ b/README.md
@@ -198,8 +198,6 @@ On Ubuntu Linux:
    
 This will create a working `libcodec2.dll` file for use with other applications (e.g. FreeDV GUI which is in wide spread use on Windows).  Please note the utility/development command line applications (e.g. `freedv_rx.exe`) may not work exactly the same on the Windows CLI compared to running on a Unix machine/shell.  For example pipes may not function as expected, and ctests are not supported.  Our primary development and test environment is Unix, and we lack the resources to support and maintain these applications for other operating systems.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -204,6 +204,23 @@ On Ubuntu 17 and above:
  wait for it to build.
  ```
 
+The above will create a working libcodec2.dll file for inclusion in other applications
+but the utility programs (e.g. `freedv-rx.exe`) are not guaranteed to work properly in 
+a native Windows environment. Usage of Codec2 via the Windows Subsystem For Linux (WSL) 
+is recommended:
```
drowe67, line 210: 1. All good up to the end of the first sentence on line 209.
2. Beyond that I feel recommending WSL is going beyond (and indeed the opposite direction) to what we agreed at PLT - it implies ongoing support, and will lead to the same confusion that we are trying to resolve.  As per our PLT decision I'd prefer to leave it at the first sentence.
3. I'd also suggest we rm 

    ```
    Building for Windows on a Windows machine
    
    mkdir build_windows (Or what ever you want to call your build dir)
    cmake -G "MinGW Makefiles" -D CMAKE_MAKE_PROGRAM=mingw32-make.exe
    Or if you use ninja for building cmake -G "Ninja" ..
    mingw32-make or ninja  depends on what you used in the last command
    wait for it to build.
    ```
    as we aren't testing that regularly.

drowe67:
@tmiw - I've had a go at editing `README.md` - see what you think ...

README.md
```diff
@@ -186,24 +186,19 @@ wav         - speech files in wave file format
...
- ```
- mkdir build_windows (Or what ever you want to call your build dir)
- cmake -G "MinGW Makefiles" -D CMAKE_MAKE_PROGRAM=mingw32-make.exe
- Or if you use ninja for building cmake -G "Ninja" ..
- mingw32-make or ninja  depends on what you used in the last command
- wait for it to build.
- ```
+This will create a working `libcodec2.dll` file for use with other applications (e.g. FreeDV GUI which is in wide spread use on Windows).  Please note the utility/development command line applications (e.g. `freedv_rx.exe`) may not work exactly the same on the Windows CLI compared to running on a Unix machine/shell.  For example pipes may not function as expected, and ctests are not supported.  Our primary development and test environment is Unix, and we lack the resources to support and maintain these applications for other operating systems.
 
+We have anecdotal reports that Windows Subsystem For Linux (WSL) provides better support for the Codec 2 command line applications.
```
tmiw, line 201: Probably not necessary to have this here as WSL is effectively a Linux VM (and thus all the normal Codec2 for Linux support applies).

README.md
tmiw: Per my added comment below WSL is a Linux VM running inside Windows (with some support from the kernel to allow seamless access to file I/O etc.) so I'm not sure this is really supporting Windows per se. Though since it is basically still Linux, perhaps we don't need to have that paragraph either.

PR #366, opened 2022-12-16
refactor to use type 'bool' consistently

cybork42:
The sources of **freedv-gui** use type _bool_ often but not consistently yet.
Apparently the same goes for the **codec2** sources.
I refactored all "enable" variables  and the interface to use _bool_ types and check for _true_ and _false_,
which also required to add "include <stdbool>" into some source files.

### DIFF
--- a/src/freedv_api.c
+++ b/src/freedv_api.c
@@ -859,7 +859,7 @@ int freedv_bits_to_speech(struct freedv *f, short speech_out[], short demod_in[]
     int decode_speech = 0;
     if ((rx_status & FREEDV_RX_SYNC) == 0) {
 
-        if (f->squelch_en == 0) {
+        if (!f->squelch_en) {
 
             /* pass through received samples so we can hear what's going on, e.g. during tuning */
 
@@ -889,7 +889,7 @@ int freedv_bits_to_speech(struct freedv *f, short speech_out[], short demod_in[]
        /* following logic is tricky so spell it out clearly, see table
           in: https://github.com/drowe67/codec2/pull/111 */
 
-       if (f->squelch_en == 0) {
+       if (!f->squelch_en) {
            decode_speech = 1;
        } else {
            /* squelch is enabled */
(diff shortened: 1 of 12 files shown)


drowe67:
@cybork42 - thank you for your PR.  I'm about to do some traveling, so it may be a few days before I have a chance to review it.

tmiw:
Note that this PR will cause Codec2 to mandate C99 compiler support at minimum. This probably isn't an issue for GCC or Clang but might become such for other compilers/platforms. Will need to investigate more to be sure.

drowe67:
> Note that this PR will cause Codec2 to mandate C99 compiler support at minimum. This probably isn't an issue for GCC or Clang but might become such for other compilers/platforms. Will need to investigate more to be sure.

@tmiw - would that already be the case?  I can see quite a few `codec2/src`, and the use of `stdbool.h`.  I think @cybork42 is making the point that current use is inconsistent, and this PR fixes that.

tmiw:
> > Note that this PR will cause Codec2 to mandate C99 compiler support at minimum. This probably isn't an issue for GCC or Clang but might become such for other compilers/platforms. Will need to investigate more to be sure.
> 
> @tmiw - would that already be the case? I can see quite a few bools in `codec2/src`, and the use of `stdbool.h`. I think @cybork42 is making the point that current use is inconsistent, and this PR fixes that.

Indeed. The question was more whether we're okay with officially mandating C99 (in which case this PR is fine) or whether  we want to remove the existing references to `bool` / `stdbool.h` in the codebase and potentially work on older compilers. 

OTOH I've never heard of any complaints about having problems compiling on a compiler that can't do C99, so I'd lean towards this PR being fine 👍

cybork42:
One day soon I will have to abandon my old Ubunto LTS 18, where the Compiler is ancient enough to not know 'bool' fight away without additional include statements. 

So, yes, those compilers exist and they are in use, i'm my own example here! 

At the same time I realized a warning during the build process about a statement 'git switch' saying that 'switch' is not a valid git command. Git in Ubuntu 18 is 2.0...something, and doesn't know switch really... 



Am 17. Dezember 2022 08:08:51 MEZ schrieb Mooneer Salem ***@***.***>:

src/freedv_api.c
```diff
@@ -859,7 +859,7 @@ int freedv_bits_to_speech(struct freedv *f, short speech_out[], short demod_in[]
     int decode_speech = 0;
     if ((rx_status & FREEDV_RX_SYNC) == 0) {
 
-        if (f->squelch_en == 0) {
+        if (!f->squelch_en) {
```
drowe67, line 863: I haven't used `bool` much, so just did some reading and confirmed `if (!bool_variable)` always means "if variable false". Makes sense.  I was curious as I am unsure what value (0 or 1) is used for `true` internally.

drowe67:
Sorry about the delay, I have been away on a sailing/camping trip :boat: 

@tmiw I am OK to merge this if you are OK re the C99 question you mentioned above.  When convenient, could you please try this branch with `freedv-gui` and test the squelch logic is still working OK?  That logic is quite complex and this PR affects some of it.

tmiw:
> @tmiw I am OK to merge this if you are OK re the C99 question you mentioned above. When convenient, could you please try this branch with `freedv-gui` and test the squelch logic is still working OK? That logic is quite complex and this PR affects some of it.

Seems to work fine on all .wav files in freedv-gui except for `all_2020b_mpp.wav`, which seems to be getting detected as 2020C instead of 2020B as expected (and thus not decoding properly). I'm not sure if that's related to this PR, though.

PR #368, opened 2022-12-18
Allow TX BPF to be set for all OFDM modes.

tmiw:
1. Per https://github.com/drowe67/freedv-gui/issues/317, the API to change TX BPF should work for all OFDM modes, not just 700D and E.
2. (experimental until [consensus](https://github.com/drowe67/freedv-gui/issues/317#issuecomment-1365488829) is reached) enable clipping and BPF by default on all OFDM modes.
3. Remove DPSK support at the FreeDV API level as this is obsolete with fast fading modes like 700E/2020B/2020C.  Bumped API number to alert end users (although I imagine this feature was not used by anyone).


CMakeLists.txt
```diff
@@ -930,7 +930,7 @@ if(LPCNET)
          set_tests_properties(test_freedv_reliable_text_awgn_1600 PROPERTIES PASS_REGULAR_EXPRESSION "1")
                             
     add_test(NAME test_freedv_reliable_text_fade_2020
-          COMMAND sh -c "cd ${CMAKE_CURRENT_SOURCE_DIR}/unittest; ./reliable_text_fade.sh 2020 -26 6 0 '${CMAKE_CURRENT_BINARY_DIR}/src'")
+          COMMAND sh -c "cd ${CMAKE_CURRENT_SOURCE_DIR}/unittest; ./reliable_text_fade.sh 2020 -26 4 0 '${CMAKE_CURRENT_BINARY_DIR}/src'")
 endif(LPCNET)
```
drowe67, line 934: Yes I've had a few fails on this test recently, with my inertnal (data mode) development.

src/freedv_api.c
```diff
@@ -1269,8 +1269,7 @@ void freedv_set_clip(struct freedv *f, int val) {
 /* Band Pass Filter to cleanup OFDM tx waveform, only supported by some modes */
 
 void freedv_set_tx_bpf(struct freedv *f, int val) {
-    if (FDV_MODE_ACTIVE( FREEDV_MODE_700D, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_700E, f->mode) 
-        || FDV_MODE_ACTIVE( FREEDV_MODE_DATAC0, f->mode) || FDV_MODE_ACTIVE( FREEDV_MODE_DATAC3, f->mode)) {
+    if (is_ofdm_mode(f)) {
         ofdm_set_tx_bpf(f->ofdm, val);
```
drowe67, line 1273: Yes I think this is the correct approach, if indeed compression is set up for all OFDM modes (voice and data).  We may be able to go a little further;
1. Look for other long lists of modes in the compression code and replace with `is_ofdm_mode(f)`.
2. Make sure there are asserts to bomb out if we hit a mode that doesn't have compression set up (like forgetting to define a filter).

I'll poke about some more over the next few days (or you are welcome to take a look).

CMakeLists.txt
tmiw: I also noticed that there aren't equivalent tests for 2020B/C, so that might be good to add too.

src/freedv_api.c
tmiw: Go for it 👍

tmiw:
FYI I created `ms-codec2-test-build` with this PR and #371 for the purposes of creating a freedv-gui 1.8.7 test build.

drowe67:
@tmiw FYI I was getting a ctest fail on `90 - test_freedv_reliable_text_fade_1600 (Failed)` on my local machine, but it was passing on the GitHub machines.  I rm-ed `unittest/fast_fading_samples.float`, used `make` to regenerate it and it passed.  I think I had manually generated longer fading files which would have mean they are different from those used to write the test.  As the tests are necessarily short, they are dependent on the particular fading file rater than the statistics of the fading channel.

drowe67:
@tmiw - OK to merge this one?  IIRC you released a freedv-gui test build with it.

tmiw:
> @tmiw - OK to merge this one? IIRC you released a freedv-gui test build with it.

Yep, go for it. 👍

PR #377, opened 2023-02-17
Add 8 to 24K conversion functions

tmiw:
This PR adds functions to convert between 24K float samples and 8K short samples (required for the Flex 6000 series). Currently experimental while testing is done to compare with soxr (currently used by the waveform).


tmiw:
Don't mind me, just doing some experimenting. Not sure if this will help at all (especially since soxr is supposed to be the best performing) but worth a shot.

tmiw:
So I don't think this helps enable the Flex waveform to run on the radio. However, I'm thinking about possibly adding support for the Flex to ezDV, so this would be needed for that. And once the waveform is actually able to run (if it can), this would reduce the number of required dependencies.

tmiw:
I wrote some ezDV code to interface with the Flex (basically a simplified version of the Flex waveform code) and am currently using this branch for testing (see https://github.com/tmiw/ezDV/commit/ae38f6be5735bd3c5219d24cda738d6132cae2e4). Looks like the upsampling and downsampling in this PR is working fine 👍 

Screenshot 2023-03-31 at 5 04 52 PM

So yeah, I'd say we can go ahead and review/merge.

drowe67:
Hi @tmiw - so I'm not convinced this should be in `codec2`.  I'd like to avoid adding any more utility or support code to codec2 unless needed by the core DSP.  As the number of platforms grows I want to avoid placing platform specific code in codec2.

As an aside I'm also planning on removing quite a bit of code from the codec2 repo, and perhaps having an experimental repo  where "anything goes" but we don't maintain or release.

A good test - are there any applications or other codec2 source files that use these resamplers?  For example `fdmdv_8_to_16` is used for the implementation of FreeDV 2020. For freedv-gui we use a generic resampler - that's actually a better way to go (e.g. generate coeffs at init time).

On a related note can you pls talk with me first before kicking of codec2 PRs?  I'd like to be involved in the decision about what goes into codec2 - and best to do this before the coding starts, at the design/architecture stage.

tmiw:
> A good test - are there any applications or other codec2 source files that use these resamplers? For example `fdmdv_8_to_16` is used for the implementation of FreeDV 2020. For freedv-gui we use a generic resampler - that's actually a better way to go (e.g. generate coeffs at init time).

ezDV operates natively at 8K sample rate (as it configures its built-in TLV320 audio codec chip/I2S interface to operate as such). The only time it does any sort of sample rate conversion is when communicating with Flex radios as Flex needs 24K floating point samples. The current waveform code also uses this instead of soxr as I was testing to see if we could get enough headroom to avoid the issues the waveform has traditionally had, but unfortunately that doesn't seem to be the case at the moment.
 
Anyway, we could just keep this PR open for now until we get some clarity on what would be needed to get the waveform working on the radio itself and/or figure out how the Codec2 code is going to be organized from here on out. At that point, this code can just be moved wherever it needs to go and the PR closed.

> On a related note can you pls talk with me first before kicking of codec2 PRs? I'd like to be involved in the decision about what goes into codec2 - and best to do this before the coding starts, at the design/architecture stage.

Will do.

drowe67:
Thanks for that explanation @tmiw. So this is experimental code to help integrate codec2 on two possible platforms?

How about we keep the branch but close the PR for now?  That way it's off my radar while the experimenting is ongoing.  I don't think there's a good reason to merge just yet but  agree that we could take a look at it once the experimentation matures.

tmiw:
> Thanks for that explanation @tmiw. So this is experimental code to help integrate codec2 on two possible platforms?

One radio, really (Flex) but yeah, two separate devices (ezDV and on the radio itself).

> How about we keep the branch but close the PR for now?

That works for me.

drowe67:
@tmiw - Another few notes on this topic - recording them here for later discussion:

1. I also just noticed I have 48 <-> 8 kHz rate conversion functions in fmddv.c, these are used to support FreeDATA (in much the same was this PR supports other applications) so arguably belong outside of libcodec2 as well.

1. I've been messing around with the fourier transform of rectangular functions lately (a sinc(x) function) - we could use these to generate filter coeffs for any filter (resampling of just a general LPF/BPF) at init time.  This would mean a single function for resampling (at any rate) , and replace the growing number of mode-specific filters in filter_coeff.h, further simplifying ofdm.c

1. If we have a user-defined filter/resampler - that might be a good reason to put the code in libcodec2 - as (a) the code will be used by many parts of libcodec2 and (b) if end-users decide to use it (say to support a certain platform) we won't need to change/re-release libcodec2 to provide utility code for that end user/platform platform.

drowe67:
4. Or we could just link with some one elses generic resample library I guess :thinking:

PR #380, opened 2023-02-25
Generate slow fading files as part of the build process

tmiw:
During some freedv-gui testing, I found that the `--mpg` flag in `ch` didn't work because `slow_fading_samples.float` wasn't getting generated. This PR causes that file to be built.

### DIFF
--- a/unittest/CMakeLists.txt
+++ b/unittest/CMakeLists.txt
@@ -114,11 +114,13 @@ ExternalProject_Add(cml
 
 # Create fading files (used for channel simulation) as part of unit test setup
 add_custom_target(fading_files ALL
+  DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/slow_fading_samples.float
   DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/fast_fading_samples.float
   DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/faster_fading_samples.float
   )
 add_custom_command(
-  OUTPUT  ${CMAKE_CURRENT_BINARY_DIR}/fast_fading_samples.float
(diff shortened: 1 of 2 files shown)


drowe67:
@tmiw as per https://github.com/drowe67/freedv-gui/issues/79#issuecomment-1445066046 I don't think we need this PR.

tmiw:
Could this potentially be done optionally in the few times it's actually needed? Alternatively, we can have something in the documentation for how to manually generate this particular fading file if needed?

drowe67:
It already chokes if a fading file doesn't exists and give instructions:
```
/src/ch ../raw/ve9qrp.raw /dev/null --mpp
ch: Fs: 8000 NodB: -100.00 foff: 0.00 Hz fading: 2 nhfdelay: 16 clip: 32767.00 ssbfilt: 1 complexout: 0
ch: Fading file finished - simulation stopping.  You may need more samples:
$ octave --no-gui
octave:24> pkg load signal
octave:24> time_secs=60
octave:25> ch_fading("faster_fading_samples.float", 8000, 2.0, 8000*time_secs)
octave:26> ch_fading("fast_fading_samples.float", 8000, 1.0, 8000*time_secs)
octave:27> ch_fading("slow_fading_samples.float", 8000, 0.1, 8000*time_secs)
-----------------------------------------------------
```

Medium term where I'd like to go with ch is on the fly generation of fading samples, so this should all be cleaned up over the term of the grant.  However that's not something I want to work on right now.

tmiw:
Actually, I'm not sure that message makes it fully obvious what's going on. I think something that specifically says that it can't find the needed fading file would be better. Low priority, of course, given that no ctests currently use the slow fading files :)

Anyway, this PR can be closed for now.

Issue #383, opened 2023-03-04
Not able to use `all` or `install` targets when cross-compiling


tyler-gilbert:
I get linker errors when I try to execute the `all` or `install` targets when building using:

```
PATH=<path to arm none eabi compiler>
cmake \
-DCMAKE_SYSTEM_NAME=Generic \
-DCMAKE_SYSTEM_PROCESSOR=arm \
-DCMAKE_C_COMPILER=arm-none-eabi-gcc \
-DCMAKE_CXX_COMPILER=arm-none-eabi-g++ \
-DCMAKE_C_FLAGS='-mcpu=cortex-m4 -mfloat-abi=hard -mfpu=fpv4-sp-d16 -mthumb -Os' \
-DCMAKE_EXE_LINKER_FLAGS='--specs=nosys.specs' \
-DCMAKE_INSTALL_PREFIX=<my output directory> \
-DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER \
-DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY \
-DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY \
-DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ONLY \
-DCMAKE_CROSSCOMPILING=ON \
-DBUILD_SHARED_LIBS=OFF \
-GNinja \
..

With this setup: `ninja codec2` works bug `ninja install` fails with linker errors.

I will send a PR that adds a CMAKE build option called `BUILD_EXECUTABLES` which will exclude executables from the build so that `ninja install` works with embedded.

drowe67:
Not sure this is an actual issue - we build for the stm32  (see sm1000) already.  I'm not familiar with ninja, so we need to decide if we will support it going forward. Any thoughts @tmiw ?

tyler-gilbert:
This also applies to `make all` or `make install` as well when using the Makefiles generator.

tmiw:
What's the use case, exactly? Typically for embedded targets codec2 is statically linked into the .bin or .hex file that ultimately gets flashed.

Also, cross-compiling with CMake typically works by setting the needed compiler, etc. arguments into another file and specifying that file with `-DCMAKE_TOOLCHAIN_FILE=...`. More information can be found [here](https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html).

tyler-gilbert:
[...]

PR #384, opened 2023-03-04
Issue383 fix install target with embedded

tyler-gilbert:
This PR address Issue #383. It adds a cmake option to exclude all `add_executable` calls from the build. This makes it easier to just build and install that library statically on embedded targets.


tyler-gilbert:
Sorry. This PR has a problem. Closing.

Issue #382, opened 2023-03-04
codec2_encode() API omits const for input speech array.


tyler-gilbert:
Changing `short speech_in[]` to `const short speech_in[]` adds a little robustness to the API.

I will send a PR for this.

tmiw:
Closing this issue since the associated PR was closed.

PR #385, opened 2023-03-04
Issue #382 add const to codec2 speech_in parameter

tyler-gilbert:
This PR adds `const` to the `speech_in` parameter when encoding. This adds a little robustness to the API for users.

### DIFF
--- a/src/codec2.c
+++ b/src/codec2.c
@@ -59,24 +59,24 @@
 
 \*---------------------------------------------------------------------------*/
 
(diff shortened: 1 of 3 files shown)


drowe67:
Thanks.  I'm not a strong believer in using `const` - I've never actually experienced a bug that using it would have fixed.  Also this breaks the codec2 API for end users, likely requiring them to make code changes.  We need a really good reason to do that.

So I'm inclined to close this one without merging, however happy to hear any other thoughts.

tyler-gilbert:
@drowe67, no problem. Thanks for taking a look.

PR #386, opened 2023-03-04
Issue383 add build executables option

tyler-gilbert:
Adds a cmake option called `BUILD_EXECUTABLES` which is `ON` by defualt. If this option is set to `OFF`, all of the `add_executable()` calls are skipped.

This Fixes Issue #383.


tmiw:
I was taking a look at my previous embedded projects that use Codec2 and noticed that one of them had the following:

```
set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
set(MICROCONTROLLER_BUILD 1)
```

Could one or both of these lines potentially solve your issue in lieu of this PR? If so, perhaps the PR can just document these a bit better?

PR #388, opened 2023-03-12
Ensure result is initialized in ofdm_complex_dot_product().

### LINKED ISSUE #387: Uninitialized variable in ofdm.c :: ofdm_complex_dot_product()
At line 698:

https://github.com/drowe67/codec2/blob/master/src/ofdm.c#L698

result is not set here to complex 0.0

This doesn't fail if ```USE_VECTOR_OPS``` or ```__EMBEDDED__``` is defined, but falling through past those options, the variable is then added to in a loop.

So ```result``` being an automatic variable has undefined contents.

From the GNU C Manual:

https://github.com/VernonGrant/gnu-c-language-manual/blob/main/markdown/Local-Variables.md

__Warning__: If the declaration of an automatic variable does not specify an initial value, the variable starts out containing __garbage__.

tmiw:
Resolves #387 by setting `result` to 0 when vector ops and `EMBEDDED` are not enabled.

### DIFF
--- a/src/ofdm.c
+++ b/src/ofdm.c
@@ -695,7 +695,7 @@ typedef float float4 __attribute__ ((vector_size (16)));
 
 static complex float ofdm_complex_dot_product(complex float *left, complex float *right, int numSamples)
 {
-    complex float result;
+    complex float result = 0;
 
 #if USE_VECTOR_OPS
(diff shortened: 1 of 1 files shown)


SoftwareToolworks:
Thanks! for the fix!

PR #403, opened 2023-06-10
Remove GNU99 check for Win32 builds

tmiw:
While building with LLVM based MinGW on macOS, I ended up running into the following error during codec2's CMake run:

```
mooneer@macaron build_windows % cmake -DCMAKE_TOOLCHAIN_FILE=~/freedv-gui/cross-compile/freedv-mingw-llvm-x86_64.cmake ..
-- The C compiler identification is Clang 15.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Users/mooneer/Downloads/llvm-mingw-20220906-ucrt-macos-universal/bin/x86_64-w64-mingw32-clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Git: /opt/local/bin/git (found version "2.39.0") 
-- Codec2 current git hash: f68a2fe4
-- Performing Test COMPILER_SUPPORTS_GNU11
-- Performing Test COMPILER_SUPPORTS_GNU11 - Failed
-- Performing Test COMPILER_SUPPORTS_GNU99
-- Performing Test COMPILER_SUPPORTS_GNU99 - Failed
CMake Error at CMakeLists.txt:95 (message):
  Compiler doesn't seem to support at least gnu99, might cause problems


-- System is MinGW.
-- Looking for include file stdlib.h
-- Looking for include file stdlib.h - found
-- Looking for include file string.h
-- Looking for include file string.h - found
-- Looking for __GNU_LIBRARY__
-- Looking for __GNU_LIBRARY__ - not found
-- Looking for _GNU_SOURCE
-- Looking for _GNU_SOURCE - not found
-- Looking for floor
-- Looking for floor - found
-- Looking for ceil
-- Looking for ceil - found
-- Looking for pow
-- Looking for pow - found
-- Looking for sqrt
-- Looking for sqrt - found
-- Looking for sin
-- Looking for sin - found
-- Looking for cos
-- Looking for cos - found
-- Looking for atan2
-- Looking for atan2 - found
-- Looking for log10
-- Looking for log10 - found
-- Looking for round
-- Looking for round - found
-- Looking for getopt
-- Looking for getopt - found
-- codec2 version: 1.1.0
-- Build type is: Debug
-- Compiler Flags: -gcodeview -Wall -Wno-strict-overflow-g -O2 -DDUMP
-- Libraries linked: 
-- Compilation date = XX2023-06-10XX
-- Configuring incomplete, errors occurred!
```

This PR suppresses this check for Win32 builds as it's assumed that the necessary C language support is there (plus that codec2 builds with no problems).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,6 +83,7 @@ add_custom_target(dist
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wno-strict-overflow")
 
 # Check for what C standard is supported.
+if(NOT WIN32)
 include(CheckCCompilerFlag)
 CHECK_C_COMPILER_FLAG("-std=gnu11" COMPILER_SUPPORTS_GNU11)
 CHECK_C_COMPILER_FLAG("-std=gnu99" COMPILER_SUPPORTS_GNU99)
@@ -94,6 +95,7 @@ elseif(COMPILER_SUPPORTS_GNU99)
 else()
    message(SEND_ERROR "Compiler doesn't seem to support at least gnu99, might cause problems" )
 endif()
+endif(NOT WIN32)
 
 # -fPIC is implied on MinGW...
 if((NOT WIN32) AND (NOT MICROCONTROLLER_BUILD))


tmiw:
@drowe67, may want to bump up the version as well after merge as this will prevent Windows builds from being generated for freedv-gui (unless the latter is updated to use master again instead of specific version tags).

drowe67:
@tmiw - just trying to feel out the context of this one:

1. Is this around cross compiling for Windows? 
2. Why have the checks stopped working?
3. Is it dangerous to remove checks and warning - this is usually bad practice?
4. Is LLVM the standard compiler we use for building Windows packages?

tmiw:
> 1. Is this around cross compiling for Windows?

Yes.

> 2. Why have the checks stopped working?

I think it might be some sort of weirdness with CMake on macOS. In my Ubuntu VM, this change doesn't seem to be necessary (whereas this change plus some unknown other changes on freedv-gui's side re: Hamlib are needed if I try to compile directly).

> 3. Is it dangerous to remove checks and warning - this is usually bad practice?

CMake is currently treating this as an error, not a warning. It's likely not dangerous considering we don't officially support Windows for everything, just enough to allow use with freedv-gui (thus only using a single compiler we're 100% sure supports the relevant standards).

> 4. Is LLVM the standard compiler we use for building Windows packages?

Yes. This was done a while ago to (a) allow builds for ARM based Windows and (b) due to the Docker container becoming less reliable over time.

drowe67:
OK @tmiw - thanks for the explanation.  I'm not entirely comfortable with removing warnings, but I figure we need to move forwards, and it's just for Windows.

I'll tag & release in (my) morning when I'm fresh :coffee: One of those things I've messed up once or twice!

PR #407, opened 2023-11-10
Fix cross compilation with MSVC on host

srcejon:
This patch enables cross compilation (to Android) using Windows host with MSVC as the host compiler.


drowe67:
This repo is [deprecated](https://github.com/drowe67/codec2-dev#readme), and [we don't support non C99 compilers](https://github.com/drowe67/codec2#ports-to-non-c99-compilers) like MSVC.

